PR #5, opened 2017-03-28
Splitting music out of core.

microsoftsam:
Splitting music out of core and adding melodies.


libs/core/_locales/core-jsdoc-strings.json
```diff
@@ -36,7 +36,7 @@
   "String.substr": "Return substring of the current string.",
   "String.substr|param|length": "number of characters to extract",
   "String.substr|param|start": "first character index; can be negative from counting from the end, eg:0",
-  "control": "Runtime and event utilities.\n\r\nRuntime and event utilities.",
+  "control": "Runtime and event utilities.\n\nRuntime and event utilities.",
```
pelikhan: new line issue in file, check it is LF

microsoftsam: How do I do that?

PR #8, opened 2017-03-29
PwmPin configuration issue

microsoftsam:
Do not Merge, this is an issue I'm trying to solve. 

Currently the list of board pins and whether they are Analog, PwmPins or DigitalPins is configured in the DevPins class. However that configuration is based on the CPLAY config, and doesn't take into account other boards like the Arduino M0 which has PwmPins D2-D13 (bar D7). 

We need a way to define different PwmPin configurations per board. In this PR I have made the logical attempt using ifdefs, however shims.d.ts doesn't take those into account, and so complains that there are duplicate fields in that class.

Not sure how to resolve this @pelikhan @mmoskal ideas?

### DIFF
--- a/libs/core/pins.h
+++ b/libs/core/pins.h
@@ -287,6 +287,61 @@
 class DevPins {
   public:
     DevicePin pins[0];
+#if PXT_BOARD_ID == BOARD_ID_ZERO || PXT_BOARD_ID == BOARD_ID_METRO || PXT_BOARD_ID == BOARD_ID_M0
+#define DigitalPin DevicePin
+#define AnalogPin DevicePin
+#define PwmPin DevicePin
+    //% indexedInstanceNS=pins indexedInstanceShim=pxt::getPin
+    //%
+    AnalogPin A0;
+    //%
+    AnalogPin A1;
+    //%
+    AnalogPin A2;
+    //%
+    AnalogPin A3;
+    //%
+    AnalogPin A4;
+    //%
+    AnalogPin A5;
+    //%
+    DigitalPin D0;
+    //%
+    DigitalPin D1;
+    //%
+    PwmPin D2;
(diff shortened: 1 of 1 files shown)


mmoskal:
One way to fix it, would be to allow targets to override files in core. You could then override the entire `pins.h` (possibly moving some common stuff to a different file). For example, we could have `pxt-adafruit/libs/core` with `pxt.json` which would say something like `"inheritsFrom": "../../node_modules/pxt-common-packages/libs/core"`. It would then have the target-specific files (`pins.h`), as well as `shims.d.ts` which are generated (they will be different for different targets).

You could make it completely file based, or you could merge `pxt.json` that is inherited. I would strongly suggest avoiding copying any files to `libs/core` (copying somewhere under `libs/core/built` is of course fine). Otherwise, you will end up editing the wrong copy. There is already something similar in `cli.ts`, search for `commonfiles`. There is no support necessary in the webapp, since it will get the full package as JSON as usual.


Now, this doesn't solve the problem that one target might want multiple boards with different pinouts.

pelikhan:
Could we add more markers in comment to help the parsing engine that generates shims.d.ts to ignore some section? (Without writing a full precompiler)

pelikhan:
File overrides is in https://github.com/Microsoft/pxt/commit/3db73bdceb92be18b022b08af578e733a045dd2e

The build will automatically slurp files from corresponding /libs/yyy folders.

pelikhan:
pins.h should probably be splitted before doing this: the whole pin map definition should stay in core and devpins should be moved in its own devpins.h file (and overriden elsewhere)

pelikhan:
You can override devpins.h and devpins.cpp to change the pin set.

microsoftsam:
Sweet. thanks

PR #9, opened 2017-04-13
support for tagged execution strategy

mmoskal:
To be merged together with PR in pxt


libs/core/core.cpp
```diff
@@ -62,71 +62,377 @@ StringData *substr(StringData *s, int start, int length) {
...
+
+namespace pxt {
 
-static const char sTrue[] __attribute__((aligned(4))) = "\xff\xff\x04\x00"
-                                                        "true\0";
-static const char sFalse[] __attribute__((aligned(4))) = "\xff\xff\x05\x00"
-                                                         "false\0";
+int toInt(TNumber v) {
+    int vv = (int)v;
+    if (vv & 1)
```
abchatra: >vv & 1 [](start = 8, length = 6)

Nit pick:  Why not have macros such as IS_TAGGED here so we can change the tagging mechanism and retry. Feel free to ignore

abchatra:
:shipit:

libs/core/core.cpp
mmoskal: Good point. Will clean up tomorrow.

PR #13, opened 2017-05-02
Synth and onboard speaker support

riknoll:
TODO:

* Sometimes sound cuts out
* Some APIs should not appear if there isn't a built-in speaker


libs/music/music.cpp
```diff
@@ -0,0 +1,106 @@
+#include "pxt.h"
+
+#include "SAMD21DAC.h"
+#include "Synthesizer.h"
+#include "DeviceConfig.h"
+
+#ifndef HAS_SPEAKER
+#define HAS_SPEAKER PIN_SPEAKER != NC
```
pelikhan: Parenthesis to be on the safe side?

libs/music/music.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+            synth->setFrequency((float) frequency);
+
+            fiber_sleep(ms);
+            synth->setVolume(0);
+            wait_ms(20);
+        }
+    }
+    else {
+        PwmPin pitchPin = lookupPin(PIN_A10);
```
pelikhan: Speaker pin?

libs/music/music.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+* Sets the output volume of the on board speaker (if available)
+* @param volume the volume 0...1023
+*/
+//% weight=96
+//% blockId=synth_set_volume block="set speaker volume %volume"
+//% parts="speaker" blockGap=8 advanced=true
+void setSpeakerVolume(int volume) {
+    synthVolume = volume;
```
pelikhan: Shouldn't you set the volume now?

libs/music/music.cpp
```diff
@@ -0,0 +1,106 @@
...
+//% help=music/play-tone weight=90
+//% blockId=music_play_note block="play tone|at %note=device_note|for %duration=device_beat" blockGap=8
+//% parts="headphone" async
+//% blockNamespace=music
+void playTone(int frequency, int ms) {
+    if (current == SoundOutputDestination::Speaker) {
+        auto synth = &getWSynthesizer()->synth;
+
+        if (frequency <= 0) {
+            fiber_sleep(ms);
```
pelikhan: set volume to 0

libs/music/music.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+        else {
+            synth->setVolume(synthVolume);
+            wait_ms(20);
+
+            synth->setFrequency((float) frequency);
+
+            fiber_sleep(ms);
```
pelikhan: if ms <= 0, leave it on

libs/music/music.cpp
riknoll: Nope. This is for when the destination is set to "pin" instead of "speaker"

libs/music/music.cpp
riknoll: why?

libs/music/music.cpp
riknoll: No, that would cause the synth to start making noise.

riknoll: Hmmm... maybe I can check if the volume isn't equal to 0

riknoll: Doesn't seem like that is exposed in CODAL.

libs/music/music.cpp
```diff
@@ -0,0 +1,109 @@
...
+    current = out;
+}
+
+/**
+* Sets the output volume of the on board speaker (if available)
+* @param volume the volume 0...1023
+*/
+//% weight=96
+//% blockId=synth_set_volume block="set speaker volume %volume"
+//% parts="speaker" blockGap=8 advanced=true
```
pelikhan: specify .min .max bounds to get slider

libs/music/music.cpp
pelikhan: Feature request for Joe or add it yourself

libs/music/music.cpp
```diff
@@ -0,0 +1,109 @@
...
+        else {
+            synth->setVolume(synthVolume);
+            wait_ms(20);
+
+            synth->setFrequency((float) frequency);
+
+            if (ms > 0) {
+                fiber_sleep(ms);
+                synth->setVolume(0);
+                wait_ms(20);
```
pelikhan: pin path has 5ms

libs/music/music.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if (current == SoundOutputDestination::Speaker) {
+        auto synth = &getWSynthesizer()->synth;
+
+        if (frequency <= 0) {
+            synth->setVolume(0);
+            fiber_sleep(ms);
+        }
+        else {
+            synth->setVolume(synthVolume);
+            wait_ms(20);
```
pelikhan: is this needed?

PR #20, opened 2017-05-09
support classes for python converter and other goodies

mmoskal:
This PR adds:
* support for uint32 and float NumberFormats when operating on buffers
* pins.packBuffer, unpackBuffer (similar for Python struct.pack)
* I2CDevice class (encapsulates address and in future lock)
* adds return codes to I2C functions
* adds control.fail

Please takes a look and see if the APIs and their names make sense.
CC @pelikhan @abchatra @microsoftsam @riknoll @tballmsft


pelikhan:
LGTM

PR #26, opened 2017-05-11
on-chip software debugger support

mmoskal:
This supports https://github.com/Microsoft/pxt/pull/2057

For full functionality also needs https://github.com/lancaster-university/codal/pull/14 (but works without it)


mmoskal:
Are we OK to merge?

pelikhan:
is it waiting for codal changes?

mmoskal:
Codal changes would be nice but it works without them (except that pausing other threads of execution doesn't work)

PR #27, opened 2017-05-11
Input section doc

ganicke:
Get the cards working so we can finally browse the input docs.


libs/core/docs/reference/input.md
```diff
@@ -1,8 +1,54 @@
...
+
+```
+
+## Light sensor #lightsensor
+
+```cards
+input.onLightConditionChanged(LightCondition.Dark, () => {
+
+})
+input.ambientColor()
```
pelikhan: this is adafruit specific

ganicke: Ah, that's right. I'll have to take that out. Good catch.

PR #33, opened 2017-05-12
Add the button docs

ganicke:
button docs added.

keeping the 'button' prefix for md names to avoid future name collisions


libs/core/buttons.cpp
```diff
@@ -52,12 +52,12 @@ DeviceButton *getButton(int id) {
...
 namespace ButtonMethods {
 /**
- * Do something when a button (``A``, ``B`` or both ``A+B``) is clicked, double clicked, etc...
+ * Do something when a button (`A`, `B` or both `A` + `B`) is clicked, double clicked, etc...
  * @param button the button that needs to be clicked or used
  * @param event the kind of button gesture that needs to be detected
  * @param body code to run when the event is raised
  */
-//% help=input/on-button-event weight=99 blockGap=8
+//% help=input/button-on-event weight=99 blockGap=8
```
pelikhan: routing should be input/button/on-event as they are instance methods of buttons. Same for is-pressed, was-pressed

PR #36, opened 2017-05-13
Add and update temperature docs

ganicke:
* on-termperate-condition.md is no more


libs/temperature/docs/reference/input/temperature.md
```diff
@@ -31,13 +33,14 @@ loops.forever(() => {
         0,
         255
     ))
-    light.pixels.showColor(light.colors(Colors.Red))
+    light.pixels.setAll(light.colors(Colors.Red))
```
pelikhan: use kid friendly block instead (the donut)

PR #37, opened 2017-05-14
Get the music docs to show up

ganicke:


libs/music/docs/reference/music.md
```diff
@@ -1,20 +1,20 @@
...
 
-Generation of music tones.
+Play tones and melodies. Make musical notes.
 
 ```cards
-pins.A8.playTone(0, 0);
-pins.A8.ringTone(0);
-pins.A8.rest(0);
+music.playTone(null,null);
+music.startMelody([], MelodyOptions.Once)
```
pelikhan: This is outdated.

PR #43, opened 2017-05-16
Add more music topics

ganicke:


libs/music/docs/reference/music/play-sound.md
```diff
@@ -0,0 +1,58 @@
+# play Sound
+
+Play some notes of music you have placed in a string.
+
+```sig
+music.playSound('')
+```
+
+Your program doesn't wait for the sound to finish. Then next part of your program starts right away.
```
pelikhan: We should have a single place to document the format in playSound/UntilDone. Otherwise, they will grow out of sync.

ganicke: Yea, probably a general/side topic like "Composing and playing music" linked from the playSoundXxx pages?

pelikhan:
That's be great.

ganicke:
Ok, ignore that commit comment. I added a side topic called 'composing sounds' and linked to it from the playSoundXxx topics.

So, I'm not a music expert but I gave it a stab to see what I could come up with.

PR #50, opened 2017-05-18
Fix the rainbow animation for the longer durations

abchatra:
- Duration was incorrectly used in the pause
- Duration was incorrectly used in offset calculation.


abchatra:
@pelikhan  review please. 
fixes https://github.com/Microsoft/pxt-adafruit/issues/152

PR #61, opened 2017-05-25
First try at some 'control' docs

ganicke:


libs/core/control.ts
```diff
@@ -1,12 +1,15 @@
...
 //% weight=10 color="#31bca3" icon="\uf110" advanced=true
 namespace control {
 
     /**
-     * Display specified error code and stop the program.
+     * Display an error code and stop the program.
+     * @param code an error number to display. eg: 5
      */
+    //% help=control/panic weight=29
+    //% blockId="control_panic" block="panic %code"
```
microsoftsam: I'm not sure we want to surface this as a block. @pelikhan thoughts?

pelikhan:
Under advanced is fine with me.

ganicke:
We can also extend a _control.md_ In **Adafruit** which leaves the advanced methods out.

PR #62, opened 2017-05-26
Add 'event' topics to control docs

ganicke:


ganicke:
This contains branch the previous (cp-control-docs-01). I'll pull out the old commit if necessary.

libs/core/control.ts
```diff
@@ -1,12 +1,15 @@
...
 //% weight=10 color="#31bca3" icon="\uf110" advanced=true
 namespace control {
 
     /**
-     * Display specified error code and stop the program.
+     * Display an error code and stop the program.
+     * @param code an error number to display. eg: 5
      */
+    //% help=control/panic weight=29
+    //% blockId="control_panic" block="panic %code"
```
microsoftsam: Same thing here. see other PR for whether or not we want to surface panic as a block.

ganicke:
pulled the commit from the older branch. "panic' was in that commit.

libs/core/docs/reference/control/on-event.md
```diff
@@ -0,0 +1,92 @@
...
+* **value**: a [number](/types/number) that tells what the cause of the event is, like: `4`.
+* **handler**: the code to run when the event happens.
+
+## Example #exsection
+
+Register two events coming from source `22`. Make pixels light up on the pixel strip when
+the events of `0` and `1` are _raised_.
+
+```blocks
+let pixelLighter = 22
```
pelikhan: use constants for events, this defeatures the way we detect duplicate event in blocks.

libs/core/docs/reference/control/raise-event.md
```diff
@@ -0,0 +1,60 @@
...
+* **src**: the identification [number](/types/number) (the source) of this event, such as: `10`.
+* **value**: a [number](/types/number) tells what the cause of the event is, like: `4`.
+
+## Example #exsection
+
+Register two events coming from source `22`. Make pixels light up on the pixel strip when
+the events of `0` and `1` are _raised_.
+
+```blocks
+let pixelLighter = 22
```
pelikhan: avoid variables.

ganicke: Initially, I tried const but the editor decompiles back to **let** statements. That's why I used the variables to match the experience.  I get your point though and I'll them to const.

ganicke:
Changed the examples to use **const** for source IDs. If you'd rather, I can just use numeric literals for IDs.

PR #63, opened 2017-05-30
Moving simulator code into common packages

riknoll:
Each package now has a `sims` folder that exports an interface for the simulator to implement. In the root of this repo I've added a `tsconfig.json` that builds all the sim folders together into one file called `common-sim.js`. I'll be opening a PR in pxt-core that builds that file as part of the target build and copies it into the target's `built` directory.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,38 @@
+{
+    "compilerOptions": {
+        "target": "es5",
+        "noImplicitAny": true,
+        "noImplicitReturns": true,
+        "out": "built/common-sim.js",
+        "rootDir": ".",
+        "newLine": "LF",
+        "declaration": true,
+        "sourceMap": false
+    },
+    "files": [
+        "libs/accelerometer/sim/accelerometer.ts",
+        "libs/accelerometer/sim/state.ts",
+        "libs/core/sim/analogSensor.ts",
+        "libs/core/dal.d.ts",
+        "libs/core/sim/buttons.ts",
+        "libs/core/sim/control.ts",
+        "libs/core/sim/core.ts",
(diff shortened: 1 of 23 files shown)


tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+        "noImplicitAny": true,
+        "noImplicitReturns": true,
+        "out": "built/common-sim.js",
+        "rootDir": ".",
+        "newLine": "LF",
+        "declaration": true,
+        "sourceMap": false
+    },
+    "files": [
+        "libs/accelerometer/sim/accelerometer.ts",
```
pelikhan: can we use wildcards here?

riknoll: Nope. That feature was added in TS 2.0 (Fun fact, I added that feature)

riknoll:
I didn't add that enum, I just renamed it. Some of those ID's are in dal.d.ts, some aren't. I'll try and switch to the DAL wherever possible.

tsconfig.json
pelikhan: Confused, we've never had to list .ts file in the past, what's the change here?

pelikhan:
Let's file a bug on getting rid of hardcoded values the CODAL enum

tsconfig.json
riknoll: By default all .ts files are included if no files are listed. We only want the sims files included, so we have to list them explicitly.

riknoll:
Looks like those IDs are in the DAL enum, they're just named differently and have different values. My guess is we've been using the wrong value but have been internally consistent so it's been fine. I'll switch them over.

PR #67, opened 2017-05-31
Build common sim as part of the mock target

riknoll:
Make sure to check that I got the package.json file changes right


riknoll:
Ignore the failed build, it executed before pxt-core was published

PR #68, opened 2017-06-01
More 'pins' topics

ganicke:
* add analog read/write topics. Might expand these as I keep going.

### DIFF
--- a/libs/core/docs/reference/pins/analog-read.md
+++ b/libs/core/docs/reference/pins/analog-read.md
@@ -1,0 +1,34 @@
+# analog Read
+
+Read an analog input value from a pin.
+
+```sig
+pins.A0.analogRead()
+```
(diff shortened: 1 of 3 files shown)


libs/core/docs/reference/pins/analog-read.md
```diff
@@ -0,0 +1,34 @@
...
+## Example #ex1
+
+Use the pixel strip as a signal meter. Read from pin `A2` and display the value as a graph on the pixel
+strip. Also, output the value to the serial port.
+
+```blocks
+loops.forever(() => {
+    let signal = pins.A2.analogRead()
+    light.pixels.graph(signal, 1023)
+    serial.writeValue("signal", signal)
```
pelikhan: graph already writes to serial

PR #70, opened 2017-06-02
Bootloader v1.23 compat

mmoskal:
See https://github.com/Microsoft/pxt-adafruit/issues/171

This will be compatible with bootloader 1.23, but also the earlier versions.

Also, it changes HID usage ID to 0xff97 0x0001 for easier identification in winrt (in line with bootloader 1.23).

Some testing would be in order.

### DIFF
--- a/libs/core/hf2.cpp
+++ b/libs/core/hf2.cpp
@@ -8,6 +8,74 @@
 
 #if CONFIG_ENABLED(DEVICE_USB)
 
+static const char hidDescriptor[] = {
+    0x06, 0x97, 0xFF, // usage page vendor 0x97 (usage 0xff97 0x0001)
+    0x09, 0x01,       // usage 1
+    0xA1, 0x01,       // collection - application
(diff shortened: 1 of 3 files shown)


mmoskal:
OK, everything seems to work. This isn't really critical for the single reset (we overwrite the value in PXT itself anyway), but it's clearer this way. The HID changes are still needed.

libs/core/hf2.cpp
```diff
@@ -8,6 +8,74 @@
 
 #if CONFIG_ENABLED(DEVICE_USB)
 
+static const char hidDescriptor[] = {
+    0x06, 0x97, 0xFF, // usage page vendor 0x97 (usage 0xff97 0x0001)
```
mmoskal: @pelikhan please use `usage 0xff97 0x0001` in addition to the existing `usage 0xff00 0x0001` in the winrt app - this matches the bootloader. Once we think all bootloaders are updated, we can remove the `0xff00`.

PR #80, opened 2017-06-07
Set decompile literals flag

riknoll:
Because of https://github.com/Microsoft/pxt/pull/2235


riknoll:
@microsoftsam @guillaumejenkins  any field editors I'm forgetting to set this on?

PR #94, opened 2017-06-22
Moving to JS random

pelikhan:
https://github.com/Microsoft/pxt/pull/2337

### DIFF
--- a/libs/core/core.cpp
+++ b/libs/core/core.cpp
@@ -467,15 +467,25 @@ TNumber atan2(TNumber y, TNumber x) {
 }
 
 //%
-int random(int max) {
-    if (max == INT_MIN)
-        return -device.random(INT_MAX);
-    else if (max < 0)
-        return -device.random(-max);
-    else if (max == 0)
-        return 0;
-    else
-        return device.random(max);
+TNumber random() {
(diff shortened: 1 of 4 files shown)


mmoskal:
I think you should get two random ints and use bits from both for randomness. You need 53 bits. Otherwise, some of the bits are not really random, and I think it may show in some situations.

For random range with ints it would be faster to use device.random directly.

pelikhan:
> For random range with ints it would be faster to use device.random directly.

Hummm I tought i was already doing that.

mmoskal:
whoops, sorry. But the 53 bits would be nice.

libs/core/core.cpp
```diff
@@ -467,15 +467,38 @@ TNumber atan2(TNumber y, TNumber x) {
...
+        double maxd = toDouble(max);
+        if (mind > maxd) {
+            double temp = mind;
+            mind = maxd;
+            maxd = temp;
+        }
+        if (maxd == mind)
+            return fromDouble(mind);
+        else
+            return fromDouble(mind + device.random(maxd - mind));
```
mmoskal: oh, for example for randomRange(0.5, 3.5) this will return 0.5, 1.5, 2.5 or 3.5, but never say 1.83434.

PR #97, opened 2017-07-17
Edits to colors "indigo" and "purple" in editor

audreyctseng:
Changed "purple" for distinguishment between "pink"; edited color of "indigo" for distinguishment between "purple" and "indigo"


riknoll:
@microsoftsam I think we just have too many colors in the "purple" family; the neopixels can't produce that many shades that appear distinct. I vote we either live with the fact that they look the same or start killing colors.

pelikhan:
Could you add screenshots of the simulator + picture of the physical device on those colors?

PR #103, opened 2017-07-26
Bug 151 fix

caitlin-hennessy:
Resolves https://github.com/Microsoft/pxt-adafruit/issues/151

### DIFF
--- a/libs/light/neopixel.ts
+++ b/libs/light/neopixel.ts
@@ -482,12 +482,11 @@ namespace light {
          * Set the photon color.
          * @param color the color of the photon
          */
-        //% blockId=neophoton_set_color block="photon set pen color %color"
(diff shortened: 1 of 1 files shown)


pelikhan:
We have to check but there are APIs to do basic math on the colors. Using  the hue comes from the LightLogo APIs which makes it easier to cycle through colors rather than use RGB.

caitlin-hennessy:
Waiting on https://github.com/Microsoft/pxt/issues/2742

Need to add second function "setPhotonHue", apply this to scripts with old target version

abchatra:
@caitlin-hennessy what is the status of this PR?

pelikhan:
We've discussed about this and we will have a color wheel color picker on the blocks side.

PR #107, opened 2017-08-15
Enable help for base lib APIs.

ganicke:
- [x] Add block and help jsDoc meta.
- [x] Include jsDoc description for _input_ namespace.


libs/core/control.cpp
```diff
@@ -33,7 +34,8 @@ void raiseEvent(int src, int value, EventCreationMode mode) {
 /**
 * Determine the version of system software currently running.
 */
-//%
+//% blockId="control_device_dal_version" block="device dal version"
```
pelikhan: this is an advanced block with low weight

PR #112, opened 2017-08-29
Inverting y axis in accelerometer

riknoll:
Fixes https://github.com/Microsoft/pxt-adafruit/issues/304

The Y axis of the accelerometer doesn't match the markings on the circuit playground or the simulator behavior. This PR inverts that value and also fixes the simulator tilt up and down events so that they match the hardware.

### DIFF
--- a/libs/accelerometer/accelerometer.cpp
+++ b/libs/accelerometer/accelerometer.cpp
@@ -116,7 +116,7 @@ public:
 
 int InvertableLIS3DH::getX() {
 #if INVERT_ACC_X_AXIS
-    return LIS3DH::getX() * -1;
+    return -LIS3DH::getX();
 #else
     return LIS3DH::getX();
 #endif
@@ -124,7 +124,7 @@ int InvertableLIS3DH::getX() {
 
 int InvertableLIS3DH::getY() {
 #if INVERT_ACC_Y_AXIS
-    return LIS3DH::getY() * -1;
+    return -LIS3DH::getY();
 #else
     return LIS3DH::getY();
 #endif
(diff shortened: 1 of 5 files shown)


pelikhan:
This should be constant that can be overriden per target. It might not be the case for other boards.

riknoll:
Sure. I guess the best way to do that is add a constants.cpp file that can be overridden?

riknoll:
@pelikhan fixed

libs/accelerometer/accelerometer.cpp
```diff
@@ -103,13 +106,46 @@ enum class Gesture {
...
+public:
+    InvertableLIS3DH(codal::mbed::I2C &_i2c, Pin &_int1, uint16_t address = LIS3DH_DEFAULT_ADDR, uint16_t id = DEVICE_ID_ACCELEROMETER, CoordinateSystem coordinateSystem = SIMPLE_CARTESIAN) : LIS3DH(_i2c, _int1, address, id, coordinateSystem) {}
+    int getX();
+    int getY();
+    int getZ();
+};
+
+int InvertableLIS3DH::getX() {
+#if INVERT_ACC_X_AXIS
+    return LIS3DH::getX() * -1;
```
pelikhan: ``- LIS3DH::getX()`` ?

riknoll:
@pelikhan yup, tested. With https://github.com/Microsoft/pxt-adafruit/pull/305 the simulator and hardware now match.

riknoll:
BTW, this is a breaking change. I'll merge it and do the required releases tomorrow

PR #114, opened 2017-09-24
Move animation queue to core to fix adafruit build

microsoftsam:


mmoskal:
I already ignore build errors in docs, do the build should be fixed now. 

Animation queue is a concept shared between targets that do.not take core from here. This will then lead to duplication of code.

microsoftsam:
Okay.

PR #118, opened 2017-10-02
Refactor blocks to remove subcategories, and order some in groups.

microsoftsam:
Lights: 
- I haven't made any changes yet since we still need to discuss how we're going to refactor the complex APIs perhaps into a package, and make it all work together.

Input:
screen shot 2017-10-02 at 1 24 09 am
screen shot 2017-10-02 at 1 24 15 am

Music:
screen shot 2017-10-02 at 1 23 59 am

Note: I made the set Tone block hidden, block usage on that was extremely low, and given it's complex usage I'm not convinced it needs to be a block.

Math: 
screen shot 2017-10-02 at 1 26 54 am

Pins:
screen shot 2017-10-02 at 1 23 44 am
screen shot 2017-10-02 at 1 23 48 am

Serial:
screen shot 2017-10-02 at 1 27 04 am

### DIFF
--- a/libs/accelerometer/accelerometer.cpp
+++ b/libs/accelerometer/accelerometer.cpp
@@ -222,7 +222,7 @@ int acceleration(Dimension dimension) {
 //% help=input/rotation
 //% blockId=device_get_rotation block="rotation (°)|%NAME"
 //% parts="accelerometer"
-//% weight=38 blockGap=8
+//% group="More" weight=38
 int rotation(Rotation kind) {
     switch (kind) {
     case Rotation::Pitch:
@@ -241,7 +241,7 @@ int rotation(Rotation kind) {
 //% blockId=device_set_accelerometer_range block="set accelerometer|range %range"
 //% weight=5
 //% parts="accelerometer"
-//% weight=15 blockGap=8
+//% group="More" weight=15 blockGap=8
 void setAccelerometerRange(AcceleratorRange range) {
     getWAccel()->acc.setRange((int)range);
 }
--- a/libs/accelerometer/accelerometer.cpp
+++ b/libs/accelerometer/accelerometer.cpp
@@ -163,12 +163,13 @@ namespace input {
  * @param gesture the type of gesture to track, eg: Gesture.Shake
  * @param body code to run when gesture is raised
  */
-//% help=input/on-gesture weight=98 blockGap=8
+//% help=input/on-gesture
 //% blockId=device_gesture_event block="on |%NAME"
 //% parts="accelerometer"
 //% gesture.fieldEditor="gridpicker"
 //% gesture.fieldOptions.width=220
 //% gesture.fieldOptions.columns=3
+//% group="Accelerometer" weight=77 blockGap=8
 void onGesture(Gesture gesture, Action body) {
     auto acc = &getWAccel()->acc;
     acc->updateSample();
@@ -193,12 +194,13 @@ int getAccelerationStrength() {
  * x=0, y=0 and z=-1023)
  * @param dimension TODO
  */
-//% help=input/acceleration weight=76
-//% blockId=device_acceleration block="acceleration (mg)|%NAME" blockGap=8
+//% help=input/acceleration
+//% blockId=device_acceleration block="acceleration (mg)|%NAME"
 //% parts="accelerometer"
 //% dimension.fieldEditor="gridpicker"
 //% dimension.fieldOptions.width=180
 //% dimension.fieldOptions.columns=2
+//% group="Accelerometer" weight=76 blockGap=8
 int acceleration(Dimension dimension) {
     switch (dimension) {
     case Dimension::X:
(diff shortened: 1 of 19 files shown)


Jaqster:
This looks good.  
Possible categories for Math: Operators, Functions, Transformations, Values.  
For the Pins, it's confusing when a block is Digital, but yet we offer the A0 as an option.  Can we constrain the field picker values to only show either Analog or Digital pins depending on the block type?  Then we could have "Digital" and "Analog" subcategories.
Also, can we add Serial Read API's for the Win10 app version?  Not too sure of the scenario here, but if we're able to read from a serial port in the app, that might be cool functionality to add.

microsoftsam:
I was under the impression we didn't want to add any groupings for Math @Jaqster. Any categorisation for Math would have to be done both on the pxt side and common-packages in this case, so it would affect other editors as well.

Jaqster:
Okay.  There's few enough blocks that we don't need it.

pelikhan:
Pins are called A0, A1, ... and all of them can be used as a digital pin (some can do PWM).

The categories have a mix of hardware component name (Accelerometer) and concept (Sound). We should pick one and remain consistent.

microsoftsam:
I hear your second point @pelikhan. Which do you think we should stick with. I couldn't come up with the concept for "Accelerometer", but can easily add "Microphone" or "Sound sensor" for Sound.

pelikhan:
I think correlating the blocks to the hardware component that enables them has good educational value so that would be my preference.

abchatra:
I am not in favor of categorization in this way for "Input". Most used or kid friendly blocks should be at the top. With the current categorization blocks which are not useful are 2nd and 3rd blocks (button A is pressed and button A was pressed). 

Also I don't see how kids understand accelerometer. Can we categorize based on events? All events are at the top and rest are categorized by most commonly used

microsoftsam:
The nice thing of the labels is it clearly separates the blocks into clear sub categories in the flyout. The issue with having all the events in one place is that you would have to scan through all of them and read the block in order to figure out what that block does and find the block you're looking for that way.
I don't mind adding events in one place but we should be consistent about it. ie: add an Events category like scratch.

abchatra:
I am not sure our categorization makes any sense for students. Events are self explanatory. Instead of ready category names one can read events itself

microsoftsam:
There are too many blocks to scan, it's true that over time students begin to read more of the blocks and understand what they do, and remember where they are. But for a beginner without the different sections there are just too many blocks to scan. 
The concept of an Event, is also something that's not taught till later through a comp course, so it's not necessarily the most basic concept. 

Any sort of categorisation of blocks is going to spark a difference of opinion and I acknowledge that. The things to keep in mind, is subcategories (in the toolbox, the old way), make navigating the toolbox difficult and we're staying away from that. This means we need to have all our input blocks in the same category, so by design you're going to end up with more blocks in the flyout. And ultimately adding sections in the flyout is what this is trying to solve. 
If you added all the Event blocks together and called that section "Events", what would you call all the reporter blocks? and what happens to the random configuration blocks (that configure the temp / acceleration / sensor thresholds). 
My argument for having them categorised by component is that all the blocks that are related to that specific component are near each other, and have a heading that describes it.

abchatra:
I acknowledge Events are not a good name either. I don't see how adding more text (category name) and unnecessary blocks will help students either. 

Here we are making an assumption that students understand the category name and skip the category all together if that is what they are not looking far. I am not sure that assumption is right. 

As per Andy kids just scan the blocks top to bottom and any kind of categorization which is fairly obvious is not going to help them.

microsoftsam:
Definitely something we need to test this week. There is clear navigational structure with adding the label. Whether that helps kids or not, this hasn't been validated yet, but something we should figure out ASAP before adding groupings to all categories.

abchatra:
My opinion is to do the "Input" categorization based on merging more blocks at the bottom of the current input blocks. 

Until we have data which shows that this categorization is better for students we should go ahead with this.

microsoftsam:
What do you mean by "merging more blocks at the bottom of the current input blocks".

pelikhan:
You add a “More” separator and stuff the more blocks under it.

microsoftsam:
Okay so Input gets the "More" category. What about others? 
I really like the categorisation of blocks under Music and Pins.

pelikhan:
We should have a consistent category strategy. In music, the note and tempo block should not be on top as they are shadow blocks where needed.

abchatra:
Your categorization of pins look good to me. They are advanced blocks and one who is looking for them knows what to look for.

microsoftsam:
Music: 
screen shot 2017-10-03 at 1 45 12 am

Sounds first (most used)
Tones next
Volume
then Tempo

Gaps between each set of blocks as pointed out in the screenshot. @pelikhan @abchatra

microsoftsam:
Input: 

screen shot 2017-10-03 at 1 58 08 am

screen shot 2017-10-03 at 1 58 14 am

All input handlers are together and on top. 
All reporter blocks together and below
All configuration blocks at the end and together.

microsoftsam:
Can I get this in @pelikhan @abchatra. Please check the latest screenshots for Music and Input categories.

PR #119, opened 2017-10-02
support for threshold/value blocks for touch buyttons

pelikhan:
Support for set threshold and reading live capacitive value from touch button


pelikhan:
@mmoskal coudl you double check that i got the instance stuff correct?

PR #122, opened 2017-10-15
Create an I2c pacakge

microsoftsam:
Refactor i2c blocks and code into their own package


microsoftsam:
- They are advanced blocks
- They require a component
- Components should be (optional) packages

pelikhan:
You need to logic to fix existing scripts that use i2c.

PR #123, opened 2017-10-15
Create a servo package

microsoftsam:
Refactor servo blocks into their own package


pelikhan:
We need logic to patch older scripts that use this API.

PR #125, opened 2017-10-16
Reverti2cservo

pelikhan:
Reverting refactoring of servo/i2c as they are API breaking changes. We need to review this and decide whether it is worth doing the change.


mmoskal:
@pelikhan @microsoftsam I think pretty much all boards that have pins, will have I2C, SPI, and PWM (servo). I tend to think of the `core` package as really the `codal-core` package. If a target doesn't use Codal (eg for an embedded Linux), then it only uses `base` and implements its own `core`.

PR #128, opened 2017-10-19
Multi-strip simulator

pelikhan:
Correct simulator support for neopixel strips on multiple pins. The current support assumed a single strip which meant we were animating LEDs on board when using external light strips.


microsoftsam:
Nice

PR #132, opened 2017-10-19
support for hue in photon

pelikhan:
Add a photon API that uses a color hue instead of RGB. Hue is needed to do cool math on colors. RGB does not cut it. The wording is changed so that there is no more confusion about this.

image

### DIFF
--- a/libs/light/neopixel.ts
+++ b/libs/light/neopixel.ts
@@ -506,7 +506,6 @@ namespace light {
         //% parts="neopixel" deprecated=1 blockHidden=true
         //% group="Photon" weight=39 blockGap=8
         setPhotonPenHue(hue: number) {
-            hue = (hue >> 0) % 0xff; // cycle through colors
             this.setPhotonPenColor(hsv(hue, 0xff, 0xff));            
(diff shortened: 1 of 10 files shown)


libs/light/neopixel.ts
```diff
@@ -477,11 +501,17 @@ namespace light {
         /**
          * This function is deprecated.
          */
-        //% blockId=neophoton_set_color block="%strip=variables_get|photon set pen color %color=colorWheelPicker"
+        //% blockId=neophoton_set_color block="%strip=variables_get|photon set pen hue %hue=colorWheelPicker"
+        //% help="light/set-photon-pen-hue"
         //% parts="neopixel" deprecated=1 blockHidden=true
```
microsoftsam: Why is this deprecated and hidden? Did you mean that?

pelikhan: That's a bug.

microsoftsam: Fixed in https://github.com/Microsoft/pxt-common-packages/pull/133

PR #133, opened 2017-10-19
Fix old light blocks

microsoftsam:
Give all new blocks new ids.
Bring back the old blocks with their old ids, and old definitions (adding the mutator dialog). 
If an old script is loaded, they will see the default instance mutator icon. 
Give old blocks new function names that start with an underscore (which is hidden from the monaco editor as of https://github.com/Microsoft/pxt-monaco-typescript/pull/8). And wrap their implementation.


libs/light/neopixel.ts
```diff
@@ -499,11 +499,11 @@ namespace light {
         }
 
         /**
-         * This function is deprecated.
+         * Sets the photon hue.
          */
-        //% blockId=neophoton_set_color block="%strip=variables_get|photon set pen hue %hue=colorWheelPicker"
+        //% blockId=neopixel_set_color block="%strip=variables_get|photon set pen hue %hue=colorWheelPicker"
         //% help="light/set-photon-pen-hue"
-        //% parts="neopixel" deprecated=1 blockHidden=true
```
pelikhan: was this by design or did i mess up the merge? setPhotonPenHue is new and needs to stay.

microsoftsam: Looks like a merge issue. Will fix

microsoftsam: Just to clarify, you want this block deprecated correct?

microsoftsam: I think in this case, there's no need for this as we will attach the neophoton_set_color block to the __setPhotonColor function and get it to call setPhotonPenColor.

microsoftsam:
Fixed merge, and removed the block around setPhotonPenHue as it's not necessary. Please verify and approve.

libs/light/neopixel.ts
pelikhan: No, photon needs hue - it does not work well with color. So setPhotonPenHue is the block that is used.

pelikhan:
setPhotonPenHue is the API to change colors on the photon. RGB does not cut it. It is too cumbersome to use. See https://github.com/Microsoft/pxt-common-packages/pull/132

libs/light/neopixel.ts
microsoftsam: Okay, but you set the block as deprecated and hidden. Did you mean to do that? 
Are you wanting to ship two blocks, one for pen color, one for pen hue?

microsoftsam:
See my comment on that PR, I think you might have accidentally set that API as hidden and deprecated

libs/light/neopixel.ts
pelikhan: Let's see: ``setPhotonColor`` is the old API with a confusing name which stays for legacy. ``setPhotonPenHue`` is the one that's useful so that's the one to add. The ``setPhotonPenColor`` one, I'm not sure - in doubt we can remove it.

microsoftsam: Okay so we need to remove the deprecated flag on the setPhotonPenHue and give it a new ID that does not need the __ hack.

microsoftsam: Fixed, please verify

pelikhan:
Let's see how far we go with this.

PR #138, opened 2017-10-21
Adjust light docs paths and signutures

ganicke:
- [x] Move docs for **NeoPixelStrip** class down level to _/light/neopixelstrip/_.
- [x] Fix signatures in the cards for _light.md_
- [x] Change help paths for the **NeoPixelStrip** methods.
- [x] Surface **defaultPin** from _neopixel.cpp_.
- [x] Fix **See also** links to _neopixelstrip_ API topics.
- [x] Remove **pixels** instance from signatures and examples.
---
|Dependencies|...|
| --- | --- |
| **Related issue** | https://github.com/Microsoft/pxt-adafruit/issues/387 |
| **Downstream PR** | https://github.com/Microsoft/pxt-adafruit/pull/409 |


libs/light/docs/reference/light.md
```diff
@@ -40,8 +40,8 @@ light.createStrip().setPhotonMode(PhotonMode.PenUp)
 light.rgb(255, 255, 255);
 light.hsv(255, 255, 255);
 light.colors(Colors.Red);
-//light.fade(0, 0)
-//light.animation(LightAnimation.Rainbow)
+light.fade(0, 0)
+light.animation(0)
```
pelikhan: this one is legacy, let's remove it from the docs

ganicke: Ok. 

Also, are we good with the layout having the strip related APIs under _/light/neopixelstrip/*_?

PR #140, opened 2017-10-23
Support for multiple boards in a single target

mmoskal:
* pins and buttons are now identified by hardware ID (allowing for redefinition from TS)
* the `PXT_BOARD_ID`, `devpins.h/cpp`, and `PIN_*` definitions are no more; boards are defined in TS
* the SWD-pins used for neopixels (as in Metro) are now handled correctly
* the `dal.d.ts` file is cleaned up (using `dalDTS` config in `pxt.json`)

This requires changes in targets.


mmoskal:
After merging would be good to bump major version, since this is a breaking change
Also, this doesn't include wiring - but probably best to first merge this

pelikhan:
Wiring?

mmoskal:
By wiring I mean the rendering of breadboard on the screen. This is missing here - this is mostly about the C++ code.

pelikhan:
Could you bump the major version in PR?

mmoskal:
Bumped. Note that this requires https://github.com/Microsoft/pxt/pull/3309

libs/buttons/sim/buttons.ts
```diff
@@ -55,6 +55,7 @@ namespace pxsim {
     export class CommonButtonState {
         usesButtonAB: boolean = false;
         buttons: CommonButton[];
+        buttonsByPin: Map<CommonButton> = {};
```
riknoll: You can use `{[index: number]: CommonButton}` instead of a `Map` to avoid converting to a string below.

PR #141, opened 2017-10-23
don't call show() inside animation

pelikhan:
This is a change to the animation API so that animation don't call show on their own. The animation engine should be in charge of calling show.

This allows to run multiple animations at once and compose them.


pelikhan:
@microsoftsam review?

libs/light/neopixel.ts
```diff
@@ -568,7 +573,12 @@ namespace light {
         //% weight=87 blockGap=8
         showAnimationFrame(animation: NeoPixelAnimation) {
             if (!animation) return;
-            animation.showFrame(this);
+            const buf = this.buffered();
+            this.setBuffered(false);
+            animation.renderFrame(this);
+            this.setBuffered(buf);
+            this.show();
+            loops.pause(50);
```
microsoftsam: No more variable delay between animation frames?

pelikhan:
Actually showAnimationFrame should probably not have a delay in it. Each animation should handle timing internally and let the animation engine call them when needed.

libs/light/neopixel.ts
pelikhan: removed

PR #142, opened 2017-10-24
LightAnimation enum to constants

microsoftsam:
Use constant namespace property attributes instead of enum for LightAnimation. 

Convert LightAnimation from an enum to have it's values as namespace properties. 
The editor should tree shake and not include the instances of these animations unless they're used. 

Relies on https://github.com/Microsoft/pxt/pull/3307

### DIFF
--- a/libs/light/neopixel.ts
+++ b/libs/light/neopixel.ts
@@ -69,27 +69,6 @@ enum LightMove {
     Shift
 }
 
-enum LightAnimation {
-    //% blockImage=1
-    //% block="rainbow"
-    Rainbow,
-    //% blockImage=1
-    //% block="running lights"
-    RunningLights,
-    //% blockImage=1
-    //% block="comet"
-    Comet,
-    //% blockImage=1
-    //% block="sparkle"
-    Sparkle,
-    //% blockImage=1
-    //% block="theater chase"
-    TheaterChase,
-    //% blockImage=1
-    //% block="color wipe"
-    ColorWipe
-}
-
 /**
  * A determines the mode of the photon
  */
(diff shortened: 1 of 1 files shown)


libs/light/neopixel.ts
```diff
@@ -1354,3 +1327,24 @@ namespace light {
         }
     }
 }
+
+namespace LightAnimation {
+    //% blockImage=1
+    //% block="rainbow" whenUsed
+    export const Rainbow = new light.RunningLightsAnimation(0xff, 0, 0, 50);
```
pelikhan: The problem with this design is that animations are singleton so if you run the same animation on multiple LED strips they will interfere. It actually has to be a factory function (which is better for laziness).

libs/light/neopixel.ts
```diff
@@ -1153,7 +1126,7 @@ namespace light {
         public showFrame(strip: NeoPixelStrip): void { }
     }
 
-    class RainbowCycleAnimation extends NeoPixelAnimation {
+    export class RainbowCycleAnimation extends NeoPixelAnimation {
```
mmoskal: I think you can just use: `export const Rainbow: NeoPixelAnimation = new RainbowCycleAnimation(...)` without `export` here. But maybe it's good to be able to derive them.

libs/light/neopixel.ts
mmoskal: The laziness is similar with the `whenUsed` attribute. You could still use the same object and maybe have a number or two for animation on the strip.

pelikhan:
We are now using fixedInstances for this.

PR #145, opened 2017-10-25
Rename serial blocks

caitlin-hennessy:


riknoll:
We were discussing this at standup. The API is not obvious and every time someone tries to write a string or a number they drag out `writeString` or `writeNumber` and are confused when it doesn't "work". In my opinion, we shouldn't have low-level APIs that need explaining like that in the blocks.

@microsoftsam @abchatra @Jaqster

pelikhan:
Then the proper fix is to move out serial into a package. This category is pretty much useless without those.

abchatra:
Can we keep the low api in typescript & remove from blocks? 

Also light.graph writing to serial doesn't make sense to me. It is not obvious that it writes to serial. There is no documentation which says so. And if user writes explicit serial.write then light.graph interferes with it.

caitlin-hennessy:
You can do with writeLine everything you can do with writeString and writeNumber, so the category isn't any less useful.

pelikhan:
It just works: in micro:bit when you do a plotbargraph, it automatically displays the data stream. You don't have to read the docs to get it working.

If you remove this feature from light.graph, users will have to discover that feature by accident in the advanced -> serial category.

pelikhan:
If we are about cleaning up, we should not call this category 'serial' because it is not (it is data piping over HF2). It is not compatible with serial consoles like Putty or TerraTerm.

I am proposing to remove serial all together and add the blocs needed to surface data logging.

caitlin-hennessy:
@pelikhan so your suggested TODO items are:
- Rename the "serial" category to "data logging"
- Add other blocks for data logging (what would these blocks be?)

pelikhan:
I'm not sure it should live under advanced -> data which is pretty much undiscoverable. Maybe it would work better if we were able to "graph" a variable and have that block sit under variable.
We should design the overall data-logging experience and have a block that supports it. Having a new block deep down in the advanced section won't help.

caitlin-hennessy:
@pelikhan That's fine with me, but it sounds like a change that is separate from this PR.  Can we merge this now, and figure out where data logging blocks should live later?

pelikhan:
This PR is introducing a new block "plot value".

pelikhan:
Happy to sit F2F monday and discuss about this. This fix feels like bandaid. We are removing existing blocks without updating the docs, renaming a block which now does not match the typescript version (we do that in other places, but we should try to avod it) and we have a category that is still not named properly.
If we need to fix serial, we should pull out all the blocks.

caitlin-hennessy:
It's not a new block, it's a new label on an old block.  Changing the label to "plot value" instead of "write value" makes sense now that we have plotting.

pelikhan:
All blocks in that category are consistent with the ``write`` terminology and provide access to IO over serial (which is not really serial). If we need a new kid friendly block to plot, it should not go under the serial category.

pelikhan:
There is no obvious connection with ``serial`` and the new data logging support. If we really want kids to discover and use that feature, we need better support for it in the API/block editor. Renaming the serial block does not solve that issue IMO.

caitlin-hennessy:
The intent of this PR isn't to solve that issue - just to take a step in that direction by removing/renaming obsolete/confusing blocks.  We can do more cleanup and reorganization later - that shouldn't block this PR though.

abchatra:
Lets wait till Peli is back for this PR. We have time to merge next week.

abchatra:
I do agree with Caitlin on this.

pelikhan:
We are shipping a v1 so if we think serial is confusing, we should remove it all together. We can/should not take it back afterwards.

caitlin-hennessy:
@pelikhan @riknoll Ping for review of new changes

libs/base/console.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference no-default-lib="true"/>
+
+/**
+ * Reading and writing data over a serial connection.
```
pelikhan: « to the console output »

libs/base/console.ts
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Write a name:value pair as a line of text to the serial port.
+     * @param name name of the value stream, eg: "x"
+     * @param value to write
+     */
+    //% weight=88 blockGap=8
+    //% help=console/log-value
+    //% blockId=console_log_value block="console|log value %name|= %value"
+    export function logValue(name: string, value: number): void {
+        serial.writeString(name + ":" + value + "\r\n");
```
pelikhan: use serial.writeValue

libs/base/console.ts
pelikhan: Disregard this one.

pelikhan:
We need a round of docs update to reference those from the reference -> console path

pelikhan:
@ganicke

abchatra:
Thanks Caitlin

ganicke:
Ok, will add it in. It should be very consoling.

Jaqster:
Ha!  Good one Galen!

ganicke:
RE: #178.

Maybe change to a stream abstraction in the future to allow 'console' output to go to a sim window, LEDs, etc.

pelikhan:
Yup, that’s the idea.

abchatra:
@ganicke  do you need a tracking bug for serial->console blocks update?

ganicke:
Yea, it will help my metrics...just kidding.

PR #146, opened 2017-10-26
Fix some wording in 'simulator.md'

ganicke:


libs/core/docs/device/simulator.md
```diff
@@ -1,6 +1,6 @@
 # Simulator
 
-The JavaScript simulator allows to test and execute most @boardname@ programs in the browser.
-It allows to emulate sensor data or user interactions.
```
pelikhan: We should use ``simulate`` instead of ``emulate``. It is not 100% faithful.

ganicke: OK. Can't add 'display' or 'show lights' since not all boards will have them. Maybe add ```## #board-specifc```

PR #148, opened 2017-10-26
Lazy per-pixel brightness

pelikhan:
In most scenario, the user has a single LED strip and the brightness buffer is not necessary: we can apply the brightess value when rendering.
This change optimizes for that case to retain memory in the common case.


microsoftsam:
Looks good. Can we pass an optional parameter to setPixelColor with the brightness as well?

pelikhan:
Thinking that this should probably be a separate function. Currently, setPixelColor / pixelColor is symmetric. This would break that.

PR #150, opened 2017-10-26
Add cable package image

microsoftsam:
Add cable image. The image lives in the target though


pelikhan:
Will this work in the cloud?

microsoftsam:
Not sure I want to try

PR #151, opened 2017-10-26
Update IR and cable blocks

microsoftsam:
Split IR handlers into three handlers, only surfacing the one sending a single number. 
Did the same for cable. 
Added JSdoc and updated IR documentation

Tested with two Circuit playgrounds. (IR is very finicky)

screen shot 2017-10-26 at 11 23 21 am

Fixes https://github.com/Microsoft/pxt-adafruit/issues/392

### DIFF
--- a/libs/cable/cable.ts
+++ b/libs/cable/cable.ts
@@ -42,18 +42,38 @@ namespace network {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


pelikhan:
We should also have equivalent send*** functions

PR #155, opened 2017-10-31
moving to fixed instances for animations

pelikhan:
Work in progress to migrate animation to use fixedInstances

- [x] Refactor animation class to create "renderers"
- [x] support for showAnimationFrame
- [x] support for image drop down for fixed instances using ``jres`` resource


pelikhan:
@microsoftsam @riknoll can you add a field editor that uses jres to load images? it is not support for fixedInstances drop downs.

pelikhan:
@microsoftsam ready for review

PR #164, opened 2017-11-06
Fix sim shims

microsoftsam:
Fixes https://github.com/Microsoft/pxt-adafruit/issues/461

```
missing in sim: pxsim.BufferMethods.toHex
missing in sim: pxsim.serial.writeBuffer
missing in sim: pxsim.pxtcore.lookupComponent
missing in sim: pxsim.pxtcore.getMultiButton
missing in sim: pxsim.DigitalPinMethods.pushButton
missing in sim: pxsim.AnalogPinMethods.touchButton
```

### DIFF
--- a/libs/base/sim/control.ts
+++ b/libs/base/sim/control.ts
@@ -1,6 +1,6 @@
 namespace pxsim.control {
-    export var runInBackground = thread.runInBackground;
-    export var delay = thread.pause;
+    export let runInBackground = thread.runInBackground;
+    export let delay = thread.pause;
 
     export function reset() {
         U.userError("reset not implemented in simulator yet")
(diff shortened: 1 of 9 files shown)


riknoll:
Why not just add an implementation for `toHex` and `getMultiButton`? I'm pretty sure the sim will still crash if we use those methods and they return undefined so there isn't really any point to adding placeholder functions.

microsoftsam:
I dont know why toHex has a shim for it whereas the other buffer functions dont have it, and I dont want to implement multibutton as I’m not familiar with were its used. It still crashes today, at least now we have a TODO

riknoll:
Well, we shouldn't mask the error until we're ready to have a real implementation. At the very least throw a user error like we do in `control.reset()`

PR #167, opened 2017-11-08
add plumbing for event flags

tballmsft:
CODAL supports executing an event under different concurrency/queueing models.  This change exposes the CODAL events flags as a default parameter.  No change of behavior for existing targets.  For example of use, see 

https://makecode.com/_1xWL07DrvM23


tballmsft:
we can't use the enum name because of cpp.ts limitations (and other things, I imagine)

PR #190, opened 2017-11-29
Allow targets to override melody definitions

cait1inhennessy:
Also rename "piano.ts" to "music_overrides.ts" - more descriptive.


microsoftsam:
I don't think music_overrides is a good name necessarily. You're using it in this case to override some definitions, but the file should be descriptive of it's contents.

pelikhan:
This PR is just moving code around. Why do we need this?

pelikhan:
Can we fix this melody for all targets instead having to override it?

pelikhan:
@cait1inhennessy what's the status of this PR?

cait1inhennessy:
This issue was resolved by a different PR.

PR #194, opened 2017-12-11
Add pause until block support as a top level block

microsoftsam:


libs/base/loops.ts
```diff
@@ -37,7 +37,7 @@ namespace loops {
         _pollEventQueue = undefined;
     }
 
-    function queuePollEvent(timeOut: number, condition: () => boolean, handler: () => void) {
+    export function __queuePollEvent(timeOut: number, condition: () => boolean, handler: () => void) {
```
pelikhan: let's move this whole mess in the control namespace instead.

PR #198, opened 2017-12-26
Fix type of control.deviceDalVersion in docs.

chaw:
A minor fix to the docs:
As implied by the original example, the return type is string, not number.


pelikhan:
Thanks!

PR #203, opened 2018-01-05
Add PI, E and other Math constants

samelhusseini:
Needed for more interesting animations

### DIFF
--- a/libs/base/math.ts
+++ b/libs/base/math.ts
@@ -1,13 +1,13 @@
 namespace Math {
 
-    export let E = 2.718;
-    export let LN2 = 0.693;
-    export let LN10 = 2.303;
-    export let LOG2E = 1.443;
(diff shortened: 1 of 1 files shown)


libs/base/math.ts
```diff
@@ -1,4 +1,14 @@
-namespace Math {    
+namespace Math {
+
+    export let E = 2.718;
+    export let LN2 = 0.693;
+    export let LN10 = 2.303;
+    export let LOG2E = 1.443;
+    export let LOG10E = 0.434;
+    export let PI = 3.14159;
```
pelikhan: We need the full precision used in JavaScript.

samelhusseini: I updated the values with the values returned from a Chrome console. Is that what you meant?

samelhusseini:
@pelikhan updated values

PR #204, opened 2018-01-16
adding a fork block

pelikhan:
Allows to fork execution from within a code block -- rather than a top level block like control.runInBackground.


riknoll:
LGTM. Only question is if we want to use the name "fork"

pelikhan:
It could be "run in parralel" or "start task". @tballmsft @abchatra @mmoskal POV?

mmoskal:
How about changing run in background not to be a top level block? I think it would make more sense and the name is most descriptive.

pelikhan:
It would be a breaking change for blocks.

PR #209, opened 2018-01-23
Runinparallel

pelikhan:
- [x] add control.runInParralel statement block
- [x] deprecate control.runInBackground
image


libs/base/docs/reference/control/run-in-background.md
```diff
@@ -1,81 +1,3 @@
...
-    }
-})
-spinit = true;
-for (let i = 0; i < 5; i++) {
-    loops.pause(1000);
-    pixels.setPixelColor(0, Colors.Blue);
-}
-```
\ No newline at end of file
+This function is deprecated and replace by [run in parallel](/reference/control/run-in-parallel).
```
riknoll: replace -> replaced

libs/base/control.cpp
```diff
@@ -43,12 +43,12 @@ namespace control {
...
 
     /**
-     * Run other code in the background.
+     * Run other code in the parallel.
      */
-    //% help=control/run-in-background blockAllowMultiple=1 afterOnStart=true
-    //% blockId="control_run_in_background" block="run in background" blockGap=8
-    void runInBackground(Action a) {
-      pxt::runInBackground(a);
+    //% help=control/run-in-parallel handlerStatement=1
```
riknoll: Are we cool with C++ breaking change? Not sure if this actually impacts any targets but it could.

libs/base/docs/reference/control/run-in-parallel.md
```diff
@@ -0,0 +1,81 @@
...
+
+Run some other code at the same time that your main program code runs.
+
+```sig
+control.runInParallel(() => {})
+```
+
+Sometimes you want your program to work on more than one thing at a time. The main part of your program is
+always put in [``||on start||``](/blocks/on-start). But, you can also put some other part of your
+program in ``||control:run in background||``. This is a block that runs code seperately from the code
```
riknoll: still says run in background (also, below)

libs/base/control.cpp
pelikhan: This is not a documented API so yes.

PR #212, opened 2018-01-25
Support for gamepad

pelikhan:
Partial support (buttons only) for gamepad. Throttle and move crash the device.
FYI @deanm1278


deanm1278:
@pelikhan sorry looks like I had a couple local changed I forgot to push. Maybe joystick would work w/ most recent commit here:
https://github.com/adafruit/codal-core

PR #223, opened 2018-02-10
Only enable serial on webusb after we got some command

mmoskal:
This seems to fix issues with WebUSB serial lockups - https://github.com/Microsoft/pxt-adafruit/issues/632


mmoskal:
Also, with this one in, we seem to be getting serial data in the web app in Chrome

pelikhan:
Woot! Will test as well

pelikhan:
We handle the device connect events so we could add this next week. Isn't the that ping event?

PR #227, opened 2018-02-14
Make forever and pause be top level

riknoll:
Moving both of those functions to be top-level, just like `pauseUntil()`. In other words:
1. `loops.forever(() => {})` is now `forever(() => {})`
2. `loops.pause(500)` is now `pause(500)`

This shouldn't break anybody. The new functions I defined have the same block id as the old functions so hopefully old block scripts will be automatically converted when we compile to TS.

The old functions are still there so that we don't break TS scripts, but they have `deprecated=true` and I changed the block ids.

Adding DO NOT MERGE because I want to make sure everyone is aware of this change before I merge it in.

### DIFF
--- a/libs/base/control.ts
+++ b/libs/base/control.ts
@@ -69,7 +69,7 @@ namespace control {
             while (this.running
                 && !this.isCancelled(evid)
                 && render()) {
-                loops.pause(this.interval);
+                pause(this.interval);
             }
 
             // check if the animation hasn't been cancelled since we've been waiting
@@ -132,7 +132,7 @@ namespace control {
                     }
                 }
             }
-            loops.pause(50);
+            pause(50);
         }
         // release fiber
         _pollEventQueue = undefined;
--- a/libs/base/control.ts
+++ b/libs/base/control.ts
@@ -11,7 +11,7 @@ namespace control {
     export function runInBackground(a: () => void) {
         control.runInParallel(a);
     }
-    
+
     /**
      * Display an error code and stop the program.
      * @param code an error number to display. eg: 5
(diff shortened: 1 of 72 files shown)


mmoskal:
I can't even tell you how glad I am to see `pause` leave `loops`. It always twisted my mind in weird ways...

libs/base/control.ts
```diff
@@ -178,6 +178,27 @@ function pauseUntil(condition: () => boolean, timeOut?: number): void {
...
+
+/**
+ * Pause for the specified time in milliseconds
+ * @param ms how long to pause for, eg: 100, 200, 500, 1000, 2000
+ */
+//% help=loops/pause weight=99
+//% async block="pause %pause=timePicker|ms"
+//% blockId=device_pause blockNamespace="loops"
+function pause(ms: number): void {
+    loops.pause(ms);
```
pelikhan: shouldn it be the other way around? loops.pause calls into pause?

libs/base/loops.cpp
```diff
@@ -16,9 +16,9 @@ void forever(Action a) {
  * Pause for the specified time in milliseconds
  * @param ms how long to pause for, eg: 100, 200, 500, 1000, 2000
  */
-//% help=loops/pause weight=99
+//% help=loops/pause weight=99 deprecated=true
 //% async block="pause %pause=timePicker|ms"
-//% blockId=device_pause
+//% blockId=device_pause_deprecated
```
samelhusseini: If you change the block id, it's no longer backwards compatible..

samelhusseini: oh nevermind, you have device_pause elsewhere.

samelhusseini:
Question RE highlighting? I don't see anything setting the highlight colors of that. Most top level blocks use the standard monaco blue (eg: while, for, etc). I would say we highlight those with the same color. You can add that in monaco.tsx:defineEditorTheme

riknoll:
It should probably use the loops color right? Just whatever we set as blockNamespace since that will be the color in the blocks

pelikhan:
Yes loops color to match the category color.

samelhusseini:
I'm on the fence on that. On the one hand you want it to appear as a lang construct, hence why it's top level. 
On the other hand, it's in loops in Blockly, so maybe should match color there.. @others?

pelikhan:
We have a convention that all blocks under a category have the same color, let's keep that. I know that Chibi already has an exception.

samelhusseini:
All good.

PR #228, opened 2018-02-24
Tag cp doc sections for overrides - 01

ganicke:
First batch of mods for CP docs for sections to override in targets.

- [x] Add section ids for 'Example' and 'See also'.
- [x] Remove block styling from 'See also' links.
- [x] Update 'acceleration' topic.

### DIFF
--- a/libs/accelerometer/docs/reference/input/acceleration.md
+++ b/libs/accelerometer/docs/reference/input/acceleration.md
@@ -1,6 +1,6 @@
 # acceleration
 
-GGet the acceleration value (milli g-force) in one of three dimensions, or the combined force in all directions (x, y, and z).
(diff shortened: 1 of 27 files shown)


libs/accelerometer/docs/reference/input/acceleration.md
```diff
@@ -1,47 +1,65 @@
-# Acceleration
+# acceleration
 
-Get the acceleration value (milli g-force), in one of the three dimensions.
+GGet the acceleration value (milli g-force) in one of three dimensions, or the combined force in all directions (x, y, and z).
```
pelikhan: Gget

ganicke: Ah, ugh!

PR #230, opened 2018-02-24
Bresenham's line algorithm

pelikhan:


libs/screen/image.cpp
```diff
@@ -571,6 +571,51 @@ void _drawIcon(Image img, Buffer icon, int xy, int c) {
     decrRC(ii);
 }
 
+// Bresenham Algorithm
+//%
+void _drawLine(Image img, int xy0, int xy1, int c) {
```
mmoskal: After all I did take a stab at the line, since I had time after integrating into 32, see https://github.com/Microsoft/pxt-common-packages/commit/e1716db5b0282b18b7c53dd68a87b11baa939f3a

Compared to this, it computes intersect with screen bounding box and has the (untested) C++ version. However, your core code seems simpler.

pelikhan:
Yours looks ok. Let’s move on :)

PR #231, opened 2018-02-25
screen and image support

mmoskal:
This adds `Image` type with associated drawing operations. It doesn't add support for any physical screen components (just yet), but it's easy to plug them in. The simulator is supported. It is fairly complete.

It also adds game+sprite library. That one is to a large extent work in progress.

There is also a bugfix for `Segment::insert`

See also https://github.com/Microsoft/pxt-32/blob/master/TODO.md


libs/game/game.ts
```diff
@@ -0,0 +1,92 @@
...
+            screen.printCenter("Score: " + game.score(), top + 23, 2, image.font5)
+            if (!effect)
+                loops.pause(1000) // wait for users to stop pressing keys
+            waitAnyKey()
+            control.reset()
+        })
+    }
+
+    export function score() {
+        return _score
```
pelikhan: Call initScore

libs/game/game.ts
```diff
@@ -0,0 +1,92 @@
...
+        })
+    }
+
+    export function setScore(score: number) {
+        initScore()
+        _score = score | 0
+    }
+
+    export function addToScore(points: number) {
+        setScore(_score + points)
```
pelikhan: InitScore

libs/game/pxt.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "game",
```
pelikhan: Maybe we should split out the sprites

PR #247, opened 2018-03-17
Use the new blockCombine

mmoskal:
Depends on https://github.com/Microsoft/pxt/pull/3983

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -82,7 +82,7 @@ namespace game {
      * @param img 
      */
     //% weight=24
-    //% _blockId=backgroundaddimage block="add background image|distance %distance|aligned % alignment|image %img"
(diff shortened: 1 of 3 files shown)


pelikhan:
What happens when we define a set of const predefined const images? Will it still work without “fixedinstance”?

mmoskal:
Well, you could still expose these static images as blocks, maybe even as a single block if we have something like the `blockCombine` also on variables. I guess the question is do you want to allow something like this in blocks:

```
let img = images.create(12,12)
img.set(1,2,3)
```

Right now, there are no fixed instances of images, so we get an error message from the blocks layer when there is fixedInstances on the type.

mmoskal:
OK, it seems it's already used somewhere

PR #264, opened 2018-03-29
Scenes

riknoll:
Adding a stack for the game state on top of Peli's stack for event handlers


libs/game/game.ts
```diff
@@ -14,11 +71,13 @@ namespace game {
     export let debug = false;
     export let flags: number = 0;
```
pelikhan: Migrate this one too

pelikhan:
The keys should just work since they are compatible with control.eventContext.

libs/game/game.ts
```diff
@@ -14,11 +71,13 @@ namespace game {
...
-    let __isOver = false
+
+    export let scene: Scene;
+    let sceneStack: Scene[];
+
+
     let __waitAnyKey: () => void
-    let __background: Background;
-    let __tileMap: tiles.TileMap;
+    let __isOver = false;
```
mmoskal: I might have started it, but I'm not sure why we have all these underscores here...

pelikhan:
The Monaco editor hides members stating with __ in inyellisense. Not useful if not exported.

PR #271, opened 2018-04-02
Adding string prompt to game library

riknoll:
As part of this change I also had to change the keys library so that we can turn off user events. Any other prompt or "system" UI APIs will need to do the same.

### DIFF
--- a/libs/game/prompt.ts
+++ b/libs/game/prompt.ts
@@ -1,0 +1,541 @@
+namespace game {
+    export interface PromptTheme {
+        colorPrompt: number;
+        colorInput: number;
+        colorInputHighlighted: number;
+        colorInputText: number;
+        colorAlphabet: number;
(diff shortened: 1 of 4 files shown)


libs/game/prompt.ts
```diff
@@ -0,0 +1,541 @@
...
+                    colorBottomText: 1,
+                };
+            }
+            this.cursorRow = 0;
+            this.cursorColumn = 0;
+            this.upper = false;
+            this.inputIndex = 0;
+        }
+
+        _draw() {
```
pelikhan: We should really have a consistent naming convention. Should this be a private function?

riknoll: I think so, it's the internal drawing function. Really the only public API on this class right now is `show()`

pelikhan:
private keyword?

PR #275, opened 2018-04-04
Game color picker and image shadow blocks

riknoll:
All images are now shadow blocks and all color pickers use the new "color index" picker instead of a number field. This requires the latest PXT core

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -127,7 +127,7 @@ namespace game {
      */
     //% group="Background"
     //% weight=25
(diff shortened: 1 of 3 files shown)


pelikhan:
Do you have more details about "This is a breaking change for existing blocks projects. Shouldn't affect the TS" ?

riknoll:
The blocks that previously used the sprite editor now have a value instead of a field for their image arguments. So, if you had an old blocks project and you try to open it you'll probably lose all the sprites you created. Shouldn't affect any examples in the documentation.

PR #285, opened 2018-04-14
Adding animations for sprites

riknoll:
This defines two type of animations: one that advances based on movement and one that advances based on time.

image

The first two blocks are for movement; the way it works is that you add frames (images) for a movement direction and then the frames are automatically advanced when the sprite moves. Right now the threshold for advancement is not configured in blocks (not sure if that's useful or not).

The remaining blocks are for timed animations. Unlike movement animations, timed animations are not tied to a sprite so they can be created once and then fired multiple times on multiple sprites.

The `fireAnimation()` API just adds an animation to a sprite's animation queue and is non-blocking. We can't really make it blocking because that would totally break the game loop if you fired it within the update function. Also `fireAnimation()` has an optional parameter for flipping the animation along the x, y, or both x and y axes.

There currently isn't a way to do something when a timed animation ends. I don't know what the best API for that is.

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -25,6 +25,25 @@ interface SpriteLike {
     __draw(camera: scene.Camera): void;
 }
 
+interface AnimationAction {
+    animation: sprites.TimedAnimation;
+    durationMs: number;
+    flipX: boolean;
+    flipY: boolean;
+}
+
+
+enum FlipOption {
+    //% block=none
+    None,
+    //% block="flip x"
+    FlipX,
+    //% block="flip y"
+    FlipY,
+    //% block="flip x+y"
+    FlipXY
+}
+
 /**
  * A sprite on screem
  **/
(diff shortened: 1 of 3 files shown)


libs/game/sprite.ts
```diff
@@ -300,6 +344,60 @@ class Sprite implements SpriteLike {
...
+
+    /**
+     * Queues a timed animation on the sprite without blocking the current fiber
+     * @param animation The animation to show
+     * @param duration The duration of the animation
+     * @param flip Show the animation flipped over an axis
+     */
+    //% blockId=spritefireanimation block="%sprite fire animation %animation %duration=timePicker|ms||%flip"
+    //% group="Animations" weight=79
+    fireAnimation(animation: sprites.TimedAnimation, duration: number, flip = FlipOption.None) {
```
pelikhan: probably name it "startAnimation"  ?

PR #291, opened 2018-04-19
Make sprite handlers nestable

riknoll:
Maybe we should discuss before merging this, but very often when you create a sprite you do it in a function (e.g. creating projectiles) and top-level event handlers are useless in that case. In general it is more useful for these blocks to be nestable.


pelikhan:
Sounds good to me. We did that a lot of TouchDevelop games. We also had a “on update” handler to attach game loop events to individual sprites.

PR #293, opened 2018-04-21
Standard key support

pelikhan:
@mmoskal can you check the conversion of the first character to character code?

### DIFF
--- a/libs/keyboard/keyboard.cpp
+++ b/libs/keyboard/keyboard.cpp
@@ -119,7 +119,7 @@ namespace keyboard {
     //% blockGap=8 weight=99
     void key(String key, KeyboardKeyEvent event) {
         if (!key->length) return;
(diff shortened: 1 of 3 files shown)


libs/keyboard/keyboard.cpp
```diff
@@ -112,6 +112,27 @@ namespace keyboard {
...
     }
 
+    /**
+    * Sends a media key command
+    */
+    //% blockId=keyboardStandardKey block="keyboard key %key|%event"
+    //% blockGap=8 weight=99
+    void key(String key, KeyboardKeyEvent event) {
+        if (!key->length) return;
+        uint16_t ckey = pxt::toUInt(pxt::fromInt(key->data[0]));
```
mmoskal: Just do `ckey = key->data[0]`

PR #307, opened 2018-05-04
spi support

pelikhan:
Basic support for SPI (no changing of pins yet)

### DIFF
--- a/libs/core/docs/reference/pins/spi-mode.md
+++ b/libs/core/docs/reference/pins/spi-mode.md
@@ -1,0 +1,16 @@
+# SPI Format
+
+Sets the SPI format
+
+```sig
+pins.spiMode(3, 8);
+```
+
+## Parameters
+
+* ``mode``,
+* ``bits``, 
+
+## See also
+
+[SPI](https://developer.mbed.org/handbook/SPI)
--- a/libs/core/spi.cpp
+++ b/libs/core/spi.cpp
@@ -1,0 +1,55 @@
+#include "pxt.h"
+#include "SPI.h"
+#include "ErrorNo.h"
+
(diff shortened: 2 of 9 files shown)


libs/core/docs/reference/pins/spi-mode.md
```diff
@@ -0,0 +1,16 @@
+# SPI Format
+
+Sets the SPI format
+
+```sig
+pins.spiMode(3, 8);
```
mmoskal: I would say drop number of bits - most chips only support 8 bit

libs/core/spi.cpp
```diff
@@ -0,0 +1,54 @@
...
+        return spi->write(value);
+    }
+
+    /**
+    * Writes a given command to SPI bus, and afterwards reads the response.
+    */
+    //% help=pins/spi-transfer weight=4 advanced=true
+    //% blockId=spi_transfer block="spi transfer %command into %response"
+    void spiTransfer(Buffer command, Buffer response) {
+        initSPI();
```
mmoskal: Allow for command and/or response being null - pass NULL,0 to the codal function then

libs/core/spi.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+    * Sets the SPI mode and bits
+    * @param mode the mode, eg: 3
+    * @param bits the number of bits, eg: 8
+    */
+    //% help=pins/spi-mode weight=3 advanced=true
+    //% blockId=spi_mode block="spi mode|mode %mode|bits %bits"
+    void spiMode(int mode, int bits = 8) {
+        initSPI();
```
mmoskal: Drop bits

PR #310, opened 2018-05-07
updated block names

pelikhan:
- [x] Remove complex background blocks
image
- [x] renamed dx/dy
image
- [x] moving images to advanced
- [x] move "add movement frame" up
- [x] rename timed animaton to animation
- [x] moving "facing up" to top and disconnect from "animation"
- [x] remove second "pause for any key"


riknoll:
"left-right keys" or "left-right move"? Buttons is probably too long.

PR #319, opened 2018-05-16
Tilemap performance

riknoll:
This PR removes the obstacle flag for sprites and significantly improves the performance of the tilemap (more than 10x faster). All collision blocks now appear in the scene category and are specific to tiles but `onOverlap()` still exists for sprites. Also, instead of passing references to objects around the new APIs refer to tiles by color:

image

In the future we could consider bringing back sprite collisions but we would probably need to use a separate set of blocks. There isn't anything technical preventing it right now; it's just confusing to explain. Maybe it should be a TypeScript only feature? In any case, I made `Obstacle` an interface so that we could create other ones down the road.

Random other changes:
* Added a typings file in libs\game so that vscode stops complaining about type errors (it's ignored by PXT)
* Fixed all of those compiler errors in the build about controller not existing
* Added FPS to the render stats
* Added a few functions for tracking performance events; the other part of that change needs to be done in PXT so right now they are a no-op


libs/base/eventcontext.ts
```diff
@@ -86,8 +86,11 @@ namespace control {
                     this.timeInSample += runtime
                     this.framesInSample++
                     if (this.timeInSample > 1000 || this.framesInSample > 30) {
-                        if (EventContext.onStats)
-                            EventContext.onStats(`render: ${Math.round(this.timeInSample / this.framesInSample * 1000)}us`)
+                        if (EventContext.onStats) {
+                            const fps = Math.round(this.framesInSample / (this.timeInSample / 1000));
+                            const renderMicro = Math.round(this.timeInSample / this.framesInSample * 1000)
+                            EventContext.onStats(`FPS: ${fps} render: ${renderMicro}us`)
```
mmoskal: Just drop the `us`

mmoskal: oh, and maybe then just say `${fps}fps`

libs/game/scene.ts
```diff
@@ -56,9 +63,11 @@ namespace scene {
             // render sprites 90
             this.eventContext.registerFrameHandler(90, () => {
                 if (this.flags & Flag.NeedsSorting)
-                    this.allSprites.sort(function (a, b) { return a.z - b.z || a.id - b.id; })
+                this.allSprites.sort(function (a, b) { return a.z - b.z || a.id - b.id; })
```
pelikhan: let's measure sort perf too

PR #327, opened 2018-05-24
Top level events for sprites

pelikhan:
Adds top level events for created, destroed, collision, overlap. None of the existing APIs have been removed yet.

Test script: https://makecode.com/_hMuaE7PCxgre

image

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -83,11 +83,15 @@ class ArcadePhysicsEngine extends PhysicsEngine {
             const overSprites = scene.physicsEngine.overlaps(sprite);
             for (const overlapper of overSprites) {
                 // overlap handler
+                const tmpsprite = sprite;
+                const tmp = overlapper;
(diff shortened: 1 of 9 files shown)


riknoll:
LGTM! What do we want for the updated authoring? Maybe something like this:

```typescript
//% block="on %kind1{0} %sprite1{2} overlap with %kind2{1} %sprite2{3}"
onOverlap(kind1: SpriteKind, kind2: SpriteKind, cb: (sprite1: Sprite, sprite2: Sprite) => void) {

}
```

Where the `{0}` indicates the argument index (which I guess parameters to the callback would also be given)? @samelhusseini what do you think?

samelhusseini:
Looks good @riknoll

riknoll:
Also it would probably be a breaking change for all our targets.

pelikhan:
AAARGG. No. Let's do the curly stuff

PR #333, opened 2018-05-31
Servo blocks

pelikhan:
Adds friendlier blocks to control servos.
image

(would be nice to have a dedicated angle picker)


abchatra:
Are Motors top level category?

PR #344, opened 2018-06-07
Add core---stm32f401re

mmoskal:
Adds support for STM32F401 (from maker), also:
* add support for ST7735 screens (moved from maker)
* support getting settings from bootloader area
* add config-indirect versions of getPin and getButton functions
* minor fixes to dialogs to align color and mono versions


libs/buttons/buttons.cpp
```diff
@@ -47,6 +47,13 @@ Button *getButtonByPin(int pin, int flags) {
     return btn;
 }
 
+//%
+Button *getButtonByPinCfg(int key, int flags) {
+    int pin = getConfig(key);
+    if (pin == -1) target_panic(42);
```
pelikhan: nice

libs/game/game.ts
```diff
@@ -58,13 +58,9 @@ namespace game {
 
     function showDialogBackground(h: number, c: number) {
         const top = (screen.height - h) >> 1;
-        if (screen.isMono) {
-            screen.fillRect(0, top, screen.width, h, 0)
-            screen.drawLine(0, top, screen.width, top, 1)
-            screen.drawLine(0, top + h - 1, screen.width, top + h - 1, 1)
-        } else {
-            screen.fillRect(0, top, screen.width, h, c)
```
pelikhan: no more color?

mmoskal: It's only the background, I think black with the white lines look better. It was all black&white for a while because of `screen.isMono` bug. @riknoll is going to re-do these anyways

pelikhan: yup

PR #355, opened 2018-06-13
Serial blocks

pelikhan:
Resurect and refactor Serial (UART) blocks into a separate project. Requires https://github.com/lancaster-university/codal-mbed/pull/1 for CPX

- [x] move serial blocks to separate project (most users will not use this)
- [x] added block "serial attach to console" to pipe console.log to serial


libs/serial/enums.d.ts
```diff
@@ -0,0 +1,31 @@
+// Auto-generated. Do not edit.
+
+
+    declare const enum BaudRate {
+    //% block=115200
```
mmoskal: You can go up 960-something

libs/core/pins.h
```diff
@@ -129,6 +129,14 @@
 #define CODAL_I2C CODAL_MBED::I2C
 #endif
 
+#ifndef CODAL_SERIAL
+#define CODAL_SERIAL CODAL_MBED::Serial
+#endif
+
+#ifndef CODAL_SERIAL_HEADER
+#define CODAL_SERIAL_HEADER "MbedSerial.h"
```
mmoskal: You can just add include mbed serial.h to platform.h and skip this

libs/core/usb.cpp
```diff
@@ -74,11 +76,18 @@ void usb_init() {
...
+}
+
+void setSendToUART(void (*f)(const char *, int)) {
+    pSendToUART = f;
 }
 
 #else
 void usb_init() {}
 __attribute__((weak)) void sendSerial(const char *data, int len) {}
+__attribute__((weak)) void setSendToUART(void (*f)(const char *, int)) {}
```
mmoskal: If there is no USB we probably should also support sending to pins

PR #362, opened 2018-06-20
adding repeated event

pelikhan:
Remove setRepeat, add event.


mmoskal:
Does it rise repeat also on first press? I think it should. Maybe it should be up, down, press?

pelikhan:
Only on second down repeat for now. I think you can handle both differently. Maybe repeat is “running”.

PR #364, opened 2018-06-22
New arcade physics

riknoll:
Changes:

1. All tiles are now 16x16 px
2. Physics now has a maximum velocity of 15px per step
3. Collisions are now based on hitboxes and not overlaps

Hitboxes for sprites are calculated based on the content of the image but are always rectangular. Swapping images will recalculate the hitboxes. Sprites that are larger than 15x15 px have multiple hitboxes because there is a hard limit based on the tile size;, having a ton of hitboxes doesn't really have a noticeable impact on perf so we're all good.

Things I still need to do before merging:

1. Render hitboxes when debug is on
2. Figure out what to do if an image is changed directly (i.e. when you draw to an image we should maybe recalculate hitboxes)

*NOTE:* This PR only affects the TileMap. Overlap behavior is still exactly the same.

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -273,8 +273,14 @@ class Sprite implements SpriteLike {
                 font);
         }
         // debug info
-        if (game.debug)
-            screen.drawRect(l, t, this.width, this.height, 3);
+        if (game.debug) {
+            let color = 1;
+            this._hitboxes.forEach(box => {
+                this._image.drawRect(box.ox, box.oy, box.width, box.height, color);
+                color++;
+                if (color >= 15) color = 1;
+            });
+        }
     }
 
     __update(camera: scene.Camera, dt: number) {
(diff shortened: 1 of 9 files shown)


libs/game/sprite.ts
```diff
@@ -135,6 +142,7 @@ class Sprite implements SpriteLike {
     setImage(img: Image) {
         if (!img) return; // don't break the sprite
         this._image = img;
+        this._hitboxes = game.calculateHitBoxes(this);
```
mmoskal: What happens if someone does `mySprite.image.setPixel(10, 10, 12)` ? I guess we shouldn't support it....

riknoll: Yeah, doesn't work. I'd like to support it but I don't have any good ideas on how

mmoskal: there's dirty flag on images, currently only used on `screen` - I guess we could use it, but maybe some time later

PR #367, opened 2018-06-26
Hud/info and splash screen improvements

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/145
Fixes https://github.com/Microsoft/pxt-arcade/issues/144

Changes to the splash screen:
1. Now uses my dialog stuff so you get a hint to press A to dismiss
1. Text automatically scrolls when larger than screen
1. Subtitle is in a smaller font then the main title

Changes to the info category:
1. Score, countdown, and timer are now all displayed with backgrounds and borders to make them more legible
1. Life is now displayed using heart icons.
1. Score font will get smaller if the score is greater than 1,000,000 and no longer overlap with countdown
1. Countdown now handles minutes and seconds and flashes for the last ten seconds
1. Added events for when life reaches 0 and when the countdown ends. If the events are not used then we just do game over
1. High score is not displayed (it still shows up on the game over screen)

For the High score, I really don't think it's important enough to take up screen real estate.
Another thing that games often do is change how the score looks when you get a new high score.
We could change the color or add a star next to it or something.

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -65,23 +65,6 @@ namespace game {
         return top;
     }
 
-    /**
-     * Show a title, subtitle menu
-     * @param title
-     * @param subtitle
-     */
-    //% group="Gameplay"
-    //% weight=90 help=game/splash
-    //% blockId=gameSplash block="splash %title||%subtitle"
-    //% group="Prompt"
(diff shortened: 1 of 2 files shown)


riknoll:
New info display:
arcade_info

New splash screen:
arcade_splash

PR #368, opened 2018-06-28
Rename sprite variable and adding an API for creating an "empty" sprite

riknoll:
"sprite" is now "agent" and the new create empty sprite API is in the "Lifecycle" category


riknoll:
Also now the projectile block shows its "kind" parameter by default to match the other create block

PR #373, opened 2018-07-12
Animation package

abchatra:
This changes removes existing animation library and creates a new set as a package.
Idea is you create a animation and associate a action (example: jumping, idle or running). You then add that animation to a sprite. You will need to activate the action when you want to see the animation

I reused the onUpdate instead of registering a seperate handle.


abchatra:
Fixes https://github.com/Microsoft/pxt-arcade/issues/173

abchatra:
@riknoll ready for review

libs/animation/animation.ts
```diff
@@ -0,0 +1,136 @@
...
+            game.eventContext().registerFrameHandler(16, () => {
+                _onSpriteUpdate.forEach(element => {
+                    element();
+                });
+            });
+        }
+
+        _onSpriteUpdate.push(() => {
+            if (sprite._action === set.action) {
+                sprite.setImage(set.getImage())
```
riknoll: Better to not set the image every update because it will re-trigger the hitbox calculations

PR #374, opened 2018-07-16
Arcade bugs 

abchatra:
- Exposes backgroundImage and backgroundColor as API's (Fixes https://github.com/Microsoft/pxt-arcade/issues/190)
- "Stay in screen" is the default for setflag API (Fixes https://github.com/Microsoft/pxt-arcade/issues/183). We decided to skip another API for auto destroy.
- Get and Set image of the sprite are back to facilitate custom animations.


abchatra:
Docs is a just a placeholder

PR #375, opened 2018-07-16
Bubble dialog

ashaltu:
This fixes Microsoft/pxt-arcade#177.

Remade the say block so that the text will be centered and the box size fits a certain length and caps at a certain length which enables the text to scroll.


libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+        let font = image.font8;
+        let startX = 2;
+        let startY = 2;
+        let bubbleWidth = text.length * font.charWidth + bubblePadding;
+        let maxOffset = text.length * font.charWidth - maxTextWidth;
+        let spriteTop = this._hitboxes[0].top;
+
+        for (let i = 0; i < this._hitboxes.length; i++) {
+            if (this._hitboxes[i].top < spriteTop) {
+                spriteTop = this._hitboxes[i].top;
```
riknoll: The more JavaScript-y way to write this is:
`spriteTop = Math.min(spriteTop, this._hitboxes[i].top)`

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+            }
+        }
+
+        spriteTop = this.y - spriteTop;
```
riknoll: I think this calculation is just cancelled out on line 267 so you can probably remove this line and change the other one to `this.bubbleBoxSprite.y = spriteTop - ((font.charHeight + bubblePadding) >> 1) - 2; `

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+        this.updateSay = () => {
+            // update box stuff as long as timeOnScreen doesn't exist or it can still be on the screen
+            if (!timeOnScreen || timeOnScreen > control.millis()) {
+                this.bubbleBoxSprite.image.fill(textBoxColor);
+                // the minus 2 is how far the bottom of bubbleBoxSprite to sprites top is
+                this.bubbleBoxSprite.y = this.y - spriteTop - ((font.charHeight + bubblePadding) >> 1) - 2;
+                this.bubbleBoxSprite.x = this.x;
+                // pauses at beginning of text for holdTextTimer length
+                if (holdTextTimer > 0) {
+                    holdTextTimer = holdTextTimer - game.eventContext().deltaTime;
```
riknoll: `holdTextTimer -= game.eventContext().deltaTime;`

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+        if (!text) {
+            this.updateSay = undefined;
+            return;
+        }
+
+        if (timeOnScreen) {
+            timeOnScreen = timeOnScreen + control.millis();
+        }
+        let pixelsOffset = 0;
+        let holdTextTimer = 1.5;
```
riknoll: I would rename this variable to `holdTextSeconds` or something like that because you are using milliseconds for the other timer

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+                    }
+                }
+                // if maxOffset is negative it won't scroll
+                if (maxOffset < 0) {
+                    this.bubbleBoxSprite.image.print(text, startX, startY, textColor, font);
+                } else {
+                    this.bubbleBoxSprite.image.print(text, startX - pixelsOffset, startY, textColor, font);
+                }
+
+                // left side padding
```
riknoll: Instead of using for-loops for padding you could just use the `image.fillRect()` API

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+
+        spriteTop = this.y - spriteTop;
+
+        if (bubbleWidth > maxTextWidth + bubblePadding) {
+            bubbleWidth = maxTextWidth + bubblePadding;
+        } else {
+            maxOffset = -1;
+        }
+
+        this.bubbleBoxSprite = sprites.create(image.create(bubbleWidth, font.charHeight + bubblePadding));
```
riknoll: If `bubbleBoxSprite` is already defined then you need to destroy the existing one or else it will leak

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
+                // corners removed
+                this.bubbleBoxSprite.image.setPixel(0, 0, 0);
+                this.bubbleBoxSprite.image.setPixel(bubbleWidth - 1, 0, 0);
+                this.bubbleBoxSprite.image.setPixel(0, font.charHeight + bubblePadding - 1, 0);
+                this.bubbleBoxSprite.image.setPixel(bubbleWidth - 1, font.charHeight + bubblePadding - 1, 0);
+            } else {
+                // if can't update because of timeOnScreen then destroy the bubbleBoxSprite and reset updateSay
+                this.bubbleBoxSprite.destroy();
+                this.updateSay = undefined;
+                return;
```
riknoll: No need to return

libs/game/sprite.ts
```diff
@@ -214,18 +215,103 @@ class Sprite implements SpriteLike {
...
     //% help=sprites/sprite/say
-    say(text: string, millis?: number) {
-        this._say = text;
-        if (!millis || millis < 0)
-            this._sayExpires = -1;
-        else
-            this._sayExpires = control.millis() + millis;
+    say(text: string, timeOnScreen?: number, textColor = 15, textBoxColor = 1) {
+        if (!text) {
+            this.updateSay = undefined;
```
riknoll: Also destroy the existing sprite if it exists

libs/game/sprite.ts
```diff
@@ -85,11 +85,12 @@ class Sprite implements SpriteLike {
...
     //% blockCombine block="lifespan"
     lifespan: number;
-    private _say: string;
-    private _sayExpires: number;
     private _image: Image;
     private _obstacles: sprites.Obstacle[];
-    
+
+    private updateSay: () => void;
+    private bubbleBoxSprite: Sprite;
```
riknoll: Could we rename this to `sayBubbleSprite`? Just to be more consistent

PR #370, opened 2018-07-03
DM: update I2C object

deanm1278:
update i2c object for samd21. This fixed some weirdness with i2c reads

PR #390, opened 2018-08-08
update function keys

deanm1278:
We had received reports of function keys not working, so I've updated the enumerations to the keys found here:
http://www.freebsddiary.org/APC/usb_hid_usages.php

note that this requires this codal PR:
https://github.com/lancaster-university/codal-core/pull/31

I currently have no way to test this until it goes up because I cannot get the keyboard module and the codal core to build on a local machine.


ladyada:
@riknoll @mmoskal can we get a looksee & merge?

mmoskal:
I think the `CPlayI2C.h` change doesn't belong in this PR. Also I don't this it requires the codal changes - the enum for function keys is defined completely separately here. I guess we need to make sure the one defined here has the right order.

deanm1278:
I'm actually kind of confused as to why the I2C object change appears here since it was already merged here: https://github.com/Microsoft/pxt-common-packages/pull/370

did I mess something up with git accidentally?

deanm1278:
ok @mmoskal I've syncd our fork and re-added just the keyboard changes

PR #394, opened 2018-08-20
Arcade tilemap bugs

jwunderl:
Fixes Microsoft/pxt-arcade#159, fixes Microsoft/pxt-arcade#218

Tilemap transparency, so background color / image shows through, instead of replacing all transparent portions with black

Issue 218 caused by hitboxes changing when a sprite's image is changed;  attempts to resolve this by bumping sprite back towards a 'safe' area if the sprite now extends into a wall. This implementation only attempts to resolve this in the case of a sprite that fits within a single tile, as otherwise there isn't a guarantee that the sprite actually fits anywhere on the map.

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -10,6 +10,8 @@ class PhysicsEngine {
 
     removeSprite(sprite: Sprite) { }
 
+    moveSprite(s: Sprite, tm: tiles.TileMap, dx: number, dy: number) { }
+
     draw() { }
 
     /** Apply physics */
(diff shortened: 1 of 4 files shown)


riknoll:
Instead of redoing a lot physics logic it might be easier to make `moveSprite()` public from `physics.ts` and set `dx` and `dy` to the difference between the top/left of the old sprite and the top/left of the new sprite

riknoll:
And by top/left of the sprite I mean the top/left of the hitboxes of the sprite

jwunderl:
Well I completely forgot about that moveSprite did this already, even though I just fixed the ghost bug a few days ago..

PR #400, opened 2018-08-25
Dynamic Sprite.say scroll speed

ChaseMor:
I have come across issues when using Sprite.say and also giving a timeOnScreen parameter. It seems that there is a arbitrary amount of time that the sprite waits before scrolling and the scrolls at an arbitrary speed.

This is especially frustrating as it means that you have less control of the timeOnScreen parameter as it now has a min value for displaying all of the text.

I propose, adjusting the sprite's scroll wait period and speed based the timeOnScreen is given.


riknoll:
@ashaltu would you mind testing this change out if you have time?

riknoll:
@ChaseMor would you mind making the changes Abdul suggested?

ChaseMor:
@riknoll

PR #404, opened 2018-08-30
showLongText tweaks

jwunderl:
* Resolves Microsoft/pxt-arcade#254; up and down arrow keys can be used to scroll through the pages of the text (down arrow key acting as an alternative to A to move to the next chunk, up arrow key going back a screen / page (if possible))
* Allow more of the dialog box to be used to display text

### Current

Current Full Screen

Current Bottom Screen

Current Right Screen

### New

New Full Screen

New Bottom Screen

New Right Screen

### DIFF
--- a/libs/game/textDialogs.ts
+++ b/libs/game/textDialogs.ts
@@ -197,9 +197,9 @@ namespace game {
             let current = "";
 
             while (strIndex < str.length) {
-                const lastIndex = strIndex + charactersPerRow - 1;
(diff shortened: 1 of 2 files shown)


jwunderl:
@riknoll Yeah, I was thinking on that, I didn't really have a strong opinion so I just defaulted to down and A being the same behavior, I'll switch it if you all prefer it that way~

riknoll:
I don't have a strong opinion either

PR #409, opened 2018-09-04
DNM: network driver

mmoskal:
CC @jamesadevine


libs/network/network.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    JDDevice *getDevice() { return &device; }
+
+    ~JDProxyDriver() { decrRC(methods); }
+};
+
+//%
+JDProxyDriver *addNetworkDriver(int driverClass, RefCollection *methods) {
+    static int deviceId = 3030;
+    return new JDProxyDriver(JDDevice(0, 0, 0, driverClass), methods, ++deviceId);
```
jamesadevine: You'll need to pass flags which tells the logic driver what mode to run the driver in. I've got a commit that has an enum instead of a #define for the flags.

mmoskal: Sure thing. I'm assuming you're handling the serial number?

jamesadevine: Will do.

libs/network/network.ts
```diff
@@ -0,0 +1,59 @@
...
+namespace network {
+    //% shim=pxt::programHash
+    export function programHash(): int32 { return 0 }
+
+    //% shim=network::addNetworkDriver
+    function addNetworkDriver(deviceClass: int32, methods: ((p:Buffer) => void)[]): NetworkDriverStatus { 
+        return null
+    }
+
+    export function addDriver(deviceClass: int32, n: NetworkDriver) {
```
pelikhan: deviceClass not used?

libs/network/pxt.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "network",
```
pelikhan: should we have a more unique name? network could be confusing.

libs/base/core.cpp
```diff
@@ -482,11 +482,11 @@ TValue ptrneqq(TValue a, TValue b) {
 }
 } // namespace langsupp
 
+
 #define NUMOP(op) return fromDouble(toDouble(a) op toDouble(b));
 #define BITOP(op) return fromInt(toInt(a) op toInt(b));
 namespace numops {
 
-//%
```
pelikhan: @mmoskal  are these changes needed or just a merge fluke?

jamesadevine:
👍

PR #413, opened 2018-09-12
code-gen re-work

mmoskal:
Support for https://github.com/Microsoft/pxt/pull/4671

* memory leaks are now tracked via logging, not on-device `set<>`
* fixes to comparison and various type conversions


mmoskal:
This needs to be merged at the same time as `pxt-core` PR

PR #417, opened 2018-09-26
Allow for reading config from SAMD51 bootloader

mmoskal:


libs/base/core.cpp
```diff
@@ -923,12 +924,17 @@ int getConfig(int key, int defl) {
...
-        if (cfgData[i] == key)
-            return cfgData[i + 1];
-        if (cfgData[i] == 0)
-            break;
-    }
+#ifdef PXT_BOOTLOADER_CFG_MAGIC
+    cfgData++;
+    if ((void*)0x200 <= cfgData && cfgData < (void*)PXT_BOOTLOADER_CFG_ADDR &&
+        cfgData[-1] == (int)PXT_BOOTLOADER_CFG_MAGIC)
+#endif
```
pelikhan: so the for loop does not execute if the 'if' is false?

mmoskal:
Yes, we only want the loop if the magic is there.

PR #423, opened 2018-10-05
Support for GC, and some new APIs

mmoskal:
This is in support of https://github.com/Microsoft/pxt/pull/4782

Also:
* add `control.dmesgValue()`
* add `control.gc()`
* add `control.profilingEnabled()`
* add `control.dumpPerfCounters()`
* add pseudo-API `control.enablePerfCounter()` (enables profiling for function if present; handled by compiler)
* add `JSON.parse` and `JSON.stringify` (in TS)
* adjust to codal `master`
* synchronous handlers for collisions (this will need to be changed to drop-when-running) CC @riknoll 
* frowny face on panic on hardware


mmoskal:
Updated comment, should be good to merge (along with the PXT PR)

PR #424, opened 2018-10-08
Add sensor of Airquality CSS811

sabas1080:
Hi @pelikhan 

I have created the air quality package based on the CCS811 sensor, based on sparkfun work, tested in SAMD21

Thanks


pelikhan:
Hi @sabas1080 - for sensor, it will be best to host it in a github under your name/organization and add it to the "approvedRepos" in the targetconfig of maker (or other). That way you can iterate your package faster than the editor. Makes sense?

pelikhan:
@sabas1080 we could also support external boards as github repos.

sabas1080:
@pelikhan  ok, for the sensors I agree with that you are saying, I think there should be a library manager like Arduino's manager library https://www.arduino.cc/en/Guide/Libraries as a simple way for people to find third-party libraries.

about the external boards has my collaborate I like more than they are officially in the Microsoft repository

pelikhan:
Right let's keep the boards here.
So if you move your sensor package into a separate repor, we can surface it in the "Extensions" dialog.
We use GitHub itself as our library manager.

pelikhan:
Go to "Extensions" dialog and click on "Github login" - then you can create and edit your packages from the maker editor itself.

PR #426, opened 2018-10-11
Local multiplayer

riknoll:
Adds support for local multiplayer 

Things left to do:
- [ ] Didn't implement the controlSprite API from the controller
- [ ] Player connect/disconnect events
- [ ] Set an icon and color


riknoll:
Moved everything under the controller category in a sub-group called "multiplayer".

PR #427, opened 2018-10-12
local multiplayer APIs

jwunderl:
Adding a few APIs for local multiplayer. ~Still in progress for score / lives; I'll finish it up tonight most likely as I have course material to finish during working hours :) (so do not merge for now)~

### local-multiplayer/local

* ``control player {} with vx {} vy {}``
    * Keeps track of player, so that if the player changes sprites movement switches to that sprite
* ``player {} sprite``
    * Returns sprite currently attached to player
* add ``spritescreate``  as a shadow block for ``setPlayerSprite``; feels a bit ambiguous what belongs there to me with just a hole. (Bug with shadow blocks that causes it to be not ideal, though: Microsoft/pxt-arcade#249)

### game/info:

* getters for the colors of the HUD
* functions to turn on / off display of life / score / countdown

### local-multiplayer/info 

* scores and lives for players one and two, as per Richard's beautiful drawing below (& my handwriting). Only players one and two will be displayed (others will be tracked still), with their player number in a colored box with clear text.
* ``info.onPlayerLifeZero`` to set behavior when player runs out of life
* ``info.playerHasLife`` to check if a player is alive or not (js only)
* ``info.playerInfo`` gets the PlayerInfo object which defines the location / colors / behaviors of the player info box (js only)

Standard info api on left, local multiplayer on right; text always smaller, hearts smaller and always using multiplication instead of displaying more than one heart.

Standard Info vs Multiplayer Info

### DIFF
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -286,7 +286,7 @@ namespace info {
      * Set whether life should be displayed
      * @param on if true, lives are shown; otherwise, lives are hidden
      */
-    export function ShowLife(on: boolean) {
+    export function showLife(on: boolean) {
         initLife();
         updateFlag(Visibility.Life, on);
     }
@@ -295,7 +295,7 @@ namespace info {
      * Set whether score should be displayed
      * @param on if true, score is shown; otherwise, score is hidden
      */
-    export function ShowScore(on: boolean) {
+    export function showScore(on: boolean) {
         initScore();
         updateFlag(Visibility.Score, on);
     }
@@ -304,7 +304,7 @@ namespace info {
      * Set whether score should be displayed
      * @param on if true, score is shown; otherwise, score is hidden
      */
-    export function ShowCountdown(on: boolean) {
+    export function showCountdown(on: boolean) {
         updateFlag(Visibility.Countdown, on);
     }
 
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -25,7 +25,7 @@ namespace info {
     let _borderColor: number;
     let _fontColor: number;
     let _countdownExpired: boolean;
-    let _visibilityFlag: number;
+    let _visibilityFlag: number = Visibility.None;
 
 
     let _lifeOverHandler: () => void;
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -10,7 +10,7 @@ namespace info {
         None = 0,
         Countdown = 1 << 0,
         Score = 1 << 1,
-        Lives = 1 << 2,
+        Life = 1 << 2,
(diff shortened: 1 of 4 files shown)


jwunderl:
@riknoll I think I had a good idea on the drive down to Olympia on making the info variables better. Blocks behavior will be the same as discussed, but internally use a config for the players

```typescript
namespace info {
    export interface PlayerInfo {
        score: number;
        life: number;
        showScore: boolean;
        showLife: boolean;
        h?: () => void; // onPlayerLifeOver handler
        x?: number;
        y?: number;
        bg?: number; // background color
        // ... needs a few more
    }
}
```

By default players one and two will be displayed, but this way it can be configured in javascript with a getter (choose their own colors, display the two highest scoring players, cycle through displaying all players in top right, etc)

Seem reasonable?

jwunderl:
Content complete now, I'm planning on refactoring drawPlayer a bit though; lots of the elements got interconnected to make them look right so I just need to look at it again with fresh eyes tomorrow and clean it up a bit so it's less spaghetti like.



Players one and two show by default, others are suppressed by the ``PlayerInfo`` object that is accessible through ``info.playerInfo(player)``. An indefinite number of players is supported, with players 4+ sharing placement / color configuration. Colors match Smash Bros, though I might have to change yellow to orange as the yellow is a bit... bright and hard to read.


(Just noticed I need to fix the x/y position in cases when only the player number is shown and the info is set to display leftwards or upwards)

A few buggy / weird behaviors to ask about:

* The weird type conversion I had to do when [using an enum to access an array index here](https://github.com/Microsoft/pxt-common-packages/pull/427/files#diff-39ce1f607fb82bb1cf43b525f163280cR133). (I didn't have to do it in other circumstances where I indexed with enum values. It stopped occurring once I explicitly defined a condition for three, but can be reproduced if you remove the branch prior to the noted line (that is, the ``else if (player === controller.PlayerNumber.Three)`` branch, which leaves the ``else`` responsible for players 3 and 4.
* In typescript, the autocomplete for the methods I added is broken; it has the parameters as ``PlayerNumber.One`` (etc), where they should be ``**controller**.PlayerNumber.One``. I would assume making an info.PlayerNumber enum or moving controller.PlayerNumber would fix the autocomplete.

pelikhan:
Ready?

jwunderl:
One question, looks like the reason that the local-multiplayer blocks don't decompile from script to blocks is due to the PlayerNumber enum being in the controller namespace (it also causes the trouble with autocomplete listed above) - it works fine when PlayerNumber is outside of the controller namespace. Would it be appropriate to move it out of the namespace to fix those issues?

PR #429, opened 2018-10-18
WebUSB build fix

pelikhan:
Fixing "storage" library.

### DIFF
--- a/libs/storage/pxt.json
+++ b/libs/storage/pxt.json
@@ -1,6 +1,6 @@
 {
     "name": "storage",
-    "description": "USB Pen-drive support and flash storage - beta",
+    "description": "USB Pen-drive support and flash storage",
     "files": [
         "SNORFS.cpp",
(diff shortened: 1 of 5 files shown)


libs/storage/pxt.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "storage",
-    "description": "USB Pen-drive support and flash storage - beta",
+    "description": "USB Pen-drive support and flash storage",
     "files": [
```
mmoskal: Doesn't this need DEVICE USB as well?

pelikhan:
DEVICE_USB is set in the codal target. I guess it does not hurt to add it (BLE nano does not have that define).

mmoskal:
Ble nano doesn't have usb. I guess you're right storage could work without usb

mmoskal:
Usb is currently only supported on samd21. STM 32 was close iirc and d51 should be easy

PR #430, opened 2018-10-19
Storage blocks + datalog refresh

pelikhan:
Make it work for maker


libs/datalogger/datalogger.ts
```diff
@@ -145,7 +150,9 @@ namespace datalog {
      * Sets the minimum number of milli seconds between rows
      * @param millis milliseconds between each sample, eg: 50
      */
-    //% blockId=datalogSetSamplingInterval block="set datalog sampling interval to %millis|(ms)"
+    //% group="Configuration"
+    //% blockId=datalogSetSamplingInterval block="set datalog sampling interval to $millis|(ms)"
```
mmoskal: The blocks are still called `datalog`. I'm pretty sure you will find a number of teachers and TEALS-like volunteers who heard about Datalog and will be confused about this. It's enough that we misname TypeScript to be JavaScript, let's avoid confusion about others.

pelikhan:
I missed that one. It should be “data logger”

PR #431, opened 2018-10-19
Storage defaults

pelikhan:


libs/storage/storage.cpp
```diff
@@ -79,44 +79,66 @@ snorfs::File *getFile(String filename) {
...
 
-/** Append string data to a new or existing file. */
+/** 
+* Append string data to a new or existing file. 
+* @param filename name of the file, eg: "log.txt"
+*/
 //% part="storage" 
 //% blockId="storage_append" block="append file $filename with $data"
 void append(String filename, String data) {
+    if (!filename || !filename->length) return;
```
mmoskal: The `!filename` case should no longer happen - when the user passes `null` and the C++ wants a string, the C++ will get `"null"` as per JS semantics. It will also get `"undefined"` for `undefined`, `"17"` for `17`, etc.

mmoskal: Actually, it would probably make more sense to have `getFile()` return null when it fails (also eg when the file name is too long or contains invalid characters, depending on fs type), and have the check there.

PR #448, opened 2018-11-01
Fix tilemap camera issues

jwunderl:
Fixes issues when tile map is smaller than screen. This issue comes from Math.clamp being passed a value such that ``max < min``, bumping the camera and drawing in the negative direction / not matching the location of the tiles for collisions.

Additionally, remove the default 1 pixel _map, and only 'create' the _map when it is assigned from ``setMap``. This primarily comes up when setting up a map before using it, like [this example](https://makecode.com/_YiedK43MeV3X). This is most apparent in the [simple maze](https://arcade.makecode.com/tutorials/simple-maze) tutorial, as steps 4 and 5 will temporarily break the game.

Example:

[example game](https://makecode.com/_3iCPPuEzrXpV): tilemap is 5*8, making it half the width of the screen.

### fixed

Question: should it be possible to disable tile maps? That would allow for situations like 'over world' maps where players can enter and leave houses / dungeons that are tile maps. I can add that in here if it sounds like a decent idea.

### DIFF
--- a/libs/game/tilemap.ts
+++ b/libs/game/tilemap.ts
@@ -61,24 +61,18 @@ namespace tiles {
         private _tileSets: TileSet[];
 
         constructor() {
-            this._map = img`1`;
             this._tileSets = [];
             this._layer = 1;
 
             this.z = -1;
-
-            const sc = game.currentScene();
-            sc.allSprites.push(this);
-            sc.flags |= scene.Flag.NeedsSorting;
-            this.id = sc.allSprites.length;
         }
 
         offsetX(value: number) {
-            return Math.clamp(0, (this._map.width << 4) - screen.width, value);
+            return Math.clamp(0, Math.max(this.areaWidth() - screen.width, 0), value);
         }
 
         offsetY(value: number) {
-            return Math.clamp(0, (this._map.height << 4) - screen.height, value);
+            return Math.clamp(0, Math.max(this.areaHeight() - screen.height, 0), value);
         }
 
         areaWidth() {
@@ -105,7 +99,15 @@ namespace tiles {
         }
 
         setMap(map: Image) {
-            this._map = map;
+            if (!this._map) {
+                const sc = game.currentScene();
+                sc.allSprites.push(this);
+                sc.flags |= scene.Flag.NeedsSorting;
+                this.id = sc.allSprites.length;
+            }
+            if (map) {
+                this._map = map;
+            }
         }
 
         public getTile(col: number, row: number): Tile {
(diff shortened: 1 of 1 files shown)


riknoll:
Can we rename the function to `scene.setTileMapEnabled()`?

libs/game/scenes.ts
```diff
@@ -104,6 +104,17 @@ namespace scene {
         scene.tileMap.setMap(map);
     }
 
+    /**
+     * Set whether the current Tile Map should be enabled (used) or not (ignored).
+     * @param enabled
+     */
+    export function setTileMapEnabled(enabled: boolean) {
+        const scene = game.currentScene();
+        if (!scene.tileMap)
```
pelikhan: don't allocate with enabled = false?

jwunderl: I suppose this shouldn't allocate in either case. (setting the map image for the first time will set enabled to true anyway)

libs/game/tilemap.ts
```diff
@@ -105,7 +100,16 @@ namespace tiles {
         }
 
         setMap(map: Image) {
-            this._map = map;
+            if (!this._map) {
```
pelikhan: what if map is undefined?

jwunderl: It gets handled below in line 110, but I should just check that first and return early in that case; no reason to set up if there is no map

libs/game/tilemap.ts
```diff
@@ -105,6 +100,14 @@ namespace tiles {
         }
 
         setMap(map: Image) {
+            if (!map) return;
```
pelikhan: if map is undefined, you should clear the current _map and disable it?

libs/game/tilemap.ts
```diff
@@ -100,7 +100,12 @@ namespace tiles {
         }
 
         setMap(map: Image) {
-            if (!map) return;
+            if (!map) {
+                this._map = undefined;
```
pelikhan: is ``!!this._map`` equivalent to ``this.enabled`` ?

pelikhan: ok - is this a scenario?

jwunderl: I suppose the ones where I'd want to use it are somewhat better used with ``game.pushScene`` (e.g. a tilemap town where you can enter a house and free roam, or using setTileAt), and otherwise it's not a big cost to just set map to undefined and reset later. I'll remove it

pelikhan:
Memory is expensive, compute is cheap.

libs/game/tilemap.ts
```diff
@@ -98,6 +96,10 @@ namespace tiles {
             }
         }
 
+        enabled(): boolean {
```
pelikhan: you can make this a property ``get enabled(): boolean``

jwunderl: For some reason I blanked on that, thanks

PR #452, opened 2018-11-05
Fixes Sprite ID collision issues

ChaseMor:
Added a sprite counter to distribute unique sprite IDs.

Previously, two sprites can have the same ID since their ID is based on how many sprites were in the scene when they were created.

Fixes Microsoft/pxt-arcade#333

### DIFF
--- a/libs/game/scene.ts
+++ b/libs/game/scene.ts
@@ -29,6 +29,7 @@ namespace scene {
         background: Background;
         tileMap: tiles.TileMap;
         allSprites: SpriteLike[];
+        spriteCount: number;
         spritesByKind: Sprite[][];
         physicsEngine: PhysicsEngine;
         camera: scene.Camera;
@@ -55,6 +56,7 @@ namespace scene {
             if (this.allSprites) return;
 
             this.allSprites = [];
+            this.spriteCount = 0;
             scene.setBackgroundColor(0)
             // update controller state
             this.eventContext.registerFrameHandler(8, () => {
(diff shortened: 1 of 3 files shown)


libs/game/scene.ts
```diff
@@ -115,12 +115,5 @@ namespace scene {
             // update screen
             this.eventContext.registerFrameHandler(200, control.__screen.update);
         }
-
-        addSprite(sprite: SpriteLike) {
```
pelikhan: hummm. i think we can to keep this. spriteCount and "allSprites" are internal field to this class.

libs/game/scene.ts
```diff
@@ -115,5 +115,10 @@ namespace scene {
             // update screen
             this.eventContext.registerFrameHandler(200, control.__screen.update);
         }
+
+        addSprite(sprite: SpriteLike) {
```
ChaseMor: @pelikhan How is this?

libs/game/scene.ts
```diff
@@ -29,6 +29,7 @@ namespace scene {
         background: Background;
         tileMap: tiles.TileMap;
         allSprites: SpriteLike[];
+        spriteNextId: number;
```
pelikhan: make this private to enfore encapsulation

riknoll:
That last commit added an error, but I guess our test suite doesn't actually test?

riknoll:
@ChaseMor has a fix

PR #454, opened 2018-11-06
Add Serial read (NOT MERGE)

sabas1080:
Error


libs/airQuality/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
pelikhan: please add those as separate packages.

pelikhan:
Please clean out the additional packages from this PR>

PR #455, opened 2018-11-06
Init serial read function (NOT MERGE)

sabas1080:
Hi

I'm trying to create the functions of read in the serial, but I have some problems in the part of typescript

```
Type '0' is not assignable to type 'string'.
shims.d.ts(11,14): error TS2384: Overload signatures must all be ambient or non-ambient.
shims.d.ts(19,14): error TS2384: Overload signatures must all be ambient or non-ambient.
```

Thanks

### DIFF
--- a/libs/serial/serial.cpp
+++ b/libs/serial/serial.cpp
@@ -59,25 +59,14 @@ SINGLETON(WSerial);
 
 namespace serial {
 
-      /**
-     * Read a line of text from the serial port and return the buffer when the delimiter is met.
-     * @param delimiter text delimiter that separates each text chunk
-     */
-    //% help=serial/read-until
-    //% blockId=serial_read_until block="serial|read until %delimiter=serial_delimiter_conv"
-    //% weight=19
+    //%
     String readUntil(String delimiter) {
       auto f = getWSerial()->serial.readUntil(delimiter->data);
       auto res = mkString(f.toCharArray(), f.length());
       return res;
     }
 
-    /**
-    * Read the buffered received data as a string
-    */
-    //% help=serial/read-string
-    //% blockId=serial_read_buffer block="serial|read string"
-    //% weight=18
+    //%
     String readString() {
       int n = getWSerial()->serial.getRxBufferSize();
       if (n == 0) return mkString("", 0);
--- a/libs/serial/serial.cpp
+++ b/libs/serial/serial.cpp
@@ -75,12 +75,7 @@ namespace serial {
       return res;
     }
 
-    /**
-    * Register an event to be fired when one of the delimiter is matched.
-    * @param delimiters the characters to match received characters against.
-    */
-    //% help=serial/on-data-received
-    //% weight=18 blockId=serial_on_data_received block="serial|on data received %delimiters=serial_delimiter_conv"
+    //%
     void onDataReceived(String delimiters, Action body) {
       getWSerial()->serial.eventOn(delimiters->data);
       registerWithDal(DEVICE_ID_SERIAL, CODAL_SERIAL_EVT_DELIM_MATCH, body);
--- a/libs/serial/serial.cpp
+++ b/libs/serial/serial.cpp
@@ -27,6 +27,21 @@ enum class BaudRate {
   BaudRate300 = 300
 };
 
+enum Delimiters {
+    //% block="new line"
+    NewLine = 1,
+    //% block=","
+    Comma = 2,
+    //% block="$"
+    Dollar = 3,
+    //% block=":"
+    Colon = 4,
+    //% block="."
+    Fullstop = 5,
+    //% block="#"
+    Hash = 6,
+};
+
 namespace pxt {
   class WSerial {
     public:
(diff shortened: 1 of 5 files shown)


libs/serial/serial.cpp
```diff
@@ -43,6 +58,47 @@ SINGLETON(WSerial);
...
 
 namespace serial {
+
+      /**
+     * Read a line of text from the serial port and return the buffer when the delimiter is met.
+     * @param delimiter text delimiter that separates each text chunk
+     */
+    //% help=serial/read-until
+    //% blockId=serial_read_until block="serial|read until %delimiter=serial_delimiter_conv"
+    //% weight=19
```
pelikhan: move these annotations to the .ts file and leave a //%

libs/serial/sim/serial.ts
```diff
@@ -1,4 +1,12 @@
 namespace pxsim.serial {
+    export function readUntil(delimiter: string) {
+
```
pelikhan: return empty line at least

sabas1080:
@pelikhan thanks for comments, I have made some changes but still I can not receive data

pelikhan:
What is your test?

sabas1080:
ok, the blink does not stop anymore, but I do not get anything

pelikhan:
Did you try compiling a pure coral sample to see if the serial c++ is working?

pelikhan:
Does it lock on readString?

sabas1080:
no, blink ok, but not read, apparently do not pass the data from C to typescript

sabas1080:
my test is now

```
let dato = ""
serial.setBaudRate(BaudRate.BaudRate9600)
forever(function () {
    dato = serial.readString()
    console.log(dato)
    pins.LED.digitalWrite(true)
    pause(500)
    pins.LED.digitalWrite(false)
    pause(500)
})
```

makecode maker blocks javascript editor 1

pelikhan:
Skip console, do serial.writeLine instead. I need to find the docs to attach the debugger.

sabas1080:
but without console, no "console device", does it come from the device or not? and it works with webusb o no?

pelikhan:
Console.log should work with WebUSB always.  You should add more logging to trace down where it breaks down. Is it not receiving any bytes?

sabas1080:
I only receive that is in typescript

https://github.com/ElectronicCats/pxt-common-packages/blob/serial-read-test/libs/serial/serial.ts#L59

example if change "" for "a" receive "a"

pelikhan:
Ha! You are missing shim=serial::readString !

sabas1080:
Oh!

and if it is already in shims.d.s

https://github.com/ElectronicCats/pxt-common-packages/blob/serial-read-test/libs/serial/shims.d.ts#L18

no matter? Or what happens?

pelikhan:
You should not need that one anymore - definitely wrong. Something is wrong with the notation. I think you need the shim notation on the c++ function comment.


libs/serial/serial.ts
```diff
@@ -37,5 +37,55 @@ namespace serial {
...
     }
+    
+    /**
+     * Read a line of text from the serial port and return the buffer when the delimiter is met.
+     * @param delimiter text delimiter that separates each text chunk
+     */
+    //% help=serial/read-until
+    //% blockId=serial_readUntil block="serial|read until %delimiter=serial_delimiter_conv"
+    //% weight=19
+    export function readUntil(del: string): string {
```
pelikhan: on a second though, you don't need the TypeScript implementatiuon fro this. Move the //% back to the C++ function and delete this function

libs/serial/serial.ts
```diff
@@ -37,5 +37,55 @@ namespace serial {
...
+        return readString();
+    }
+    
+    /**
+    * Read the buffered received data as a string
+    */
+    //% help=serial/read-string
+    //% blockId=serial_read_buffer block="serial|read string"
+    //% weight=18
+    export function readString(): string {
```
pelikhan: same here, delete

libs/serial/serial.ts
```diff
@@ -37,5 +37,55 @@ namespace serial {
...
+        return ""
+    }
 
+    /**
+    * Register an event to be fired when one of the delimiter is matched.
+    * @param delimiters the characters to match received characters against.
+    */
+    //% help=serial/on-data-received
+    //% weight=18 blockId=serial_on_data_received block="serial|on data received %delimiters=serial_delimiter_conv"
+    export function onDataReceived(delimiters: string, body: () => void): void{
```
pelikhan: same here delete

pelikhan:
You will want to add function implementations in /libs/serial/sim/serial.ts

sabas1080:
with the changes when starting the simulator

`warning: Program Error: Cannot read property 'vtable' of undefined`

sabas1080:
@pelikhan 
I have managed to send C ++ information to typescript, but there is no reception

https://github.com/ElectronicCats/pxt-common-packages/blob/serial-read-test/libs/serial/serial.cpp#L73

but the condition is also not met

https://github.com/ElectronicCats/pxt-common-packages/blob/serial-read-test/libs/serial/serial.cpp#L72

libs/serial/serial.cpp
```diff
@@ -69,10 +74,10 @@ namespace serial {
     //%
     String readString() {
       int n = getWSerial()->serial.getRxBufferSize();
-      if (n == 0) return mkString("", 0);
-      auto f = getWSerial()->serial.read(n, SerialMode::ASYNC);
-      auto res = mkString(f.toCharArray(), f.length());
-      return res;
+      if (n == 0) return mkString("null", 4);
```
pelikhan: why?

sabas1080: test just to know if it enters there

sabas1080:
Blink for ok, but in read.String break and block of chips
My code example is
```
let da = ""
serial.attachToConsole()
serial.setBaudRate(BaudRate.BaudRate9600)
forever(function () {
    for (let i = 0; i < 4; i++) {
        pins.LED.digitalWrite(true)
        pause(500)
        pins.LED.digitalWrite(false)
        pause(500)
        pins.LED.digitalWrite(true)
        pause(500)
        pins.LED.digitalWrite(false)
        pause(500)
    }
    da = serial.readString()
})
```

libs/serial/serial.cpp
sabas1080: never enter this point

sabas1080: buffer with "garbage"?

PR #459, opened 2018-11-08
adding serial readstring

pelikhan:
Just add serial.readstring

### DIFF
--- a/libs/serial/serial.cpp
+++ b/libs/serial/serial.cpp
@@ -43,6 +43,26 @@ SINGLETON(WSerial);
 }
 
 namespace serial {
+    #define MSTR(s) codal::ManagedString((s)->data, (s)->length)
+
+    static inline String PSTR(codal::ManagedString s) {
+        return mkString(s.toCharArray(), s.length());
+    }
+
+
+    /**
+    * Read the buffered received data as a string
+    */
+    //% help=serial/read-string
+    //% blockId=serial_read_buffer block="serial|read string"
+    //% weight=18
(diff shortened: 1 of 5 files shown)


pelikhan:
Do we already have macros to convert ManagedString to BoxedString* ?

libs/serial/serial.cpp
```diff
@@ -43,6 +43,26 @@ SINGLETON(WSerial);
...
+    * Read the buffered received data as a string
+    */
+    //% help=serial/read-string
+    //% blockId=serial_read_buffer block="serial|read string"
+    //% weight=18
+    String readString() {
+      auto service = getWSerial();
+      int n = service->serial.getRxBufferSize();
+      if (n == 0) return mkString("", 0);
+      return PSTR(service->serial.read(n, SerialMode::ASYNC));
```
mmoskal: I think to be compliant with C++ memory mgmt, you need to do something like:
```
auto s = return PSTR(service->serial.read(n, SerialMode::ASYNC));
return PSTR(s);
```

Otherwise, the compiler might deallocate return value from serial read before you copy it.

mmoskal:
I don't think we already have macros

libs/serial/serial.cpp
pelikhan: you mean?
```
auto s = service->serial.read(n, SerialMode::ASYNC);
return PSTR(s);
```

mmoskal: Yes, sorry

PR #463, opened 2018-11-13
Sprite overlap events: async with drop when running

jwunderl:
fix second to last bullet point from #423

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -101,16 +101,17 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                 const oh = sprite.overlapHandler;
                 if (oh)
                     control.runInParallel(() => oh(tmp))
-                scene.overlapHandlers
-                    .filter(h => h.type == sprite.type && h.otherType == overlapper.type)
-                    .filter(h => h.current.indexOf(tmpsprite) == -1 && h.current.indexOf(tmp) == -1)
-                    .forEach(h => control.runInParallel(() => {
-                        h.current.push(tmp);
-                        h.current.push(tmpsprite);
-                        h.handler(tmpsprite, tmp);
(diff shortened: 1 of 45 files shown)


libs/game/physics.ts
```diff
@@ -97,15 +97,27 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
+                            higher._overlappers.push(lower.id);
+                            sprite.overlapHandler(overlapper);
+                            higher._overlappers.removeElement(lower.id);
+                        });
+                    }
+
+                    scene.overlapHandlers
+                        .filter(h => h.type == sprite.type && h.otherType == overlapper.type)
+                        .forEach(h => control.runInParallel(() => {
+                            higher._overlappers.push(lower.id);
```
riknoll: Also needs to be hoisted out of `runInParallel()`.

libs/game/physics.ts
```diff
@@ -97,15 +97,29 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
-                    .forEach(h => h.handler(tmpsprite, tmp));
+                // Maintaining invariant that the sprite with the higher ID has the other sprite as an overlapper
+                const higher = sprite.id > overlapper.id ? sprite : overlapper;
+                const lower = higher === sprite ? overlapper : sprite;
+
+                if (higher._overlappers.indexOf(lower.id) === -1) {
+                    if (sprite.overlapHandler) {
+                        higher._overlappers.push(lower.id);
+                        control.runInParallel(() => {
+                            sprite.overlapHandler(overlapper);
```
mmoskal: isn't `sprite` overridden by the loop? (capture of for/of argument)

mmoskal: nvm, in JS it shouldn't be, let me see if we have a bug ;)

PR #465, opened 2018-11-13
add fast random mathUtil class

jwunderl:
Fast random class @riknoll

### DIFF
--- a/libs/game/mathUtil.ts
+++ b/libs/game/mathUtil.ts
@@ -28,4 +28,62 @@ namespace Math {
         }
         return list[Math.randomRange(0, list.length - 1)];
     }
+
+    /**
+     * Fast, 16 bit, seedable (pseudo) random generator.
+     */
+    export class FastRandom {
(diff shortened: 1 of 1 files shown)


libs/game/mathUtil.ts
```diff
@@ -28,4 +28,62 @@ namespace Math {
...
+            return this.lfsr = (this.lfsr >> 1) ^ ((-(this.lfsr & 1)) & 0xb400);
+        }
+
+        /**
+         * @param min the minimum value to generate
+         * @param max the maximum value to generate
+         * @returns a random value between min and max (inclusive)
+         */
+        randomRange(min: number, max: number): number {
+            return min + this.next() % (max - min);
```
riknoll: only positive numbers right?

jwunderl: if

```
min == -60
max == 0
this.next() == 42
```

then this will return ``-60 + 42 % (0 -  -60)`` which would be -18 which works, right?

I suppose I'm missing the invariant that ``min >= max``, though

PR #467, opened 2018-11-13
multi-channel sound

mmoskal:
This adds `mixer` packages. Sound output works on STM and in the sim.

JacDac bits might need to be synchronized with https://github.com/Microsoft/pxt-common-packages/pull/409


mmoskal:
legacy.ts should be in pxt.json, will fix

I think we will keep music for platform with only PWM (lego etc)

mmoskal:
I guess I should test the blocks as well...

PR #469, opened 2018-11-14
cross-compat changes for serial

pelikhan:
Ensure serial in microbit is compatiable with this one.


sabas1080:
@pelikhan I can try this in pxt-maker if you like

PR #470, opened 2018-11-14
Fix onGesture jsdoc description

ganicke:
Too many "when"s in the description.

RE: https://github.com/Microsoft/pxt-adafruit/issues/849


ganicke:
Yea, I like 'happens' better.

PR #472, opened 2018-11-14
JacDac console

pelikhan:
2 services for jacdac:
* logger sends console log messages to jacdac
* log listener receives and pipes console log messages

Also added a "ConsolePriority" concept in the current console.


libs/base/console.ts
```diff
@@ -1,15 +1,49 @@
...
+     */
+    export let minPriority = ConsolePriority.Log;
 
     //% whenUsed
-    const listeners: Listener[] = [function(text: string) { control.__log(text); }];
+    const listeners: Listener[] = [function(priority: ConsolePriority, text: string) { control.__log(text); }];
+
+    export function add(priority: ConsolePriority, text: string) {
+        if (priority < minPriority) return;
+        // pad text on the 32byte boundar
```
mmoskal: This doesn't do that

libs/base/console.ts
```diff
@@ -1,15 +1,49 @@
...
+        text += "\r\n";
```
mmoskal: What's wrong with just LF?

PR #473, opened 2018-11-15
adding menu

pelikhan:
Adds a basic system menu that is able to display a console view. Needs cleanup but this is a first drop.


riknoll:
I added jsdoc here: https://github.com/riknoll/pxt-arcade-packages/blob/iso-app/menu/menu.ts

Sorry, meant to put that in master.

PR #476, opened 2018-11-17
Neopixel DMA and SAMD21 no mbed fixes

mmoskal:


libs/light/neopixel.cpp
```diff
@@ -1,36 +1,89 @@
 #include "pxt.h"
 #include "neopixel.h"
 
+#ifndef NEOPIXEL_SPI
+#define NEOPIXEL_SPI 1
```
pelikhan: is it always on now?

mmoskal: The idea is that it's disabled unless you do `strip.spi = true`. Additionally, `platform.h` can disable it all together.

PR #481, opened 2018-11-19
Reword controlSprite and reorder toolbox

jwunderl:
Reword sprites.controlSprite so that it's a bit easier to understand for new developers; ``move {mySprite} with buttons`` and give vx / vy default values so they can be suppressed.

Reorder toolbox so that game is below controller ( closes Microsoft/pxt-arcade#349 )

Todo:

* [ ] authoring so that ``vx`` and ``vy`` pop out with the ``+`` at the same time, like ``motors.largeA.run`` in lego
* [ ] upgrade path from controlSprite -> moveSprite to avoid breaking everyone who's used it so far with ts change (should be this I think: https://github.com/jwunderl/pxt-arcade/commit/b96d56128265b5771f5816036a560561427ace20)

Also, since this will require an upgrade path already, I can add in the changes to tilemap apis here if we can agree on something ( https://github.com/Microsoft/pxt-arcade/issues/334 )

### DIFF
--- a/libs/controller/controller.ts
+++ b/libs/controller/controller.ts
@@ -10,7 +10,7 @@ enum ControllerButtonEvent {
 /**
  * Access to game controls
  */
-//% weight=97 color="#e15f41" icon="\uf11b"
+//% weight=98 color="#e15f41" icon="\uf11b"
 namespace controller {
     let _userEventsEnabled = true;
     let _activeButtons: Button[];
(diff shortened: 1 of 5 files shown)


jwunderl:
Err, I was meaning to wait on the first todo, need to fix related docs in here and arcade so arcade build doesn't break with next bump. Will make a pair of PRs to fix all that, one sec

PR #488, opened 2018-11-20
Fix move sprite docs

jwunderl:
broken due to move from ``controlSprite`` to ``moveSprite`` with https://github.com/Microsoft/pxt-common-packages/pull/481.

Will need a bump to 5.1.7 to make everything match in arcade as well


pelikhan:
I'll bump

PR #493, opened 2018-11-21
Prevent recursive panic

pelikhan:
Store first panic code and prevent recursion.


libs/screen---st7735/screen.cpp
```diff
@@ -131,18 +133,22 @@ static void drawPanic(int code) {
 }
 
 extern "C" void target_panic(int statusCode) {
+    if (panicCode)
```
mmoskal: statusCode can be zero, maybe a bool flag would be better

PR #517, opened 2018-12-03
Fixed bug with projectiles and tile maps

ChaseMor:
Previously, projectiles are placed by default on the edge of the screen (when a velocity is specified) such that it isn't fully shown, but rather moves into frame.

This is problematic when tile maps are used since they will be instantly detected as hitting the wall and will be destroyed.

This places the sprites fully on the screen when a tile map is present to prevent them from initially colliding with a tile map wall.

### DIFF
--- a/libs/game/sprites.ts
+++ b/libs/game/sprites.ts
@@ -65,20 +65,25 @@ namespace sprites {
     //% expandableArgumentMode=toggle
     export function createProjectile(img: Image, vx: number, vy: number, kind: number, sprite?: Sprite) {
         const s = sprites.create(img, kind);
+        const sc = game.currentScene();
         s.vx = vx
         s.vy = vy
 
         // put it at the edge of the screen so that it moves towards the middle
+        // If the scene has a tile map, place the sprite fully on the screen
+
+        const xOff = sc.tileMap ? -(s.width >> 1) : (s.width >> 1) - 1;
+        const yOff = sc.tileMap ? -(s.height >> 1) : (s.height >> 1) - 1;
 
         if (vx < 0)
-            s.x = screen.width + (s.width >> 1) - 1
+            s.x = screen.width + xOff
         else if (vx > 0)
-            s.x = -(s.width >> 1) + 1
+            s.x = -xOff
 
         if (vy < 0)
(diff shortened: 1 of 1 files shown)


pelikhan:
I found projectiles to be very confusing. First they don't move, then they position themselves randomly once you change the velocity. IMO, we should remove this magic and let the user position them "manually".

ChaseMor:
To be honest, I can see strong arguments for both sides. On one hand, it does make it more complicated and can seem like magic. It can be hard for a student to recognize what is going on and they will be making games that they don’t fully understand.

On the other hand, projectiles are a special type of sprite. Having just the flags and velocity set might not be enough for them to warrant their own block. These are simple things that a user would probably understand and they would just be using projectile blocks to make their code shorter and simpler. 

I agree the placement of projectiles is non-trivial for a beginner and they probably wouldn’t be able to create it themselves. I think the crux of the decision is whether or not we want to include blocks that can be created using other blocks, but beginners might not know how to? Do we want to allows beginners to do this because it allows them to make more advanced and interesting games, or do we not want to since they don’t fully understand what’s going on in the background?
From what I can gather, the only other blocks that do this are moveSprite and one could make an argument for sprites.allOfKind.

jwunderl:
But ``moveSprite`` and ``sprites.allOfKind`` do exactly what they say; make a sprite move with the arrow keys / buttons, and return an array of sprites. The exact implementation doesn't particularly matter, but there aren't really unexpected side effects that would surprise the user; changing the ``vx`` on ``moveSprite`` doesn't make the sprite change colors when left is pressed.

``projectiles`` do have a bit of that; you drag it into the project and it is in the center like normal sprites, then you set a velocity and all of a sudden it is on the side of the screen and you don't know why. You have to read the comment (which is a bit much to expect) or be told *why* it works that way, which is a bit different than not understanding *how* it works yet.

pelikhan:
**I think this block should be removed all together.** It has way to much hidden semantics that will confuse the kids. They can learn to create a sprite, make it auto-destroy, give it a speed and position.

ChaseMor:
@pelikhan Should we merge this to prevent current bugs while we figure out what to with projectiles?

pelikhan:
Yes

PR #519, opened 2018-12-04
reword scene tilemap apis

jwunderl:
Fixes microsoft/pxt-arcade#334 (see issue and comments for reasoning on change)

Once we confirm what the apis should be I'll make the changes to the documentation here and in arcade, and add a patch to fix existing scripts


jwunderl:
Yup, was looking for some discussion on this, just need to nail it down so I can fix the courses and all as needed.

Updated to match the discussion in slack, but not sure on the javascript name; maybe setTilesByColor or something of the sort?

PR #522, opened 2018-12-04
various arcade tweaks

pelikhan:
- [x] use yes/no in gameover
image
- [x] change default speed of projectile to vy =100
- [x] rename sprite property type -> kind, add dedicated blocks under overlaps to set/read kind
image
- [x] organize info toolbox
image


abchatra:
There nice set of changes. Thanks Peli

PR #524, opened 2018-12-04
Move RefCollection data to GC heap

mmoskal:
This pre-allocates the entire GC heap, leaving 12k for codal. Codal seems to need around 10k in Arcade. This can be tweaked with `SYSTEM_HEAP_BYTES` config.

Also, add support for simulating low-memory devices, for example add the following:
```
namespace config {
    export const LOW_MEM_SIMULATION_KB = 64
}
```
To simulate 64k memory device.

From now on, panic 20 is out of codal memory, panic 21 is out of GC memory. After 21 you can do `pxt heap` to dump the heap.


mmoskal:
Needs `codal-core` `3cb6adcd3c40ab8085e229c2833cd48030032876` or later

PR #526, opened 2018-12-04
Increase default system memory to 16k (needed for mixer)

mmoskal:


libs/base/gc.cpp
```diff
@@ -280,7 +280,7 @@ __attribute__((noinline)) static void allocateBlock() {
         target_panic(PANIC_GC_OOM);
     }
     auto lowMem = getConfig(CFG_LOW_MEM_SIMULATION_KB, 0);
-    auto sysHeapSize = getConfig(CFG_SYSTEM_HEAP_BYTES, 12 * 1024);
+    auto sysHeapSize = getConfig(CFG_SYSTEM_HEAP_BYTES, 16 * 1024);
```
pelikhan: should this be a constant that can be modified via project yotta config?

mmoskal:
It can already be modified in config.ts, not sure we need more.

PR #531, opened 2018-12-05
Better support for systemmenu

pelikhan:
@riknoll please review what i am doing to your UI system :)

### DIFF
--- a/libs/game/menu.ts
+++ b/libs/game/menu.ts
@@ -878,6 +878,29 @@ namespace menu {
         flow: VerticalFlow;
         items: ListItem[];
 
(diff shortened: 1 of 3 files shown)


libs/game/menu.ts
```diff
@@ -1166,7 +1216,9 @@ namespace menu {
                 .to(screen.width - this.margin)
                 .duration(200)
                 .onEnded(() => {
+                    console.log(`show list`)
```
riknoll: Looks like you're using `log()` instead of `console.log()` everywhere else

pelikhan: Yes, it is a menu specific log that I can turn on/off

PR #537, opened 2018-12-06
regorg music blocks

pelikhan:
image
image

### DIFF
--- a/libs/mixer/melody.ts
+++ b/libs/mixer/melody.ts
@@ -30,6 +30,7 @@ namespace music {
     //% volume.min=0 volume.max=256
     //% help=music/set-volume
     //% weight=70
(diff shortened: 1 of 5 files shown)


pelikhan:
@ganicke melody stop needs docs.

ganicke:
Fixed instances only for sounds? Is this a temporary thing? You'll need a screen full of playTones to make your own sounds (melodies).

mmoskal:
that's a good point - we could have a block that plays a string. In TS you can do: `new Melody("a5 ...").play()`

However, the current music blocks also have no way to play a string

ganicke:
Some ``music.createMelody("")`` melody factory. Good way to stress your GC.

PR #540, opened 2018-12-07
use GC heap for almost all allocations

mmoskal:
This replaces codal's implementations of malloc/free with our own, which will generally use our GC heap for allocations, unless we're inside of an interrupt or in the middle of a GC cycle.

There are also changes to limit fragmentation of the GC heap (we generally try to fit allocations within the first half of free memory, and force GC when that is not possible).

Requires codal-core `b059fa81208655a8094283a3110810087a5a8fa4`

Requires https://github.com/Microsoft/pxt/pull/5068


mmoskal:
Waiting on https://github.com/lancaster-university/codal-core/pull/59

PR #543, opened 2018-12-07
Fix melodies without specified beats

jwunderl:
Right now nothing plays when you use it in arcade


pelikhan:
What about all the other melodies? Why is this broken? Is the default tempo not set?

jwunderl:
The others work because the notes that omit the number of beats don't have the octave after them (e.g. in `jumpDown` the notes are `'g5:1 f e d c'`, where g5 is assigned 1 explicitly and the other notes are just the letter). That should be easy to make work I suppose (at a glance it looks like just the switch here https://github.com/Microsoft/pxt-common-packages/blob/5c08f6413a078576e1126e4cd05e58fdac087496/libs/mixer/melody.ts#L318-L325 would have to account for it). I can add that in real quick if that's preferred

PR #544, opened 2018-12-07
added feature to render to svg

pelikhan:
Refactoring useful code to render canvas into svg as a datauri.


riknoll:
Wait, how is this an SVG? Isn't it just an SVG that is embedding a PNG?

pelikhan:
It's an svg:image element that gets updated in the SVG tree.

riknoll:
`toDataURL()` returns an encoded PNG. LGTM but I would rename the function because it's confusing.

pelikhan:
Renamed.

PR #556, opened 2018-12-12
Projectile

pelikhan:
* split out projectile blocks
image
* uses a random velocity is vx,vy are 0

### DIFF
--- a/libs/game/sprites.ts
+++ b/libs/game/sprites.ts
@@ -58,12 +58,43 @@ namespace sprites {
      * The sprite auto-destroys when it leaves the screen. You can modify position after it's created.
      */
     //% group="Create"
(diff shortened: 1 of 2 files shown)


mmoskal:
Are we sure kind==1 is Projectile?

libs/game/sprites.ts
```diff
@@ -57,13 +57,44 @@ namespace sprites {
...
+    }
+
+
+    
+    /**
+     * Create a new sprite with given speed, and place it at the edge of the screen so it moves towards the middle.
+     * The sprite auto-destroys when it leaves the screen. You can modify position after it's created.
+     */
+    //% group="Projectiles"
+    //% blockId=spritescreateprojectilefromsprite block="projectile %img=screen_image_picker from %mySprite=variables_get with vx %vx vy %vy"
```
abchatra: Should %mysprite be %sprite?

pelikhan: Missed it will fix

PR #558, opened 2018-12-12
init commit from arcade, needs heavy linting / reorg

pelikhan:


jwunderl:
I was planning on linting and changing this a bit more before sending it up, I'll clean it up a bit and move over to a separate package (outside of game) when I'm done working tonight :)

pelikhan:
On a second though, they should definitely be integrated into the sprite api

pelikhan:
@abchatra @riknoll @mmoskal i'm thinking that particles could live in their package and get a dedicated toolbox section when imported. toughts?

pelikhan:
can you move the particles to a separate package? it will give us more time to evolve them.

riknoll:
Yeah, separate package for sure

abchatra:
Yes agree

jwunderl:
If you want it under sprites, maybe a subcategory that pops out, like ``neopixel`` pops out of ``light`` in adafruit? Sprites is very overloaded blocks wise right now

Definitely was planning on moving stuff around and all, just wanted to fix the front page projectiles first

PR #563, opened 2018-12-14
remove createEmptySprite

pelikhan:
Fix for https://github.com/Microsoft/pxt-arcade/issues/385


jwunderl:
I'll handle the places it is referred to / used in the docs later tonight

PR #571, opened 2018-12-14
Don't immediately destroy particles on ParticleSource destroyed

jwunderl:
On ParticleSource.destroy, leave particles alive until their lifespan runs out

Obligatory example:

https://makecode.com/_Wvhef4EgPfAp

```typescript
enum SpriteKind {
    Player,
    Projectile
}

controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
    const s = sprites.createProjectileFromSide(img`
    2
    2
    `, 0, -100);
    s.data = new particles.ParticleSource(s, 50, new particles.SprayFactory(30, 0, 50));
    s.x = Math.randomRange(0, screen.width);
});

sprites.onDestroyed(SpriteKind.Projectile, (s) => {
    if (s.data) (s.data as particles.ParticleSource).destroy();
});
```

### DIFF
--- a/libs/particles/particles.ts
+++ b/libs/particles/particles.ts
@@ -199,8 +199,12 @@ namespace particles {
          * Destroy the source
          */
         destroy() {
-            sources.removeElement(this);
-            game.currentScene().allSprites.removeElement(this);
+            this.enabled = false;
+            control.runInParallel(() => {
+                pauseUntil(() => this.head == null);
+                sources.removeElement(this);
+                game.currentScene().allSprites.removeElement(this);
+            });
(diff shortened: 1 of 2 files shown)


jwunderl:
Will clean up the tests file and add some more here in a second. 

Looks good on hardware, not sure how to share a gif because teams thinks mine are just images :)

jwunderl:
Added to tests, put everything that is more than one function in separate namespaces to make it a bit more clear what the distinct cases are

space trail pong is at the bottom commented out as it requires local-multiplayer as well

PR #583, opened 2018-12-18
particle source lifespan and destroy all effects at anchor

jwunderl:
lifespan for sources for making temporary effects (e.g. a fire that goes out after 5 seconds), destroy effects at the given anchor (destroys all that match x and y).

### DIFF
--- a/libs/particles/effects.ts
+++ b/libs/particles/effects.ts
@@ -32,11 +32,8 @@ namespace particles {
     export function removeEffects(anchor: ParticleAnchor) {
         const sources = game.currentScene().data.particleSources as particles.ParticleSource[];
         if (!sources) return;
-        sources.forEach(ps => {
-            if (ps.anchor == anchor || ps.anchor.x == anchor.x && ps.anchor.y == anchor.y) {
-                ps.destroy();
-            }
-        });
+        sources.filter(ps => ps.anchor == anchor || ps.anchor.x == anchor.x && ps.anchor.y == anchor.y)
+                .forEach(ps => ps.destroy());
     }
 
     function createEffect(factoryFactory: () => ParticleFactory): ParticleEffect {
--- a/libs/particles/effects.ts
+++ b/libs/particles/effects.ts
@@ -1,4 +1,6 @@
 namespace particles {
+    let ongoingEffects: ParticleSource[] = [];
+
     //% fixedInstances
     export class ParticleEffect {
         private sourceFactory: (anchor: ParticleAnchor, pps: number) => ParticleSource;
@@ -18,7 +20,7 @@ namespace particles {
         //% group="Effects"
         start(anchor: ParticleAnchor, particlesPerSecond: number): void {
             if (!this.sourceFactory) return;
-            this.sourceFactory(anchor, particlesPerSecond);
+            ongoingEffects.push(this.sourceFactory(anchor, particlesPerSecond));
         }
     }
 
@@ -26,6 +28,26 @@ namespace particles {
         const factory = factoryFactory();
         if (!factory) return undefined;
         return new ParticleEffect((anchor: ParticleAnchor, pps: number) => new ParticleSource(anchor, pps, factory));
+    }
+
+    /**
+     * Removes all effects at anchor's location
+     * @param anchor the anchor to remove effects from
+     */
+    //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
+    //% group="Effects"
+    export function removeEffects(anchor: ParticleAnchor) {
+        const remainingEffects: ParticleSource[] = [];
+
+        ongoingEffects.forEach(ps => {
+            if (ps.anchor.x == anchor.x && ps.anchor.y == anchor.y) {
+                ps.destroy();
+            } else {
+                remainingEffects.push(ps);
+            }
+        });
+
+        ongoingEffects = remainingEffects;
     }
 
     //% whenUsed
--- a/libs/particles/effects.ts
+++ b/libs/particles/effects.ts
@@ -1,4 +1,44 @@
 namespace particles {
+
+    //% fixedInstances
+    export class ParticleEffect {
+        private sourceFactory: (anchor: ParticleAnchor, pps: number) => ParticleSource;
+
+        constructor(sourceFactory: (anchor: ParticleAnchor, particlesPerSecond: number) => ParticleSource) {
+            this.sourceFactory = sourceFactory;
+        }
+
+        /**
+         * Attaches a new particle animation to the sprite or anchor
+         * @param anchor 
+         * @param particlesPerSecond 
+         */
+        //% blockId=particlesstartanimation block="start %effect effect on %anchor=variables_get(mySprite) at rate %particlesPerSecond p/s"
+        //% particlesPerSecond.defl=20
+        //% particlesPerSecond.min=1 particlePerSeconds.max=100
+        //% group="Effects"
+        start(anchor: ParticleAnchor, particlesPerSecond: number): void {
+            if (!this.sourceFactory) return;
+            this.sourceFactory(anchor, particlesPerSecond);
+        }
+    }
+
+    /**
+     * Removes all effects at anchor's location
+     * @param anchor the anchor to remove effects from
+     */
+    //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
+    //% group="Effects"
+    export function removeEffects(anchor: ParticleAnchor) {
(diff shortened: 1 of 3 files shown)


libs/particles/effects.ts
```diff
@@ -28,6 +30,26 @@ namespace particles {
...
+     * Removes all effects at anchor's location
+     * @param anchor the anchor to remove effects from
+     */
+    //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
+    //% group="Effects"
+    export function removeEffects(anchor: ParticleAnchor) {
+        const remainingEffects: ParticleSource[] = [];
+
+        ongoingEffects.forEach(ps => {
+            if (ps.anchor.x == anchor.x && ps.anchor.y == anchor.y) {
```
pelikhan: shoulnd' you compare the reference?

jwunderl: I was making it compare based off the anchor values so that things other than sprites work nicely; if it's based off of the reference matching then you wouldn't be able to turn off the effects here after they are created

tmap

pelikhan: why not? can't you cycle through the tiles again and disable it?

jwunderl: tiles aren't cached, so the array returned by ``array of all tiles`` will be new instances of the tiles

pelikhan: comparing x/y is brittle

jwunderl: I suppose just

```typescript
ps.anchor == anchor || ps.anchor.x == anchor.x && ps.anchor.y == anchor.y
```

can handle it a bit more clearly (I can't think of a case where it would be necessary though, except if the x and y getters were to have side effects)

Ideally I think it should still work even for something like

```
for (let i = 10; i < 160; i += 20) {
    particles.fireEffect.start({ x: i, y: screen.width / 2 }, i);
}

for (let i = 10; i < 160; i += 20) {
    pause(1000);
    particles.removeEffects({ x: i, y: screen.width / 2 });
}
```

though I suppose that's up for debate

libs/particles/effects.ts
```diff
@@ -1,4 +1,6 @@
 namespace particles {
+    let ongoingEffects: ParticleSource[] = [];
```
pelikhan: this should be in the game context

libs/particles/effects.ts
pelikhan: ref test makes it clearer so yes add it.

libs/particles/effects.ts
jwunderl: Not sure I follow, where should I move that?

pelikhan: All game state should be stored In the game currentContext since we can push games on top of each other’s.

pelikhan: We'll clean this up later. The same applies to the state of multiplayer controllers. They are leaked as global but should be stored in the context.

jwunderl:
If a sprite is destroyed the particle source remains as is. As a destroyed sprite just gains the `sprites.Flag.Destroyed` flag when it is destroyed until it is garbage collected, using `removeEffects` will work even if the sprite is destroyed: e.g.

arcade-screenshot 6

pelikhan:
The sprite will not be garbage collected because it is rooted via the particle effect -> memory leak.

jwunderl:
True. I'll add a lifespan (not immediately destroy to allow effects on destroy) to the source if the anchor has 1) flags and 2) the destroyed flag is set

libs/particles/particles.ts
```diff
@@ -91,6 +101,17 @@ namespace particles {
 
         _update(dt: number) {
             this.timer -= dt;
+
+            if (this.lifespan !== undefined) {
+                this.lifespan -= dt;
+                if (this.lifespan <= 0) {
+                    this.lifespan = undefined;
```
pelikhan: destroy should unset the anchor to allow GC to collect it.

libs/particles/particles.ts
```diff
@@ -341,6 +341,7 @@ namespace particles {
     //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
     //% group="Effects"
     export function removeEffects(anchor: ParticleAnchor) {
+        if (!this.sources) return;
```
pelikhan: this does not exist here

pelikhan:
The game context should have a “data: any” getter that allows to store additional state — like “sources”. That way we don’t leak game state in the global context.

libs/particles/particles.ts
jwunderl: Yup, noticed this typo after I committed :)

jwunderl:
So `data: any = {}` in the scene class, and then this would be stored in `game.currentScene().data.particlesSources` or something of the sort?

libs/particles/effects.ts
```diff
@@ -1,4 +1,44 @@
...
+     * Removes all effects at anchor's location
+     * @param anchor the anchor to remove effects from
+     */
+    //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
+    //% group="Effects"
+    export function removeEffects(anchor: ParticleAnchor) {
+        const sources = game.currentScene().data.particleSources as particles.ParticleSource[];
+        if (!sources) return;
+        sources.forEach(ps => {
+            if (ps.anchor == anchor || ps.anchor.x == anchor.x && ps.anchor.y == anchor.y) {
```
pelikhan: javascript way: use filter

libs/particles/particles.ts
```diff
@@ -269,14 +271,16 @@ namespace particles {
...
+        const data = game.currentScene().data;
+        if (data.particleSources) return;
+        data.particleSources = [] as ParticleSource[];
         lastUpdate = control.millis();
         game.onUpdate(updateParticles);
         game.onUpdateInterval(250, pruneParticles);
     }
 
     function updateParticles() {
+        const sources = game.currentScene().data.particleSources as particles.ParticleSource[];
```
pelikhan: add helper

libs/particles/particles.ts
```diff
@@ -291,7 +295,7 @@ namespace particles {
     }
 
     function pruneParticles() {
-        const sources = game.currentScene().data.particleSources as particles.ParticleSource[];
+        const sources = particleSources();
         for (let i = 0; i < sources.length; i++) {
```
pelikhan: is sources guaranteed to exist?

jwunderl:
Should I go ahead and change `controller.moveSprite` and the `local-multiplayer` extension to store state in a similar way?

info also seems to store a decent amount in global context

pelikhan:
Let’s fix those in a future PR

libs/particles/effects.ts
```diff
@@ -22,6 +23,19 @@ namespace particles {
...
+     * Removes all effects at anchor's location
+     * @param anchor the anchor to remove effects from
+     */
+    //% blockId=particlesremoveeffect block="remove effects on %anchor=variables_get(mySprite)"
+    //% group="Effects"
+    export function removeEffects(anchor: ParticleAnchor) {
+        const sources = game.currentScene().data.particleSources as particles.ParticleSource[];
+        if (!sources) return;
+        sources.filter(ps => ps.anchor == anchor || ps.anchor.x == anchor.x && ps.anchor.y == anchor.y)
+                .forEach(ps => ps.destroy());
```
pelikhan: remove destroyed source from sources?

libs/particles/particles.ts
```diff
@@ -91,7 +106,18 @@ namespace particles {
...
             this.timer -= dt;
-            while (this.timer < 0 && this._enabled) {
+
+            if (this.lifespan !== undefined) {
+                this.lifespan -= dt;
+                if (this.lifespan <= 0) {
+                    this.lifespan = undefined;
+                    this.destroy();
+                }
+            } else if (this.anchor && this.anchor.flags !== undefined && (this.anchor.flags & sprites.Flag.Destroyed)) {
```
pelikhan: initialize flags to 0 always

libs/particles/particles.ts
```diff
@@ -157,15 +191,15 @@ namespace particles {
         }
 
         get enabled() {
-            return this._enabled;
+            return (this.flags & Flag.enabled) != 0;
```
pelikhan: !!(this.flags & Flag.enabled)

libs/particles/effects.ts
jwunderl: The `ParticleSource` removes itself from sources when it is destroyed after the final particle from the source is destroyed in ``prune``, so that the existing particles can finish off their lifespan in peace; I'll add a quick note in destroy to indicate this.

PR #590, opened 2018-12-20
Use arg reporters for sprite events

guillaumejenkins:


pelikhan:
Changes are in arcade /beta - we won’t upgrade until the end of the contest.

PR #598, opened 2018-12-27
Auto connect to jacdac server

pelikhan:
When an arcade detects a jacdac controller on the bus, ask user if he wants to join.


pelikhan:
When connecting two arcades on the same bus, the user gets prompted to join the other game.

PR #602, opened 2018-12-29
full screen effects and disintegration, bubbles, and snow effects

jwunderl:
## normal effects

minor cleanup throughout, replace old tests

'ashes' / disintegration effect

## Full screen effects

Subclass of ParticleEffect that can occur over entire screen as background / foreground effects


pelikhan:
Background effect? Looks amazing!

jwunderl:
I think background effect does works better. Only a minor chance of misinterpreting it to have a correlation to the z index (e.g. background as opposed to foreground), but I think that's reasonable / easy to see when it occurs and fix as needed

pelikhan:
Scene effects?

libs/particles/effects.ts
```diff
@@ -53,32 +53,32 @@ namespace particles {
...
+    export class FullScreenEffect extends ParticleEffect {
         protected source: ParticleSource;
 
         constructor(sourceFactory: (anchor: ParticleAnchor, particlesPerSecond: number) => ParticleSource) {
             super(sourceFactory);
         }
 
         /**
-         * Creates a new Background Effect that occurs over entire screen
+         * Creates a new effect that occurs over theentire screen
```
pelikhan: typo

libs/particles/factories.ts
```diff
@@ -368,37 +368,50 @@ namespace particles {
         }
     }
 
+    interface ColorCount {
```
pelikhan: use a class

jwunderl:
As a side note, it's fairly trivial to add a true 'disintegration effect' (where the pixels that get chosen for the effect are removed from the source if possible) to the ashes effect, and it does look pretty cool. it would be a really quick way to run out of memory on hardware though. Still fun:

pelikhan:
Even if you make it faster and reduce the particle lifespan?

pelikhan:
is this tested on HW?

jwunderl:
Issue I was thinking of was that it would have to clone the image each time the effect is used, to 0 out the selected pixels. I ran all the games that were gifs in the original post, I'll test the variant where it removes the selected pixels and see if it does actually work

jwunderl:
Hm, just noticed hardware doesn't initialize number array contents to 0 instead of undefined like the browser does, have to be careful about that. Also, there is no array.fill which is useful when initializing to 0 isn't handled by default.

pelikhan:
File bugs on missing array functions. Slowly adding them.

pelikhan:
Ready to go?

jwunderl:
Yup!

libs/particles/effects.ts
```diff
@@ -23,20 +23,86 @@ namespace particles {
         }
     }
 
+    /**
+     * Anchor used for effects that occur across the screen.
+     */
+    class FullScreenAnchor implements ParticleAnchor {
```
pelikhan: SceneAnchor

PR #611, opened 2019-01-03
add destroy with effect method for particles

jwunderl:
method to destroy sprite with an effect

Could change wording to something like ``destroy %mySprite with %effect effect at rate %pps ...``, but that would need to be as a separate function which seems like it would be confusing in typescript

arcade-screenshot 10

cycling through a handful of the effects (I left the default rate at 20 p/s to match others for now, but this is shown with 60 p/s)

@pelikhan ~

### DIFF
--- a/libs/particles/effects.ts
+++ b/libs/particles/effects.ts
@@ -21,12 +21,29 @@ namespace particles {
             if (!this.sourceFactory) return;
             this.sourceFactory(anchor, particlesPerSecond);
         }
+
+        /**
+         * Destroy the provided sprite with
+         * @param sprite
+         * @param particlesPerSecond
+         * @param lifespan how long the sprite will remain on the screen
+         */
+        //% blockId=particlesDestroySpriteWithAnimation block="use %effect effect to destroy %anchor=variables_get(mySprite) at rate %particlesPerSecond p/s || with lifespan %lifespan"
+        //% particlesPerSecond.defl=20
+        //% particlesPerSecond.min=1 particlePerSeconds.max=100
+        //% lifespan.defl=1500
+        //% group="Effects"
+        destroy(anchor: Sprite, particlesPerSecond: number, lifespan: number = 1500) {
+            anchor.setFlag(SpriteFlag.Ghost, true);
+            this.sourceFactory(anchor, particlesPerSecond);
+            anchor.lifespan = lifespan;
+        }
     }
 
     /**
      * Anchor used for effects that occur across the screen.
      */
-    class FullScreenAnchor implements ParticleAnchor {
+    class SceneAnchor implements ParticleAnchor {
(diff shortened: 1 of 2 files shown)


pelikhan:
Feels like we should by default clear out some pixels as it gets destroyed (just randomly)

jwunderl:
Like this?

pelikhan:
Can you destroy it faster? Like in 500ms? Maybe nuke multiple pixels at once. But yes!

jwunderl:
I have the sprite lifespan as an optional parameter, will lower the default to 500 (or I can just remove and always set lifespan to 500, but I do like the ability to prolong / shorten it)

re: nuking multiple pixels at once, for each selected pixel I could just set a random rectangle (1-2 px width and height) and see how that looks

libs/particles/effects.ts
```diff
@@ -21,12 +21,44 @@ namespace particles {
...
+        //% particlesPerSecond.min=1 particlePerSeconds.max=100
+        //% lifespan.defl=500
+        //% group="Effects"
+        destroy(anchor: Sprite, particlesPerSecond: number, lifespan: number = 500) {
+            anchor.setFlag(SpriteFlag.Ghost, true);
+            this.sourceFactory(anchor, particlesPerSecond);
+            anchor.lifespan = lifespan;
+            
+            // clear out some pixels at random to give feedback on sprite being destroyed
+            const replacementImage = anchor.image.clone();
```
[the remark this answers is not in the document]
jwunderl: Should that be under particle effects? It wouldn't emit any particles so it feels a bit off to count as one to me; maybe a ``SpriteEffect`` class that has a handful of effects that can be applied to sprites (random clear pixels from this, invert colors, one layer of the melt effect that occurs on for game over, etc). I'm probably just overthinking that though

side question, I think I remember you mentioning that icons could be made for the different effects so they show up nicely instead of just a text drop down, but I don't remember how; something along the lines of making jres images like what is done for gallery sprites?

libs/particles/effects.ts
```diff
@@ -37,24 +37,30 @@ namespace particles {
...
-            }
+            dissolveSprite(anchor);
         }
     }
 
+    /**
+     * Dissolve the given sprite by removing random sections of it's image
+     * @param sprite
+     */
+    function dissolveSprite(sprite: Sprite) {
```
pelikhan: can this be a particle effect?

jwunderl: How about as an `image effect` class with the other two I mentioned ('invert' or randomize colors and one layer of the melt effect that occurs on for game over), as there aren't any particles that come out of it. Also makes it more reusable with background images / random images, and not specifically tied to sprites implementation.

also, bonus feed the ducks game:

ezgif com-optimize 2

PR #612, opened 2019-01-04
utf8 freindly serial

pelikhan:
Ended up implementing readUntil on the TS side. @mmoskal could you double check my logic when slurping characters from the serial stream?

### DIFF
--- a/libs/serial/enums.d.ts
+++ b/libs/serial/enums.d.ts
@@ -26,12 +26,32 @@ declare const enum BaudRate {
     BaudRate1200 = 1200,
     //% block=300
     BaudRate300 = 300,
-    }
+}
 
 
-    declare const enum SerialEvent {
+declare const enum SerialEvent {
     //% block="data received"
     DataReceived = 3,  // CODAL_SERIAL_EVT_RX_FULL
-    }
+}
+
+
+declare const enum Delimiters {
+    //% block="new line"
+    NewLine = 10,
+    //% block=","
+    Comma = 44,
+    //% block="$"
+    Dollar = 36,
+    //% block=":"
+    Colon = 58,
+    //% block="."
+    Fullstop = 46,
+    //% block="#"
+    Hash = 35,
+    //% block="tab"
+    Tab = 9,
+    //% block="pipe"
+    Pipe = 124,
+}
 
 // Auto-generated. Do not edit. Really.
--- a/libs/serial/enums.d.ts
+++ b/libs/serial/enums.d.ts
@@ -26,16 +26,16 @@ declare const enum BaudRate {
     BaudRate1200 = 1200,
     //% block=300
     BaudRate300 = 300,
-}
+    }
 
 
-declare const enum SerialEvent {
+    declare const enum SerialEvent {
     //% block="data received"
     DataReceived = 3,  // CODAL_SERIAL_EVT_RX_FULL
-}
+    }
 
 
-declare const enum Delimiters {
+    declare const enum Delimiters {
     //% block="new line"
     NewLine = 10,
     //% block=","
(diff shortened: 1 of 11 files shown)


libs/serial/enums.d.ts
```diff
@@ -26,12 +26,32 @@ declare const enum BaudRate {
...
+    //% block=":"
+    Colon = 58,
+    //% block="."
+    Fullstop = 46,
+    //% block="#"
+    Hash = 35,
+    //% block="tab"
+    Tab = 9,
+    //% block="pipe"
+    Pipe = 124,
```
mmoskal: add `;` and space?

libs/serial/serial.ts
```diff
@@ -29,28 +39,50 @@ namespace serial {
...
+        const start = control.millis();
+        let r = "";
+        let buf = control.createBuffer(3);
+        let bufi = 0;
+        do {
+            const c = serial.read();
+            if (c < 0) // error -- return what we have so far
+                break;
+            else if (bufi == 0 && c == delimiter) // found it!                
+                break;
```
mmoskal: Skip bufi==0 - if it's not the case you had a malformed char, which should be OK to drop - delimiters have the high bit clear whereas multi byte chars have it set

libs/serial/serial.ts
```diff
@@ -86,7 +119,11 @@ namespace serial {
...
     //% group="Write"
     export function writeValue(name: string, value: number): void {
-        const prefix = name ? name + ":" : "";
-        serial.writeLine(prefix + value);
+        if (name) {
+            writeString(name);
+            writeString(":");
+        }
+        writeNumber(value);
+        writeString("\n");
```
mmoskal: Use NEW LINE?

libs/storage/storage.cpp
```diff
@@ -75,22 +75,11 @@ snorfs::File *getFile(String filename) {
     }
     currFilename = filename;
     incrRC(currFilename);
-    currFile = filename == NULL ? NULL : st->fs.open(filename->data);
+    // TODO: fix UTF8 encoding
```
mmoskal: This seems all right to me

libs/storage/storage.cpp
```diff
@@ -75,22 +75,11 @@ snorfs::File *getFile(String filename) {
...
-*/
-//% parts="storage" 
-//% blockId="storage_append" block="append file $filename with $data"
-void append(String filename, String data) {
-    auto f = getFile(filename);
-    if (NULL == f) return;
-    f->append(data->data, data->length);
-}
-
 /**
```
mmoskal: Don't remove these - just use getUtf8data and getutf8size - they are close enough to real UTF to be used in files

libs/storage/storage.ts
```diff
@@ -3,6 +3,8 @@
 */
 //% advanced=true color=#00c0c0 icon="\uf07b"
 namespace storage {
+    export let NEW_LINE = "\r\n";
```
mmoskal: Can we just use \n?

libs/serial/enums.d.ts
pelikhan: added

libs/serial/serial.ts
pelikhan: Fixed

libs/serial/serial.ts
pelikhan: move the check to ascii case

libs/storage/storage.ts
pelikhan: removed \r

libs/storage/sim/storage.ts
```diff
@@ -4,6 +4,14 @@ namespace pxsim.storage {
         // do nothing
     }
 
+    export function append(filename: string, data: string): void {
+        const state = storageState();
+        let buf = state.files[filename];
+        if (!buf) buf = state.files[filename] = [];
+        for (let i = 0; i < data.length; ++i)
+            buf.push(data.charCodeAt(i));
```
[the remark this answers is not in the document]
pelikhan: I was trying to avoid this by moving everything to TS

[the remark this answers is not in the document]
pelikhan: I'm going back to TS to avoid bloating the simulator. We can revisit this later on.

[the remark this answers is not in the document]
mmoskal: sure. The only downside is memory usage, if the string is long and takes long to write.

PR #614, opened 2019-01-04
Removing the ability to interrupt playTone in mixer

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/489

In our old editors (i.e. micro:bit), this code:

```typescript
control.runInParallel(function () {
    music.playTone(Note.C, 1000)
})
pause(500)
music.playTone(Note.C5, 1000)
``` 

Would result in `C` playing for 500 ms and `C5` playing for 1000 ms (the second `playTone` interrupts the first).

With this change, both notes play for 1000 ms (with 500 ms of overlap).

This is a break from our old behavior, but I talked it over with @Jaqster and we think that this is a better model anyhow. Plus, it lets you do chords!

### DIFF
--- a/libs/mixer/melody.ts
+++ b/libs/mixer/melody.ts
@@ -60,34 +60,9 @@ namespace music {
     //% weight=76 blockGap=8
     //% group="Tone"
     export function playTone(frequency: number, ms: number): void {
-        if (playToneFreq == null) {
-            let buf = control.createBuffer(10 + 1)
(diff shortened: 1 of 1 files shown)


pelikhan:
I think we should rename this entire packge to "mixer" which would allow us to do breaking changes from the "music" package.

pelikhan:
RingTone relies on the 0 duration behavior. I propose to simply remove it.

PR #620, opened 2019-01-09
Effects in libs/game by default

jwunderl:
Previous "Particles" category now under advanced as "Effects"

## game over:

can configure to have different game over effects; defaults to confetti on win, and melt effect on lose

might be worth adding an 'in front' vs 'behind' option for the effects as well, so that the user can determine whether the effect gets rendered first or the score segment; this would just change the frame handler order in game over, so that you can make the confetti / etc show up on top of the score instead of behind

confetti

melt effect as default

## Sprite.destroy:

add optional parameter to do a simple destroy effect instead of just going off screen

## Sprite.startEffect:

the non advanced alias of a normal effect, that just adds the effect to the current sprite.

## image effects

Just adding in two for now: 

melt:

and dissolve:

minor changes outside this

* when someone requests background image, if one doesn't exist, fill the new image with the current color (so that it matches what the background looks like when it is called)
* default settings for all effects so pps can be omitted and effects still all look good

screen shot 2019-01-08 at 4 11 59 pm
screen shot 2019-01-08 at 4 11 39 pm
screen shot 2019-01-08 at 4 11 21 pm
screen shot 2019-01-08 at 4 11 02 pm


pelikhan:
Why not just a method in the interface? as opposed to a field?
```
interface BackgroundEffect {
    startSceneEffect(): void;
}
```

pelikhan:
Looks good, detroy missing effect suffix.

libs/game/background.ts
```diff
@@ -32,8 +32,10 @@ namespace scene {
         }
 
         get image() {
-            if (!this._image)
+            if (!this._image) {
                 this._image = image.create(screen.width, screen.height);
+                this._image.fill(this.color);
```
pelikhan: shouldn't you fill with transparent?

libs/game/game.ts
```diff
@@ -116,31 +129,41 @@ namespace game {
     //% weight=80 help=game/over
     export function over(win: boolean = false) {
         init();
-        if (__isOver) return
+        if (__isOver) return;
         __isOver = true;
-        // clear all handlers
-        control.pushEventContext();
```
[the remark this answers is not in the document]
jwunderl: occurs in game.init, when the new scene is pushed

libs/game/game.ts
```diff
@@ -116,31 +129,41 @@ namespace game {
...
-        // register system menu again
-        scene.systemMenu.register();
+        let chosenEffect = win ? winEffect : loseEffect;
+        
+        let background = screen.clone();
         // one last screenshot
         takeScreenshot();
+        
+        // clear all handlers
+        _scene = undefined;
```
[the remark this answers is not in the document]
jwunderl: good catch, I missed that I needed to pop here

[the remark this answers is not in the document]
jwunderl: Actually that brings up a good question, it calls ``control.reset()`` later in the function which will reset everything even if there are other scenes. Should that be changed to not immediately end everything after a game over (maybe pop scene or something of the sort)?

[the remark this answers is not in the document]
riknoll: Maybe reset if at the bottom of the stack and pop if it isn't?

pelikhan:
Control.reset does a full hardware reset.

libs/game/game.ts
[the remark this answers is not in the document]
pelikhan: reset will restart the game without going through the game over sequence.

libs/game/game.ts
```diff
@@ -163,6 +155,7 @@ namespace game {
 
             pause(2000); // wait for users to stop pressing keys
             waitAnyButton();
+            popScene();
```
pelikhan: not needed, control.reset restarts the device.

jwunderl:
Wow, I didn't realize fixedInstances worked for interfaces too, that makes ``setGameOverEffect`` just work and display ``ImageEffects``, ``ParticleEffects``, and anything else that would be valid in the future. Nice.

Yes, it does the hardware reset; Richard and I were wondering if that would be expected behavior within a game over, where if it just pushed when there are other scenes it could be used with things like mini game collections ('game over' for this minigame, but continue to other mini games)

pelikhan:
You would need to add a hook to override the behavior. We can cross that bridge when we get there. Reset also cleans up all memory leaks etc...

libs/game/effects.ts
```diff
@@ -57,6 +57,9 @@ namespace effects {
         //% group="Images"
         startSceneEffect(times?: number, delay?: number): void {
             times = times ? times : 15;
+
+            // fill in colors so that the effects apply to entire background
+            scene.backgroundImage().replace(0, scene.backgroundColor());
```
pelikhan: This always creates an backagroubd image. What happens when there isn’t an image?

jwunderl: ``scene.backgroundImage()`` calls ``scene.Background.image`` which will create an image if one doesn't exist; without this https://github.com/Microsoft/pxt-common-packages/blob/387a22c48292cb5750d980f019bff7dd97d63c09/libs/game/effects.ts#L65

would still create one, but not include the background color when modifying the image. I can add a way to check if is currently a background image in `scene.Background` and just not do anything if there isn't one, but that seems confusing.

related, https://github.com/Microsoft/pxt-common-packages/blob/387a22c48292cb5750d980f019bff7dd97d63c09/libs/game/background.ts#L48 should probably be changed to `scene.drawTransparentImage` to let background color appear through clear spots in the background image, correct?

pelikhan: yes it should use drawTransparentImage.

pelikhan:
Tested on HW?

jwunderl:
Need to update my hardware first, one sec

jwunderl:
Couple things, game over sound sounds a bit weird when ``effects.dissolve`` is chosen as the game over effect (laggy), and too many particles could cause a crash when push scene was called.

Added some clean up of particles and all before pushing the scene to try to minimize memory usage, but there are still a few 0x21s occurring inconsistently so I'll investigate that for a few

Don't know about the sound though, now that I'm looking for it I can't even quite tell where it is defined - it's assigned undefined to start and doesn't

libs/game/game.ts
```diff
@@ -116,31 +124,38 @@ namespace game {
...
-        control.pushEventContext();
-        // register system menu again
-        scene.systemMenu.register();
+        let chosenEffect = win ? winEffect : loseEffect;
+        
+        let background = screen.clone();
         // one last screenshot
         takeScreenshot();
+        
+        pushScene();
```
pelikhan: you should actually pop all scenes here to release memory.

pelikhan:
You should pop all scenes before starting the game over so that you release allocated memory/sprites/etc...

libs/game/effects.ts
```diff
@@ -56,22 +58,24 @@ namespace effects {
         //% blockId=particleEffectStartScene block="apply %effect effect to background || %times times"
         //% group="Images"
         startSceneEffect(times?: number, delay?: number): void {
```
pelikhan: what happens when you call this from "update" ?

libs/game/game.ts
```diff
@@ -131,32 +134,33 @@ namespace game {
         let background = screen.clone();
         // one last screenshot
         takeScreenshot();
-        
+
```
pelikhan: add comment about releasing memory

mmoskal: all memory is only released after one additional frame - before that there might be references to the scene from running fibers. Though I guess the main memory usage would be background images, so as long as we clear these (which I think we do in pop scene) we should be fine.

jwunderl: Found the issue that was causing crashes, the ``let background = screen.clone()`` just above this snippet seems to be taking up just a tad bit too much memory. I've got a version that works around it in a way that works fine on hardware, but it's hacky / liable to break. See https://github.com/Microsoft/pxt-common-packages/blob/3ebc28fb543fabd06f35e9f1fbf64c9960f2c4fe/libs/game/game.ts#L136-L144

Any ideas on how to avoid doing something like this? If I don't clone it, the screen is cleared after the scene goes away, and the background is just a blank screen

mmoskal: I think the problem here is that the fibers won't stop until game.over exits, and the fiber keep references to scenes alive, even if you pop them.

I think we should clear some stuff in scene and event context when it is popped. In particular, try to clear the event handlers, sprite lists, background image. @riknoll ?

I guess I could add a runtime function that would set all object fields to `undefined` (`memzero` essentially), but for now we can just add a method.

pelikhan: @jwunderl try Michal suggestion, add a destroy/clean function in scene that set all fields to undefined. It should be called in popScene

libs/game/game.ts
```diff
@@ -58,6 +58,7 @@ namespace game {
     }
 
     export function popScene() {
+        _scene._clearScene();
```
pelikhan: is _scene guaranteed to be defined?

libs/game/scene.ts
```diff
@@ -127,5 +127,23 @@ namespace scene {
             this.allSprites.push(sprite);
             sprite.id = this.spriteNextId++;
         }
+
+        _clearScene() {
```
pelikhan: just call it "destroy"

jwunderl:
I got up to 16 with this nightmare fuel game before it crashed on hw, so it seems pretty reasonable?

https://makecode.com/_hem56vLhW2ec

arcade-screenshot 12

libs/game/game.ts
jwunderl: Not if popScene was just called, but even if it was guaranteed I should have a check there

pelikhan:
last nit: "on sprite ... sounds" is the typical wording for an event.
image

jwunderl:
I haven't added default lifespans for the sources yet no, but that would take like 5-10 minutes if you want them

pelikhan:
expand it all at once since ms is missing on the first expand

pelikhan:
otherise LATM

jwunderl:
Hm. One more thing, putting the start effects in a tight loop / on game update can be a bit problematic



I can make `prune` `clear` and `destroy` old effects to keep it to some constant max effect count to mitigate this?

pelikhan:
Yes please. The test case should be to turn on many effect in update and we survive.

jwunderl:
I moved the ``scene.destroy`` out of the pop for now to fix the menu. I think that's reasonable enough, as it seems a bit aggressive to just remove everything from the scene (particularly because scenes are exposed through ``game.currentScene()``, so ``popScene`` could remove values the user is trying to keep track of).

As is, with the last change, it appears to work in all cases I have so far. Lowered the max effects constant slightly for hardware, but it still works perfectly fine even if you throw effects in an on game update (though it might just look like nothing is happening, as the sources will be constantly created and destroyed within a few ms)  Maybe a setting per hardware for this would be good

arcade-screenshot 13

pelikhan:
We can do per-hardware settings later. Tested on HW?

pelikhan:
LGTM

jwunderl:
One of the changes seems to have caused some weird behavior with `endSceneEffect`; on some occasions it results in `Program Error: Cannot read property 'head' of undefined`, which is odd as head is protected / only referred to from the source itself as `this.head`.

PR #626, opened 2019-01-09
support for TCC timer

pelikhan:
Setting up system timer as in https://github.com/lancaster-university/codal-circuit-playground/blob/samd-peripherals/model/CircuitPlayground.cpp#L48

@jamesadevine


mmoskal:
I'm moving devTimer ctor to platform.cpp

PR #631, opened 2019-01-11
basic screenshaking

pelikhan:
Classic screen shaking impl -- looks better in game than on gif
* the shacking is added on top of the camera offset so that it does not result in permanent effect.

shake

shake

image

### DIFF
--- a/libs/game/background.ts
+++ b/libs/game/background.ts
@@ -46,16 +46,16 @@ namespace scene {
             return !!this._image;
         }
 
-        render() {
+        draw() {
             screen.fill(this.color);
             if (this._image)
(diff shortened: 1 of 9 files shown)


ChaseMor:
Looks great! I really like it! How does it affect projectiles or sprites with the auto-destroy flag? Should a really shaky screen be able to destroy sprites if it pushed them out of the way of the camera?

pelikhan:
Good point about the physics. I now expose a "drawOffsetX/Y" vs "offsetX/Y". drawOffset is meant for drawing operation, and includes shake translations, while offsetX/Y is the unmodified camera position for physics/collision computation.
shake

PR #632, opened 2019-01-11
scene time tracking

pelikhan:
The scene keeps track of time spent in the update loop and countdown use this timer instead of the system timer. As a result, when a menu is displayed, the game timer is not incremented.
Fix for https://github.com/Microsoft/pxt-arcade/issues/441

### DIFF
--- a/libs/base/eventcontext.ts
+++ b/libs/base/eventcontext.ts
@@ -53,8 +53,8 @@ namespace control {
         private frameWorker: number;
         private framesInSample: number;
         private timeInSample: number;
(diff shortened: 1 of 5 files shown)


pelikhan:
@jwunderl

jwunderl:
nice!

also, in case anyone is having issues you will need to update arcade after pulling this in to remove the old `game.runtime` implementation https://github.com/Microsoft/pxt-arcade/commit/1f88ef351b7549f20adf3c819ce5653b82caab2c

PR #634, opened 2019-01-11
trimming Particle effects blocks

jwunderl:
closes Microsoft/pxt-arcade#574 (see discussion there)

arcade-screenshot

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -111,8 +111,6 @@ namespace game {
      * @param win whether the animation should run on a win (true)
      * @param effect
      */
-    //% group="Gameplay"
-    //% blockId=setGameOverEffect block="set game over effect for win %win=toggleYesNo to %effect"
     export function setGameOverEffect(win: boolean, effect: effects.BackgroundEffect) {
         init();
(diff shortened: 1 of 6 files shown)


jwunderl:
@abchatra wanted it customized as an optional parameter to ``game.over``, not as a separate block

jwunderl:
here's what it'll look like with the win/lose field editor

pelikhan:
Looks great!

libs/game/game.ts
```diff
@@ -111,8 +111,6 @@ namespace game {
      * @param win whether the animation should run on a win (true)
      * @param effect
      */
-    //% group="Gameplay"
-    //% blockId=setGameOverEffect block="set game over effect for win %win=toggleYesNo to %effect"
     export function setGameOverEffect(win: boolean, effect: effects.BackgroundEffect) {
```
pelikhan: We automatically do gameover when life == 0, so we need this block to configure the effect.
https://github.com/Microsoft/pxt-common-packages/pull/634/files#diff-7c3ef1deff0b4e062963ed4e50585d99L116

PR #639, opened 2019-01-15
Fixing typescript errors

abchatra:
Errors in the built-ins


pelikhan:
Really need to turn on that flag to find those bugs early

PR #640, opened 2019-01-15
adding butotn tester in menu

pelikhan:
Can be improved later. Basic button tester.


mmoskal:
I feel this might be too much - an example "game" I think should be enough. Not sure how often this would be used.

PR #646, opened 2019-01-16
Use TC4 not TCC0 on SAMD21

mmoskal:
I guess we should check if this fixes I2C?

### DIFF
--- a/libs/core---samd/platform.cpp
+++ b/libs/core---samd/platform.cpp
@@ -6,7 +6,7 @@
 namespace pxt {
 
 #ifdef SAMD21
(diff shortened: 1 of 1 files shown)


pelikhan:
Accelerometer on CPX?

mmoskal:
I mean, I doubt it fixes anything... we just don't need TCC so we can use TC

sabas1080:
Fix I2C??

pelikhan:
All changes are in maker beta, including your touch changes

sabas1080:
ok, reported error in discord with touch, i am going to test I2C

PR #653, opened 2019-01-17
Particle 980 fix

pelikhan:
Slice sources before pruning since pruning modifies the array.
Requires https://github.com/Microsoft/pxt/pull/5198


pelikhan:
@jwunderl

jwunderl:
Sorry for missing this one, looks good

PR #656, opened 2019-01-18
star field scene effect

jwunderl:


pelikhan:
Could we have a guide on how to author both sprite effects and screen effects? This is kind of a game in the game -- creating the best effect.

jwunderl:
and here's it with your multiplayer asteroids



I should probably mark scene effects as important or something of the sort, and put more thought into which source gets destroyed when too many are added rather than just the oldest

and I'll write a guide, just have to make time for it, need to finish the rest of the javascript course first.

pelikhan:
Wow!

PR #662, opened 2019-01-21
Allow dynamic selection of accelerometer used

mmoskal:
This allows several accelerometer drivers to be compiled in and one to be chosen at runtime using `ACCELEROMETER_TYPE` configuration variable (set in bootloader already).

By default only LIS3DH is compiled in, but this is overridden in Arcade.

Also, add function to use accelerometer as source of randomness.

### DIFF
--- a/libs/accelerometer/accelhw.cpp
+++ b/libs/accelerometer/accelhw.cpp
@@ -1,0 +1,119 @@
+// This supports a few different accelerometers.
+// If desired, overrides PXT_SUPPORT_* in platform.h (note that only LIS3DH is on by default).
+// Then accelerometer can be changed with config.ACCELEROMETER_TYPE in TypeScript.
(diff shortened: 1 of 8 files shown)


libs/accelerometer/accelhw.cpp
```diff
@@ -0,0 +1,119 @@
...
+// This file can be overridden alltogether by a target if a different accelerometer is desired.
+
+#include "pxt.h"
+#include "axis.h"
+#include "Pin.h"
+#include "I2C.h"
+#include "CoordinateSystem.h"
+
+#ifndef PXT_DEFUALT_ACCELEROMETER
+#define PXT_DEFUALT_ACCELEROMETER ACCELEROMETER_TYPE_LIS3DH
```
pelikhan: Default

PR #666, opened 2019-01-23
pause on game.ask so it isn't missed

jwunderl:
It's a bit too easy to skip through the ``game.ask`` dialogue when you're playing the game (/ mashing A or B) and completely miss it, so this adds a short pause between the prompt appearing and the player being able to make a decision, similar to the behavior in game over. https://github.com/Microsoft/pxt-common-packages/blob/1735068d37288dae04eeb0de30f86ff0a0c8b275/libs/game/game.ts#L190-L192

The 500ms pause is pretty arbitrary, but it's enough that the prompt will at least give them a chance to respond; 1000ms feels alright too, but 2000ms feels sluggish to me

### Current

(notice that the first few tries where I'm pressing a to fire the laser cause the prompt to never appear, with the game just appearing to be laggy for a second then the score going down)

### DIFF
--- a/libs/game/ask.ts
+++ b/libs/game/ask.ts
@@ -13,11 +13,11 @@ namespace game {
         control.pushEventContext();
         game.showDialog(title, subtitle, "A = OK, B = CANCEL");
         let answer: boolean = null;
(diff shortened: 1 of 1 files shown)


pelikhan:
We should use the menuing system eventually since it has animations.

PR #670, opened 2019-01-25
javascript timers

pelikhan:
setInterval, setTimeout, setImmediate and the like. Makes it easier to port existing typescript libraries.


riknoll:
Yes! I've wanted this forever

PR #681, opened 2019-02-01
Spi device

pelikhan:
Allow instantiation of multiple SPI devices.


libs/base/configkeys.h
```diff
@@ -137,4 +137,11 @@
 #define CFG_PINS_PORT_SIZE 210
 #define CFG_BOOTLOADER_PROTECTION 211
 
+#define CFG_LORA_MISO 220
```
mmoskal: Better use `CFG_PIN_LORA_MISO`; given that `CFG_PIN_*` stuff is getting close to 100 (which is D0), let's move these to say 1001, 1002, ...

pelikhan: moved

pelikhan:
I’ll get to them too.

PR #688, opened 2019-02-02
Support for multiple serial comm

pelikhan:
- [x] support for multiple serial comms
- [x] migrate most of the code impl to typescript


libs/serial/serial.cpp
```diff
@@ -55,191 +55,163 @@ enum class Delimiters {
...
-        return;
-      service->serial.redirect(*tx, *rx);
-      setBaudRate(rate);
-    }
+/**
+* Opens a Serial communication driver
+*/
+//% parts=serial
+SerialDevice createSerial(DigitalInOutPin tx, DigitalInOutPin rx, int id) {
+  return new CodalSerialDeviceProxy(tx, rx, id);
```
mmoskal: Maybe we should lookup an existing proxy if any?

PR #691, opened 2019-02-04
handle missing lora

pelikhan:
Better error handling when pins are missing or LoRa is incorrect version (e.g. simulator case)


pelikhan:
@sabas1080

sabas1080:
Thanks!

pelikhan:
Live in beta.

PR #692, opened 2019-02-04
I2c device

pelikhan:
Last round of moving bus devices to TypeScript. i2c. 
note: still broken on samd21.

### DIFF
--- a/libs/core/i2c.cpp
+++ b/libs/core/i2c.cpp
@@ -18,14 +18,14 @@ public:
 
     }
   
-  bool matchPins(DevicePin* sda, DevicePin* sck) {
-      return this->sda == sda && this->sck == sck;
+  bool matchPins(DevicePin* sda, DevicePin* scl) {
+      return this->sda == sda && this->scl == scl;
   }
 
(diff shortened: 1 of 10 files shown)


libs/core/i2c.cpp
```diff
@@ -1,43 +1,105 @@
...
+int writeBuffer(I2C_ i2c, int address, Buffer buf, bool repeat = false)
+{
+  return i2c->writeBuffer(address, buf, repeat);
+}
+
+}
+
+namespace pins {
+
+static vector<I2C_> i2cs;
```
mmoskal: no STL please; use a linked list of I2C_ instances.

libs/core/spi.cpp
```diff
@@ -34,64 +46,72 @@ class CodalSPIProxy {
     }
 };
 
-typedef CodalSPIProxy* SPIDevice;
+static vector<SPI_> spis;
```
mmoskal: no stl please

pelikhan:
Will fix

mmoskal:
Thanks, otherwise looks great!

PR #698, opened 2019-02-11
Character LCD driver support

pelikhan:
basic LCD screen driver support + sim support.

image


pelikhan:
@ladyada

ladyada:
hiya - i've forgotten, how would this be tested?

pelikhan:
I will have in maker so you can test it with metro/mono LCD. It is using the pins from the guide.

pelikhan:
Tested on hardware.

ladyada:
ok let me know when its ready for me to test!

PR #704, opened 2019-02-13
Let user override controller velocity when there is no input

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/755

Right now if you use the `moveSprite()` API, you can't ever change the velocity directly because it gets immediately overwritten (the controller takes complete control of the velocity even when buttons aren't being pressed). This change makes it so that it only controls the velocity when there is actual input.

Here's an example:

image

In the current editor, nothing will happen when the apple and lemon overlap because the controller is controlling the lemon's velocity. With this change, you get this:

new_ctrl

The pink arrows are the button's I'm pressing in that gif.

Also, I made it so that pressing directional buttons in opposite directions now cancels out the movement. Before we arbitrarily gave precedence to the up and left buttons.

@jwunderl @ChaseMor FYI

### DIFF
--- a/libs/game/controller.ts
+++ b/libs/game/controller.ts
@@ -177,6 +177,7 @@ namespace controller {
         s: Sprite;
         vx: number;
         vy: number;
+        _inputLastFrame: boolean;
     }
 
     export function _moveSprites() {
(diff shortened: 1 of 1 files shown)


riknoll:
@jamesadevine

jwunderl:
I don't think anything in the course relies on it, but the comment for the method needs to change as this is a change in that behavior / it specifies that it will take over the vx and vy

```
         * Control a sprite using the direction buttons from the controller. Note that this
         * control will take over the vx and vy of the sprite and overwrite any changes
         * made unless a 0 is passed.
```

jamesadevine:
Hmmmm @riknoll, I'm not sure this fixes my original problem.

I want to be able to "jump" whilst pushing the directional buttons, I can't do this right now if I use the move sprite with buttons block as the velocity is overwritten.

riknoll:
@jamesadevine you can do that with the current APIs: https://makecode.com/_M7miU4YwDFh8

titimoby:
image

Is it the same issue in this small example?

PR #711, opened 2019-02-22
basic documentation for repeat properties

jwunderl:
quick documentation for `Button.repeatDelay` and `Button.repeatInterval`

I didn't link it in too many places as it's javascript only currently


pelikhan:
Add a configuration section in the controller page (at the bottom).

jwunderl:
Sounds good, just a list of links? I don't think the ``cards`` used for the other sections work with js only functions / properties

pelikhan:
Right, let’s do links for now.

PR #715, opened 2019-02-26
adding circle apis

pelikhan:
Circle drawing using Fx algos. (FillCircle could/should be speeded up in the future)

### DIFF
--- a/libs/screen/image.ts
+++ b/libs/screen/image.ts
@@ -135,7 +135,9 @@ namespace helpers {
         cy = cy | 0;
         r = r | 0;
         // short cuts
-        if (r == 0) {
+        if (r < 0) 
+            return;
+        else if (r == 0) {
             img.setPixel(cx, cy, col);
             return;
         } else if (r == 1) {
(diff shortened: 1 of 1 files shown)


libs/screen/image.ts
```diff
@@ -118,6 +130,121 @@ namespace helpers {
...
+        }
+
+        const fcx = Fx8(cx);
+        const fcy = Fx8(cy);
+        const fr = Fx8(r);
+        const fr2 = Fx.leftShift(fr, 1);
+
+        let x = Fx.sub(fr, Fx.oneFx8)
+        let y = Fx.zeroFx8;
+        let dx = Fx.oneFx8;
```
mmoskal: @pelikhan I just noticed oneFx8 is really 1/256 (i.e. 0.004) and twoFx8 is 2/256, which seems horribly wrong.

Is this what this code is expecting? Or does it just work 256 times slower than expected?

pelikhan: 256 slower

PR #724, opened 2019-03-05
add priority ranking for sources (don't destroy scene effects)

jwunderl:
Add a relative priority ranking, so that developers can choose which effects should be destroyed when removing effects is deemed necessary so that the game doesn't crash.

This is used to give scene effects a higher priority, so that they are not destroyed when too many smaller effects are triggered at once

arcade-screenshot 5

I didn't add in a dirty bit for sorting the array for now (e.g. scene's ``Flag.NeedsSorting``), as the array being sorted will never have more than ``MAX_SOURCES`` (7) elements for the time being. If / when a PID controller is implemented for controlling particle effects, probably would want to revisit that.

@pelikhan for handling scene effects turning off when the player causes too many effects to be created at once: here's space destroyer with firing changed to repeat and star field effect on

### DIFF
--- a/libs/game/particleeffects.ts
+++ b/libs/game/particleeffects.ts
@@ -108,6 +108,7 @@ namespace effects {
 
             this.endScreenEffect();
             this.source = this.sourceFactory(new SceneAnchor(), particlesPerSecond ? particlesPerSecond : this.sceneDefaultRate);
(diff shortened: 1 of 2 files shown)


pelikhan:
LGTM

PR #726, opened 2019-03-07
Update IR library for CircuitPlayground Express

jamesadevine:
Contains a final update to make the no-mbed version of CPX feature complete.

@pelikhan and I will look to integrate and test next week, if that's ok @abchatra ?


abchatra:
Sounds good

pelikhan:
If things look good in maker, I will produce a branch in adafruit that we can push as beta

PR #732, opened 2019-03-08
Do not maintain animations for destroyed sprites

jwunderl:
I was getting some pretty poor frame rates on hardware in a platformer I'm making as games went on even when the number of elements on screen was relatively static.

Looks like the primary cause was a memory leak in animations, as the updates were keeping old sprites around and trying to update them forever rather than ignoring them after they were destroyed. For example: https://makecode.com/_A6i4uucDtdpu

So this fixes the memory leak

### DIFF
--- a/libs/animation/animation.ts
+++ b/libs/animation/animation.ts
@@ -4,7 +4,7 @@
 //% color="#03AA74" weight=78 icon="\uf021"
 namespace animation {
     //Handles all the updates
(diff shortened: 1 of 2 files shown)


jwunderl:
Hm. That example doesn't use the animations extension, but I'll take a look through particles again when I get some extra time to see if I can find where it is losing memory - it seems like it would have to be a weird edge case or something for it to last so long then crash

PR #736, opened 2019-03-08
patching IR (tested on HW)

pelikhan:
Patch from @jamesadevine


pelikhan:
TC3 is used by the DAC as well. Looking into using a different channel.

mmoskal:
There are unused channels on the system low level timer. But it would require some api change to use that.

jamesadevine:
https://github.com/Microsoft/pxt-common-packages/pull/736/files#diff-afd3b1716803cee6eade9e78834d5de5R183

Will never be true as we have a 16 bit timer with 1 us precision, so we will wait variable times between sends.

jamesadevine:
Will use system_timer_* instead.

jamesadevine:
@mmoskal I'm using TCC2 for IR timings instead of TC3. I don't think this affects any user facing GPIO.

When it comes to integrating jacdac in the future, we won't be able to run TCC timer at the highest priority (if we run IR at all)

PR #746, opened 2019-03-13
Support for a vibrating motor in controller

pelikhan:
Support for an **optional** vibrating motor on a game controller (enabled if CFG_PIN_VIBRATION is defined).
- ignored in simulator. we could add a visual effect like shaking the sim.
- does not support native app scenarios, refactoring when that happens.


pelikhan:
Doing it differently.

PR #750, opened 2019-03-13
Controller acceleration

pelikhan:
Basic blocks to support an accelerometer in arcade. Note that the "onGesture" event runs the handler in the "update" frame handler.
image


libs/controller/accelerometer.ts
```diff
@@ -0,0 +1,101 @@
...
+     * x=0, y=0 and z=-1023)
+     * @param dimension the axis along which the acceleration if measured
+     */
+    //% blockId=ctrlaccelerationvalue block="acceleration (mg)|%NAME"
+    //% parts="accelerometer"
+    //% dimension.fieldEditor="gridpicker"
+    //% dimension.fieldOptions.width=180
+    //% dimension.fieldOptions.columns=2
+    //% group="Extras"
+    export function acceleration(dimension: ControllerDimension): number {
```
mmoskal: Can't we have controller.accelerationX, controller.accelerationY, ... and collapse them into one block like we do for sprite properties?

darzu:
What does "vibrate" do? How is it different than camera shake?

pelikhan:
Camera shake is a visual effect, vibrate turns on a vibrating motor.

libs/controller/accelerometer.ts
pelikhan: Only support on class getters/fields.

PR #754, opened 2019-03-16
Use image dropdown for particle effects

jwunderl:
There are a decent number of effects, so they don't all appear in the default drop down without scrolling:

Grid picker looks a bit nicer:


pelikhan:
We need imageeeeessssss!

jwunderl:
Any way to get gifs to work for the grid picker? Would work well with the gif generation to make a tiny example of each

Also for what it's worth I found out the other day I can't use ``pxt buildsprites``, I get what seems to be a native error:

```
Using target arcade with build engine codal
  target: v0.7.10 /Users/Joey/pxt-common/pxt-arcade
  pxt-core: v5.5.12 /Users/Joey/pxt-common/pxt-arcade/node_modules/pxt-core
node[15565]: ../src/node_zlib.cc:314:static void node::(anonymous namespace)::CompressionStream<node::(anonymous namespace)::ZlibContext>::Write(const FunctionCallbackInfo<v8::Value> &) [CompressionContext = node::(anonymous namespace)::ZlibContext, async = false]: Assertion `Buffer::HasInstance(args[4])' failed.
 1: 0x100062df2 node::Abort() [/usr/local/bin/node]
 2: 0x100062d24 node::PrintErrorString(char const*, ...) [/usr/local/bin/node]
 3: 0x1000dd7e9 void node::(anonymous namespace)::CompressionStream<node::(anonymous namespace)::ZlibContext>::Write<false>(v8::FunctionCallbackInfo<v8::Value> const&) [/usr/local/bin/node]
 4: 0x100238617 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo*) [/usr/local/bin/node]
 5: 0x100237be6 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/usr/local/bin/node]
 6: 0x1002372e0 v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) [/usr/local/bin/node]
 7: 0x47b83ecfc7d 
Abort trap: 6
```

Should have time to try and make some images this weekend though

pelikhan:
Images are good enough IMO.

jwunderl:
can someone add [do not merge] to this for now, I'm getting an issue with images not being loaded that's breaking the `image dropdown` [here](https://github.com/Microsoft/pxt/blob/718caa5069ae032faaa43c40e8f67c6990a79c10/pxtblocks/fields/field_imagedropdown.ts#L218).

jwunderl:
@pelikhan should be good to go besides the fixedInstance bug, although I think it might be nice to have the option to keep a label with the effects name eventually - I think both the name and image are useful for that

pelikhan:
> @pelikhan should be good to go besides the fixedInstance bug

What is that bug?

jwunderl:
In the gif in https://github.com/Microsoft/pxt-common-packages/pull/754#issuecomment-474059974, watch the hover over the screen effects, they don't follow the options listed so they're tiny and all in a small line

pelikhan:
Can we shrink the image size to the same size as neopixel animations?

jwunderl:
Fits nicely with 4 columns too

pelikhan:
Unfortunately it does not iconify very well. It feels like it mostly dark screenshots.

jwunderl:
Hm, I'll try zooming in a bit more, the one that looks best at a glance is the trail because its so zoomed in

pelikhan:
What is the image size? Try to match the size in adafruit as well. I would say that pixelates zoomed image would look better.

jwunderl:
The animations are rendered from svgs at 111x111px it looks like, which is approximately the same size as these (just zoomed out a bit too much). Will also try clearing out the background, will have to see if the effects still show up against the block colors though

jwunderl:
How does this one look for the hearts effect? 

Screen Shot 2019-03-18 at 9 05 39 PM

Screen Shot 2019-03-18 at 9 06 32 PM

pelikhan:
Better. The blue background and red foreground is a bad combo though.

jwunderl:
Yeah, not all are really suited for icons I suppose, the spray and blizzard both just look like random dots. Not sure 



arcade-screenshot (3)

jwunderl:
I think these are gonna need some svgs to look good, only a few of them actually look good without motion (maybe the stars, the radials, and the fire). Will revisit / open a different PR after doing that at some point

PR #762, opened 2019-03-20
Adding comment attributes to sprite getters we want to show in the debugger

gushuro:
Marking Sprite getters (x, vx, ax, y, vy, ay) to be called in the debugger.


riknoll:
It means that it's safe to invoke the getter in the debugger (no side effects)

pelikhan:
LGTM.

PR #760, opened 2019-03-20
refresh dal + fix serial event

pelikhan:
Wiring up the proper implementation for "data received" event in Serial.
@jamesadevine

PR #763, opened 2019-03-21
Serial ref update

ganicke:
Update and add new ref pages for serial lib.

RE: #760

### DIFF
--- a/libs/serial/docs/reference/serial/create-serial.md
+++ b/libs/serial/docs/reference/serial/create-serial.md
@@ -3,7 +3,7 @@
 Create a new serial communication channel on the pins.
 
 ```sig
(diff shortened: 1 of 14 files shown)


libs/serial/docs/reference/serial/create-serial.md
```diff
@@ -1,17 +1,27 @@
 # create Serial
 
-Opens a UART Serial communication channel over the given pins.
+Create a new serial communication channel on the pins.
 
 ```sig
-pins.createSerial(undefined, undefined, undefined);
+pins.createSerial(pins.A0, pins.A1, 0);
```
pelikhan: Use TX, RX

PR #764, opened 2019-03-21
Add a GC check

mmoskal:
I think it would be nice to have `PXT_GC_CHECKS` off by default and have some option somewhere to turn it on. It should definitely be off by default in Arcade. @pelikhan do we have a way of doing that?


pelikhan:
Yes we can add a toggle in project settings.

PR #772, opened 2019-03-26
Add the option to change separator in datalogger

ganicke:
Datalogger

Values of `0` are omitted from the text rows on output. Seems that ``array.join()`` treats values of `0` as `null` and drops them out of the result string. See below:

```typescript
let testRows: number[][] = []
testRows = [[23, 9, 56], [87, 0, 45], [75, 0, 0], [0, 32, 45]]
for (let row of testRows) {
	console.log(row.join(','))
}
```

result:

```
23,9,56
87,,45
75,,
,32,45
```

A suggested fix is to convert the row to string prior to formatting and output.

- [x] Convert output row to string values.
- [x] Add the option to change the separator character.

### DIFF
--- a/libs/datalogger/datalogger.ts
+++ b/libs/datalogger/datalogger.ts
@@ -1,3 +1,12 @@
+enum LogSeparator {
+    //% block="tab"
+    Tab,
+    //% block="comma"
+    Comma,
+    //% block="semicolon"
+    Semicolon
+};
+
 /**
  * A tiny data logging framework
  */
@@ -22,7 +31,7 @@ namespace datalogger {
         /**
          * Appends a row of data
          */
-        appendRow(values: number[]): void { }
+        appendRow(values: string[]): void { }
         /**
          * Flushes any buffered data
          */
@@ -80,10 +89,11 @@ namespace datalogger {
                     }
                 }
                 // append row
-                _storage.appendRow(_row);
+                let textRow: string[] = _row.map(x => "" + x);
+                _storage.appendRow(textRow);
                 if (_console) {
                     // drop time
-                    console.log(_row.slice(1, _row.length).join(','));
+                    console.log(textRow.slice(1, _row.length).join(','));
                 }
                 // clear values
                 _row = undefined;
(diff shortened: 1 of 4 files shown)


jwunderl:
Probably good to just fix https://github.com/Microsoft/pxt/blob/09bb294ec2712e5b24e18d38721ddd61fea84061/libs/pxt-common/pxt-helpers.ts#L74 to handle null and undefined explicitly to match ts semantics: http://www.typescriptlang.org/play/index.html#src=alert(%5B0%2C1%2C2%2C3%2C%20undefined%2C%20null%5D.join(%22%2C%22))

ganicke:
OK, I'll pull out the string conversion and move this to https://github.com/Microsoft/pxt/issues/5396.

jwunderl:
Actually, looks like that one can't be changed until https://github.com/Microsoft/pxt-arcade/issues/311 is fixed, as the nulls / undefined numbers (and booleans, oddly enough) are being represented as 0's; after a fix changing the line to ``r += (arr[i] === undefined || arr[i] === null) ? "" : arr[i];`` the result of 

```typescript
let myNumArr = [0, 1, 2, 3, undefined, null];
let myBoolArr = [true, false, undefined, null];

console.log("nums: " + myNumArr.join(","));
console.log("bools: " + myBoolArr.join(","));
```

is

```
nums: 0,1,2,3,0,0
bools: true,false,0,0
```

libs/datalogger/datalogger.ts
```diff
@@ -1,3 +1,12 @@
+enum LogSeparator {
+    //% block="tab"
+    Tab,
```
pelikhan: Do we support enums as strings or use the ASCII value and use String.fromCharCode

ganicke:
It seems that it's restricted to only numeric literals.

image

pelikhan:
Sorry about that!

ganicke:
First world coding problems.

libs/datalogger/docs/reference/datalogger/set-separator.md
```diff
@@ -0,0 +1,57 @@
...
+```blocks
+datalogger.setEnabled(false)
+datalogger.setSeparator(LogSeparator.Comma)
+datalogger.setEnabled(true)
+datalogger.sendToConsole(true)
+forever(function () {
+    datalogger.addValue("x", Math.randomRange(0, 10))
+    datalogger.addValue("y", Math.randomRange(0, 10))
+    datalogger.addValue("z", Math.randomRange(0, 10))
+    datalogger.addRow()
```
pelikhan: you should start with addRow but i guess it does not matter.

ganicke: What about an init row check in addValue() rather than bailing out? Then addRow() can either lead or trail addValue():

```typescript
export function addValue(name: string, value: number) {
        if (!_row) initRow();
        // happy path
```

PR #775, opened 2019-03-29
Using correct api to schedule a setTimeout in the runtime.

gushuro:
Part of the debugger work on fiber management.
Also added debug preview for images (unrelated but harmless).


pelikhan:
You might need to bump the pxt reference in this repo.

PR #778, opened 2019-04-02
add helpers to write to i2c registers

pelikhan:
Wrapper for common register interaction in i2c


libs/core/i2c.ts
```diff
@@ -16,12 +16,55 @@ namespace pins {
...
+    /**
+     * Writes a value in a I2C register.
+     * @param address I2c address of the device
+     * @param register register index
+     * @param value value to write
+     * @param registerFormat format of the register, default is UInt8BE
+     * @param valueFormat format of the value, default is UInt8LE
+     */
+    //% weight=3 group="i2c"
+    //% blockId=i2c_writereg block="i2c write register|at address $addfress|at register $register|value $value"
```
mmoskal: Typo? Addrfes?

libs/core/i2c.ts
```diff
@@ -16,12 +16,55 @@ namespace pins {
...
         pins.i2cWriteBuffer(address, buf, repeated)
     }
 
+    /**
+     * Writes a value in a I2C register.
+     * @param address I2c address of the device
+     * @param register register index
+     * @param value value to write
+     * @param registerFormat format of the register, default is UInt8BE
+     * @param valueFormat format of the value, default is UInt8LE
```
mmoskal: Le and be are the same for 8 bit. Just pick one

libs/core/i2c.ts
```diff
@@ -16,12 +16,55 @@ namespace pins {
...
+     */
+    //% weight=3 group="i2c"
+    //% blockId=i2c_writereg block="i2c write register|at address $addfress|at register $register|value $value"
+    export function i2cWriteRegister(address: number, register: number, value: number, registerFormat?: NumberFormat, valueFormat?: NumberFormat): number {
+        if (registerFormat === undefined)
```
mmoskal: I've only ever seen 8 bit registers. Maybe swap arguments so register format is last? Or nuke it?

PR #782, opened 2019-04-04
Macros for accessing buffer and string data across versions

jwunderl:
will fix https://github.com/Microsoft/pxt-microbit/issues/1941, still need to make a PR to the extension repo once this is merged though

moves the macros from microbit to be structures themselves


jwunderl:
@pelikhan I do still need to set up my local build properly to test this fully, I haven't been able to get docker to work yet. Any pointers / black magic needed to do that?

pelikhan:
bumping...

pelikhan:
You can skip docker by install yotta and other tools. Or just let the cloud do the build.

PR #784, opened 2019-04-05
toString -> toText, default to numbers

jwunderl:
See https://github.com/Microsoft/pxt-microbit/pull/1977

### DIFF
--- a/libs/base/docs/reference/text/to-text.md
+++ b/libs/base/docs/reference/text/to-text.md
@@ -1,7 +1,7 @@
-# to string
+# to text
 
 Converts a value into a text representation.
 
 ```sig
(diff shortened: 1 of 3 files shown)


libs/base/docs/reference/text/to-text.md
```diff
@@ -1,7 +1,7 @@
-# to string
+# to text
 
 Converts a value into a text representation.
 
 ```sig
-toString(123)
+toText(123)
```
pelikhan: to match the blocks, it should be "convertToText"

jwunderl: good call, switched

PR #786, opened 2019-04-10
Fix GC assertion

mmoskal:
fixes https://github.com/Microsoft/pxt-microbit/issues/1985

### DIFF
--- a/libs/base/gc.cpp
+++ b/libs/base/gc.cpp
@@ -192,7 +192,7 @@ static uint8_t *midPtr;
 
 static bool inGCArea(void *ptr) {
     for (auto block = firstBlock; block; block = block->next) {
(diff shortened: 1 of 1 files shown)


mmoskal:
inGC() is used in assertions to check that a given pointer was allocated by the GC. It wasn't working for the first pointer allocated in a GC block.

abchatra:
Shortest bug fix I have ever seen.

abchatra:
It is more interesting that repro of the bug is consistent.

mmoskal:
Things are very deterministic in these embedded systems - here there's some randomness from where you press the button and the light level, but I guess it doesn't influence anything.

mmoskal:
I wonder why we were not hitting that before, but I guess we don't trigger this check too often...

PR #787, opened 2019-04-11
Update 'convertToText' ref page

ganicke:
Fill in the 'convertToText' ref doc.


ganicke:
Just any type or any type supporting `toString()`?

jwunderl:
I think it should be anything, whether its necessarily useful is a different question https://makecode.microbit.org/_ggKLzh2LhExF

PR #788, opened 2019-04-14
Detect accelerometer type

STEMSmartLabs:


mmoskal:
This should be tested not to hang when there's no accelerometer connected.

STEMSmartLabs:
> This should be tested not to hang when there's no accelerometer connected.

I don't see any blocking call on I2C::read() driver for now.
https://github.com/lancaster-university/codal-core/blob/master/source/driver-models/I2C.cpp

However, I will verify in hardware without accelerometer connected.

STEMSmartLabs:
image

Tested with & without Accelerometer, Received the console output and accelerometer sensing.  Any suggestion ?

pelikhan:
Looks I will merge today.

PR #789, opened 2019-04-15
Always show win/lose toggle in game.over

jwunderl:
closes https://github.com/Microsoft/pxt-arcade/issues/872

Changed parameter name mainly to disambiguate upgraded / un-upgraded scripts for the upgrade rules.

### Old:

Screen Shot 2019-04-15 at 4 11 10 PM
Screen Shot 2019-04-15 at 4 10 56 PM

### New:

Screen Shot 2019-04-15 at 4 10 10 PM
Screen Shot 2019-04-15 at 4 09 55 PM

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -158,7 +158,7 @@ namespace game {
      * Finish the game and display the score
      */
     //% group="Gameplay"
-    //% blockId=gameOver block="game over %gameWon=toggleWinLose || with %effect effect"
+    //% blockId=gameOver block="game over %win=toggleWinLose || with %effect effect"
     //% weight=80 help=game/over
     export function over(win: boolean = false, effect?: effects.BackgroundEffect) {
         init();
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -158,7 +158,7 @@ namespace game {
      * Finish the game and display the score
      */
     //% group="Gameplay"
-    //% blockId=gameOver block="game over || %win=toggleWinLose with %effect effect"
+    //% blockId=gameOver block="game over %win=toggleWinLose with %effect effect"
     //% weight=80 help=game/over
     export function over(win: boolean = false, effect?: effects.BackgroundEffect) {
         init();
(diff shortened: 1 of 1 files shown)


riknoll:
I don't think we need to change the parameter name; we can detect whether it's old/new using the dom

libs/game/game.ts
```diff
@@ -158,7 +158,7 @@ namespace game {
      * Finish the game and display the score
      */
     //% group="Gameplay"
-    //% blockId=gameOver block="game over || %win=toggleWinLose with %effect effect"
+    //% blockId=gameOver block="game over %gameWon=toggleWinLose || with %effect effect"
```
pelikhan: can we keep %win ?

libs/game/game.ts
```diff
@@ -158,7 +158,7 @@ namespace game {
...
     //% weight=80 help=game/over
     export function over(win: boolean = false, effect?: effects.BackgroundEffect) {
```
pelikhan: what happens when we decompile ?
```
game.over()
```

jwunderl: It'll get a hole, same as if you decompile ``sprites.create(img`1`)``:

Screen Shot 2019-04-15 at 4 59 53 PM

libs/game/game.ts
jwunderl: Yup, undid that already, just need to fix upgrade rules to handle it appropriately

libs/game/game.ts
pelikhan: That's unfortunate, ``game.over()`` is perfectly valid.

riknoll: Maybe for primitive types the decompiler should put in the correct default? I think that's safe enough

riknoll: Well, no, never mind. Didn't think that through

pelikhan: Our api consistantely assume "false" as default so that should be good enough.

riknoll: We could at least detect primitives that have initializers in the function signatures. However, it might be a stretch to map it to the win/lose block since the XML for that might not be the same as a normal logic true/false

riknoll: https://github.com/Microsoft/pxt-arcade/issues/891

PR #790, opened 2019-04-16
Left/right support for volume brightness

pelikhan:
arcade-jdtest (1)

Otherwise, we have too many menu items and the jacdac debugger is pushed out.


pelikhan:
Eventually... for now it should be obvious that the screen brightness changes or sound level is played differently.

PR #792, opened 2019-04-18
Dynamic detection of acceleremeter

pelikhan:
1. try loading accelerometer type from CFG_ACCELEROMETER_TYPE
2. if fail, sniff i2c and try to instantiate again.

### DIFF
--- a/libs/accelerometer/accelhw.cpp
+++ b/libs/accelerometer/accelhw.cpp
@@ -11,7 +11,7 @@
 #include "CodalDmesg.h"
 
 #ifndef PXT_DEFAULT_ACCELEROMETER
-#define PXT_DEFAULT_ACCELEROMETER ACCELEROMETER_TYPE_LIS3DH
+#define PXT_DEFAULT_ACCELEROMETER -1
 #endif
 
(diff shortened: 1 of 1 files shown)


libs/accelerometer/accelhw.cpp
```diff
@@ -11,11 +11,11 @@
 #include "CodalDmesg.h"
 
 #ifndef PXT_DEFAULT_ACCELEROMETER
-#define PXT_DEFAULT_ACCELEROMETER ACCELEROMETER_TYPE_LIS3DH
+#define PXT_DEFAULT_ACCELEROMETER -1
 #endif
 
 #ifndef PXT_SUPPORT_LIS3DH
-#define PXT_SUPPORT_LIS3DH 1
+#define PXT_SUPPORT_LIS3DH 0
```
mmoskal: Will need to define PXT_SUPPORT_LIS3DH to 1 in pxt-adafruit

pelikhan: put it back... and to be sure: https://github.com/Microsoft/pxt-adafruit/pull/999

PR #793, opened 2019-04-18
game.forever events that are attached to scene

jwunderl:
fixes https://github.com/Microsoft/pxt-arcade/issues/610

Adds ``game.forever`` to replace forever in arcade. This has the events attached to the current scene, so that they will not continue indefinitely after the scene changes (e.g. after the game is over)

Also splits out ``libs/base/control.ts`` into a few different files, so that forever can be deprecated in arcade. Only changes (not moves) should be seen in ``libs/game/scene.ts`` and ``libs/game/game.ts``

Will need a few docs / upgrade rules / etc in arcade too.

arcade-screenshot (4)


libs/game/scene.ts
```diff
@@ -23,6 +23,11 @@ namespace scene {
         handler: (sprite: Sprite) => void
     }
 
+    export interface GameForeverHandlers {
+        active: boolean;
```
riknoll: Nitpick: Maybe "lock" or "running" instead of active? Active makes it sound like it has been enabled/disabled or something.

jwunderl:
Went through targets to verify splitting control.ts into separate files wouldn't break them; only one that appeared to both use common packages and override control.ts seems to be microbit, and that works fine / doesn't change as https://github.com/Microsoft/pxt-microbit/blob/master/libs/core/pxt.json is non-empty and is not replaced with the common packages version.

PR #794, opened 2019-04-18
block to set defaults for repeat events

jwunderl:
fixes https://github.com/Microsoft/pxt-arcade/issues/767

Block sets default for **all** buttons; setting them on a per - button basis is advanced / typescript only.

Also fixes a bug where the time before the first repeat was the repeatDelay plus the repeatInterval, rather than just the repeatDelay

arcade-screenshot (5)

Screen Shot 2019-04-18 at 4 25 03 PM

### DIFF
--- a/libs/game/controller.ts
+++ b/libs/game/controller.ts
@@ -38,6 +38,8 @@ enum ControllerEvent {
 //% blockGap=8
 namespace controller {
     let _userEventsEnabled = true;
+    let defaultRepeatDelay = 500;
+    let defaultRepeatInterval = 30;
 
     //% fixedInstances
     export class Button {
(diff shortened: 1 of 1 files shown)


libs/game/controller.ts
```diff
@@ -142,19 +144,35 @@ namespace controller {
...
             if (count != this._repeatCount) {
                 this.raiseButtonRepeat();
                 this._repeatCount = count;
             }
         }
     }
 
+    /**
+     * Set default values for repeat button event delay and interval
+     * @param delay default delay for buttons, eg: 500
```
riknoll: Maybe "number of milliseconds from when the button is pressed to when the repeat event starts firing"

libs/game/controller.ts
```diff
@@ -142,19 +144,35 @@ namespace controller {
...
+     * @param interval default interval for buttons, eg: 30
```
riknoll: Maybe "minimum number of milliseconds between calls to the button repeat event"

libs/game/controller.ts
```diff
@@ -142,19 +144,35 @@ namespace controller {
...
+            const count = Math.floor((this._pressedElasped - delay - interval) / interval);
             if (count != this._repeatCount) {
                 this.raiseButtonRepeat();
                 this._repeatCount = count;
             }
         }
     }
 
+    /**
+     * Set default values for repeat button event delay and interval
```
riknoll: Configures the timing of the `on button repeat` event for all of the controller buttons

riknoll:
I'm a little worried that the wording of the block makes it sound like it might affect the repeat loop. Should we add the word button somewhere?

jwunderl:
Swtched to ``set button repeat delay $delay interval $interval``, as that order matches the order it appears in the event

PR #799, opened 2019-04-24
Move remove game.forever block (set block in target instead)

jwunderl:
Moving this block to be defined as just forever in arcade


pelikhan:
Do we still have a forevor block under loops?

jwunderl:
with https://github.com/Microsoft/pxt-arcade/pull/921 it will

PR #801, opened 2019-04-25
move mixer/synth from codal to PXT

mmoskal:
I'm not sure we want to merge it just yet.

This fixes stuttering sounds in arcade. It also brings back ringTone, but it clicks every half second.

Tested on Arcade STM only so far. You can try to test it on Adafruit and possibly include there.


mmoskal:
There are memory problems here. Will push a fix today

pelikhan:
Ready to go? @mmoskal

PR #805, opened 2019-04-29
Support analog joystick and active high buttons

mmoskal:
also fixes for button multiplexer

@ladyada - it's coming!


ladyada:
:D :D :D do you need the pygamer schematic? I AM SO EXCITED!

pelikhan:
The changes are live in https://arcade.makecode.com/beta . You’ll need to update the bootloader to specify that buttons are analog.

mmoskal:
The settings are something like

Btn_left = 1100 + pb13
Btn_right = 1200 + pb13

ladyada:
aah ok so here's the current (wrong) def
https://github.com/adafruit/uf2-samdx1/blob/master/boards/arcade_pygamer/board_config.h#L68

image

So 
```
47, 0x1127, // PIN_BTN_LEFT = analog PB07
48, 0x1227, // PIN_BTN_LEFT = analog PB07
```
?

mmoskal:
No its 1100 decimal sorry. So 1100 +0x27

ladyada:
ok then 0x473 and 0x4D7  - is there a convention for converting PORTxN to a hex number? it sems to be 0x0N and 0x1N for porta and 0x2N and 0x3N for portb but just checking

mmoskal:
Right. On ATSAMD ports are 32 pins so it's like you describe. On stm32 ports are 16 pins so it's different. You can use https://microsoft.github.io/uf2/patcher/ to update the.h file - it resolves pin names and sorts config keys appropriately

ladyada:
!@ amazing!

ladyada:
no luck - arcade game 'black screen' - i changed back to buttons and they run find (but of course i cant move the player) so its just the 4 line change here:
https://github.com/ladyada/uf2-samd21/commit/565042c5132e73c065ef6482f0d73e72b83163c0
bootloader works great otherwise, but beta.arcade is unable to load. lemme know if you want me to jlink it

mmoskal:
the ADC is busted; I did some fixes, but it still doesn't work; will look into this more

ladyada:
np its probably the clock ... ;)

ill get y'all hardware very soon, putting em together now

mmoskal:
I have a test build here with fixed ADC and sound: 

https://arcade.makecode.com/app/0d0584e40b68f43b1ed93ea4a744110c76767ffa-e1f3c3de51

The ADC should work with the new button configs. If it doesn't try this program and let me know what values you get:

```typescript
game.onPaint(function () {
    screen.print("PB06: " + pins.PB06.analogRead(), 60, 2)
    screen.print("PB07: " + pins.PB07.analogRead(), 60, 20)
})
```

Error will show before you switch to D5 board; ignore it - the private build has the pins PB06 and PB07 added.

As an added bonus the SAMD51 sound should also work on this one.

You can also tweak bootloader settings from TS (they override whatever is in bootloader) by adding this on top of `main.ts`:

```typescript
namespace userconfig {
    export const PIN_BTN_LEFT = 0x473
    export const PIN_BTN_RIGHT = 0x4D7
    export const PIN_BTN_UP = 0x472
    export const PIN_BTN_DOWN = 0x4D6
}
```

ladyada:
joystick/analog read works!

sound does not full - it does for a moment on the first sound and then goes silent
also, we use a speaker enable pin for our amps, that should be set HIGH when making sounds, and you can set back to LOW when done (or keep HIGH all the time, but default is pulled low!)

mmoskal:
Awesome! Can you share your sound test program? It was working for me on pybadge.

ladyada:
here's a noisy falling duck
https://makecode.com/_VHweioUCMdzq
how do you set the speaker enable?

mmoskal:
Amp is here https://github.com/Microsoft/pxt-common-packages/blob/master/libs/mixer/melody.cpp#L58 and https://github.com/Microsoft/pxt-common-packages/blob/master/libs/mixer/melody.ts#L49 so it should just stay on.

Checking on your sample.

ladyada:
i dont see speaker_amp in the board_config.h in my boards. whats the 'pin #'?

mmoskal:
right; I'm not sure why does it work at all then!

```cpp
#define CFG_PIN_SPEAKER_AMP 26
```

ladyada:
maybe your old version had the amp always on :D ok trying with a new bootloader...

ladyada:
ok i definitely get the 'beginnings' of sound but only for a split second, then nothing on the scope for A0. so audio amp enable is now good, but audio is still :(

im using this bootloader
https://github.com/ladyada/uf2-samd21/tree/master/boards/arcade_pybadge

mmoskal:
do `music.setVolume(50)` at the start of the program; or set

```cpp
#define CFG_SPEAKER_VOLUME 59
```

to 50.

I think the default is 128 and it's somehow too much. Will investigate.

ladyada:
ok trying `#define CFG_SPEAKER_VOLUME 59`

ladyada:
ok now no soudn at all, you didnt mean that i should change
`    59, 0x200, // SPEAKER_VOLUME = 512` instead of that define?

mmoskal:
right, replace `59, 0x200` with `59, 50`

ladyada:
sorry, i took you a little too literally - trying now :)

ladyada:
yah works now! i see the issue, the DC offset is 500mV, it should be closer to 1.5V - easy fix but for now i can tweak this value till its as loud as it gets and commit

ladyada:
ok ill open another issue for volume/stereo

PR #808, opened 2019-05-01
Allow for GC out-of-order blocks

mmoskal:
Fixes https://github.com/Microsoft/pxt-microbit/issues/2032


abchatra:
Should this be ported to adafruit? How bad is this bug?

pelikhan:
We have some i2c fixes for CPX coming up too.

mmoskal:
Probably not needed on CPX if we are not hitting it yet. It will always clearly manifest as 840 no silent corruption

PR #809, opened 2019-05-01
allow for setting CPU clock speed on STM32

mmoskal:


mmoskal:
Rolling into https://github.com/microsoft/pxt-common-packages/pull/833 as this mostly makes sense on F412

PR #811, opened 2019-05-02
New menu design

riknoll:
pause_menu_new

I'm still getting out of memory errors sometimes, I have no idea what's causing them. I think it might be related to the push/pop scene because its the same behavior as the old menu.


pelikhan:
We need to make sure to avoid double double buffering

pelikhan:
@riknoll ready to go?

riknoll:
Yep! I'm not cloning the screen in this one, so I don't think that's it.

PR #814, opened 2019-05-06
JACCDAC config keys

pelikhan:
Misc updates from fashion branch: 
- [x] buffer getUint8/setUint8
- [x] console.warning -> console.warn
- [x] pin config key mask
- [x] fix onIdle

### DIFF
--- a/libs/base/shims.d.ts
+++ b/libs/base/shims.d.ts
@@ -2,8 +2,20 @@
 
 
 
-//% indexerGet=BufferMethods::getByte indexerSet=BufferMethods::setByte
+    //% indexerGet=BufferMethods::getByte indexerSet=BufferMethods::setByte
 declare interface Buffer {
+    /**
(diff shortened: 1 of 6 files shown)


libs/base/shims.d.ts
```diff
@@ -2,8 +2,20 @@
...
+     * Reads an unsigned byte at a particular location
+     */
+    //% shim=BufferMethods::getUint8
+    getUint8(off: int32): int32;
+
+    /**
+     * Writes an unsigned byte at a particular location
+     */
+    //% shim=BufferMethods::setUint8
+    setUint8(off: int32, v: int32): void;
```
mmoskal: Why do we need these? Indexers should just work and otherwise getNumber

pelikhan: The indexers don't work for jacdac-ts when we cross-compile with regular typescript.

PR #816, opened 2019-05-08
Add control.ramSize()

mmoskal:
Completely untested.

See https://github.com/microsoft/pxt-arcade/pull/958

CC @jwunderl

### DIFF
--- a/libs/base/control.ts
+++ b/libs/base/control.ts
@@ -252,6 +252,16 @@ namespace control {
             }
         }
     }
+
+    //% shim=control::_ramSize
+    function _ramSize() {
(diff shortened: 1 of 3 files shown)


jwunderl:
Hm, I'm getting ``pxt_modules/base/control.ts(263,35): error TS2339: Property 'CFG_RAM_BYTES' does not exist on type 'typeof DAL'.``, so maybe DAL is out of date? Removing the getConfigValue call works nicely for differentiating between pybadge and simulator, though

mmoskal:
Dal DTS has to be re generated in Pxt-arcade

PR #821, opened 2019-05-10
don't use Easing type

pelikhan:
It messes up Python
@darzu you should be able to recreate a repro out of this. This is different from the unexported enum.


darzu:
Will do

PR #824, opened 2019-05-11
Don't let user turn off screen brightness fully from menu

jwunderl:
Set minimum brightness to 10 when changing by menu, so that the game doesn't become unusable without a reset


pelikhan:
Can we fix setBrightness to never allow 0?

pelikhan:
And enforce the bounds

PR #825, opened 2019-05-11
Always enforce bounds on brightness

jwunderl:
re: Peli's comment on https://github.com/microsoft/pxt-common-packages/pull

### DIFF
--- a/libs/screen/image.ts
+++ b/libs/screen/image.ts
@@ -273,10 +273,9 @@ namespace helpers {
     function _setScreenBrightness(brightness: number) { }
 
     export function setScreenBrightness(img: Image, b: number) {
-        if (b >= 10 && b <= 100) {
-            _helpers_workaround.brightness = b
-            _setScreenBrightness(_helpers_workaround.brightness)
-        }
+        b = Math.clamp(10, 100, b);
+        _helpers_workaround.brightness = b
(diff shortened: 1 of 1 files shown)


libs/screen/image.ts
```diff
@@ -273,8 +273,10 @@ namespace helpers {
     function _setScreenBrightness(brightness: number) { }
 
     export function setScreenBrightness(img: Image, b: number) {
-        _helpers_workaround.brightness = b
-        _setScreenBrightness(_helpers_workaround.brightness)
+        if (b >= 10 && b <= 100) {
```
pelikhan: Can you do clamp(10,100,...) instead?

pelikhan:
Maybe also add | 0 for sanity

libs/screen/image.ts
jwunderl: I think I'll go through the math functions in a few and rename the parameters in a bit so they're nice and useable, as it's annoying to remember the order for clamp (I've seen helpers elsewhere with it set as ``clamp(min, num, max)`` as well)

pelikhan:
Any change of variable order would be a breaking change...

jwunderl:
Not change the order, just ``min, max, input`` instead of ``a, b, c`` so the tooltips tell you what to do

PR #827, opened 2019-05-12
Moving the circle filling into the cpp

riknoll:
The fixed point implementation in the TS is very slow on the hardware and the simulator. This speeds things up by moving it into the cpp like the other drawing "primitives".

Here's a demo program in the simulator that draws dozens of circles each frame (it's much smoother in non-gif form):

circle-perf-sim

It also runs at about 40 fps on the hardware, but I can't seem to make a GIF of it so you'll have to take my word for it.

Here's the demo code: https://makecode.com/_PTeCCUCt1C72 (compiles on beta+release but too slow to run)


riknoll:
I'm not sure why the TS was fixed point, it was all integer math. I first tried the integer math version in the TS, but performance was slightly better in the CPP. 

Since drawing a ton of circles is pretty reasonable for particles I think the CPP makes sense. Not much code to replicate in the sim either.

mmoskal:
Cool!

PR #829, opened 2019-05-15
Fix ANY processing

tballmsft:
This code has been around for a long time, and it is wrong. The previous lookup will find matches with DEVICE_EXT_ANY

### DIFF
--- a/libs/base/core.cpp
+++ b/libs/base/core.cpp
@@ -23,8 +23,9 @@ static HandlerBinding *handlerBindings;
 
 HandlerBinding *nextBinding(HandlerBinding *curr, int source, int value) {
     for (auto p = curr; p; p = p->next) {
-        if ((p->source == source || p->source == DEVICE_ID_ANY) && 
-            (p->value == value || p->value == DEVICE_EVT_ANY)) {
+        // DEVICE_ID_ANY == DEVICE_EXT_ANY == 0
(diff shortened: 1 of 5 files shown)


jwunderl:
This is the one that fixes https://github.com/microsoft/pxt-arcade/issues/959, correct?

tballmsft:
Yes, but I am not done yet...

libs/base/core.cpp
```diff
@@ -21,23 +21,25 @@ PXT_DEF_STRING(emptyString, "")
...
 
-HandlerBinding *findBinding(int source, int value) {
-    return nextBinding(handlerBindings);
-}
-
 HandlerBinding *nextBinding(HandlerBinding *curr, int source, int value) {
     for (auto p = curr; p; p = p->next) {
-        if ((p->source == source || p->source == DEVICE_ID_ANY) && (p->value == value || p->value == DEVICE_EVT_ANY) {
+        // DEVICE_ID_ANY == DEVICE_EXT_ANY == 0
+        if ((p->source == source || p->source == 0) &&
```
pelikhan: why not using the constants?

tballmsft: they are not available in in base, which is independent of CODAL.

tballmsft:
All done - tested on simulator and pybadge.

PR #830, opened 2019-05-16
Added string isEmpty function in cpp

ChaseMor:
[In pxt, there is currently a shim](https://github.com/microsoft/pxt/blob/dba24b33ddabde0c65627be4c70de43744d4119a/libs/pxt-common/pxt-core.d.ts#L263) mapping to a string isEmpty function. To my understanding there isn't actually a function in the cpp that implements this. 

This adds that function.

Fixes Microsoft/pxt-microbit#2048

### DIFF
--- a/libs/base/core.cpp
+++ b/libs/base/core.cpp
@@ -650,6 +650,11 @@ int includes(String s, String searchString, int start) {
     return -1 != indexOf(s, searchString, start);
 }
 
+//%
+bool isEmpty(String s) {
+    return !s || IS_EMPTY(s);
+}
+
 } // namespace String_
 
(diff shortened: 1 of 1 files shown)


libs/base/core.cpp
```diff
@@ -650,6 +650,11 @@ int includes(String s, String searchString, int start) {
     return -1 != indexOf(s, searchString, start);
 }
 
+//%
```
pelikhan: we should get rid of this C++ function and use typescript instead...

ChaseMor: Added to the typescript with Microsoft/pxt#5583

PR #836, opened 2019-05-17
keep volume aligned to 32 when using systemmenu

jwunderl:
(also remove redundant mins/maxes as that's handled in setBrightness)

### DIFF
--- a/libs/game/systemmenu.ts
+++ b/libs/game/systemmenu.ts
@@ -268,17 +268,19 @@ namespace scene.systemMenu {
     }
 
     function volumeDown() {
-        const v = music.volume();
-        music.setVolume(v - 32);
+        const v = music.volume()
+        const newVolume = ((v - 1) >> 5) << 5;
+
+        music.setVolume(newVolume);
         music.playTone(440, 500);
     }
 
     function brightnessUp() {
-        screen.setBrightness(Math.min(screen.brightness() + 10, 100));
+        screen.setBrightness(screen.brightness() + 10);
     }
 
     function brightnessDown() {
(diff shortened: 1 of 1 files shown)


mmoskal:
Maybe it would be better to have non-linear jumps for volume? our hearing response is logarithmic; though I guess people will wonder why are they getting random values...

jwunderl:
Maybe non linear jumps and just show a progress bar to indicate 'relative loudness'?

mmoskal:
I think you also want the number in case they want to call `music.setVolume()`. I think that one should stay with amplitude, since this agrees with envelopes etc.

mmoskal:
though, maybe there should be no `music.setVolume()`; we should just persist user setting...

jwunderl:
A block for setting volume does seem like a pretty easy way for someone to make a jumpscare game now that i think about it (e.g. leave it alone and then halfway through the game max out volume in a game update and play the loudest screech you can).

libs/game/systemmenu.ts
```diff
@@ -269,16 +269,18 @@ namespace scene.systemMenu {
 
     function volumeDown() {
         const v = music.volume();
-        music.setVolume(v - 32);
+        const newVolume = ((v - 1) >> 5) << 5;
```
pelikhan: when v = 0, newVolume is -32. is that expected?

pelikhan: Can we do less magic for this?

jwunderl: Yeah, I was letting setVolume ignore it. I'll rewrite it to be less bit shifty

PR #841, opened 2019-05-23
Add ability to ramp frequencies of sounds

ChaseMor:
This allows for you to ramp the frequency of a sound over time. 

Adds the ```^``` operator which is followed by which frequency the sound should end at. For example,
```
music.playSound("!1300,200^50")
```
will play a sound that starts at 1300 hz, but ramps down to 50 hz over the course of 200 ms.

### DIFF
--- a/libs/mixer/melody.ts
+++ b/libs/mixer/melody.ts
@@ -24,6 +24,8 @@ namespace music {
 
     let globalVolume: number = null
 
+    const BUFFER_SIZE: number = 12;
+
     //% shim=music::enableAmp
     function enableAmp(en: number) {
         return // for sim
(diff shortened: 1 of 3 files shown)


libs/mixer/melody.ts
```diff
@@ -398,49 +412,46 @@ namespace music {
...
+                        case 'c': case 'C': note = 1; break;
+                        case 'd': case 'D': note = 3; break;
+                        case 'e': case 'E': note = 5; break;
+                        case 'f': case 'F': note = 6; break;
+                        case 'g': case 'G': note = 8; break;
+                        case 'a': case 'A': note = 10; break;
+                        case 'b': case 'B': note = 12; break;
+                        case 'r': case 'R': hz = 0; break;
+                        case '#': note++; break;
+                        case 'b': note--; break; // doesn't do anything
```
riknoll: I think you reverted https://github.com/microsoft/pxt-common-packages/pull/838

jedgarpark:
This is super awesome.

PR #842, opened 2019-05-23
dev branch

pelikhan:
Requires https://github.com/microsoft/pxt/pull/5597
DO NOT SQUASH


libs/servo/servo.ts
```diff
@@ -36,11 +38,15 @@ namespace servos {
...
+            this._angle = this.internalSetAngle(degrees);
         }
 
-        protected internalSetAngle(angle: number): void {
+        get angle() {
+            return this._angle || 90;
+        }
 
+        protected internalSetAngle(angle: number): number {
+            return 0;
```
abchatra: Why is this always 0?

pelikhan: overriden on base class, abstract breaks our compiler

PR #843, opened 2019-05-23
hide setGradient block

jwunderl:
fixes https://github.com/microsoft/pxt-adafruit/issues/1037

Should probably just remove the block for this, or remove the easing param, as it doesn't decompile and requires the handler as is - wanted to check if it's needed anywhere @pelikhan?


jwunderl:
the block doesn't compile to javascript, as there is no current handling for optional handler params (easing causes it to show up as an event, and then it gets mad when trying to decompile)

PR #846, opened 2019-05-28
Inlined jacdac

pelikhan:
To avoid tons of inclusion issues with npm / common-packages, copy the JACDAC-ts source in our sources.


jamesadevine:
Submodules? This is gonna be a mess

pelikhan:
Will fix once we merge common packages. Submodules are not supported by our cloud.

PR #854, opened 2019-05-31
Add rotary encoder support

mmoskal:


pelikhan:
I will add the sim stuff next week

mmoskal:
well, it needs blocks as well

libs/base/configkeys.h
```diff
@@ -95,6 +95,9 @@
 #define CFG_PIN_PWREN 77
 #define CFG_DISPLAY_TYPE 78
 
+#define CFG_ROTARY_ENCODER_PIN_A 79
+#define CFG_ROTARY_ENCODER_PIN_A 80
```
mmoskal: B?

libs/base/configkeys.h
```diff
@@ -95,6 +95,9 @@
 #define CFG_PIN_PWREN 77
 #define CFG_DISPLAY_TYPE 78
 
+#define CFG_ROTARY_ENCODER_PIN_A 79
+#define CFG_ROTARY_ENCODER_PIN_B 80
```
mmoskal: actually I think this should be `CFG_PIN_ROTARY_ENCODER_A` so we can use `PIN(ROTARY_ENCODER_A)`

libs/rotary-encoder/rotary.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Rotary and other encoders
+ */
+namespace encoders {
+/**
+ * Create a new rotary encoder connected to given pins
+ */
+//% blockId=inputCreateRotaryEncoder block="create rotary encoder with pins $pinA and B $pinB"
+//% weight=99
```
mmoskal: can you use this ctor function in blocks if you have `fixedInstances` on it?

pelikhan: no you can't so i'm not exposing it in blocks.

PR #855, opened 2019-06-01
Allow tile map to scale by powers of 2

jwunderl:
Converts some hardcoded ``<< 4`` and ``16``s to be based off a `scale` for the tile map -- where ``1 << scale`` corresponds to the width / height per tile. (couldn't think of a better name immediately). This also allows for changing the tile size to other powers of 2 - e.g. 8x8 as requested [here](https://forum.makecode.com/t/bug-cant-make-tilemaps-that-would-use-8x8-tiles/222):

creating as a draft as I haven't thoroughly tested / ran off dev yet. ~There's also one notable bug I've run into so far but haven't looked at yet at all: ``camera follow sprite`` is misbehaving with scales other than the default (still works with 16x16 tiles.~ fixed this

~Also, not sure how this would be exposed - just as a field isn't great (could make it an enum with 8x8, 16x16, 32x32?). And probably not at all in blocks to start at least?~ making scale an optional parameter when defining map; doesn't seem too common to set dynamically in blocks, but that can be done by just calling set map again if needed.

Also fixes an issue where tiles can have larger images than they should that creep into the transparency of the tiles to the right / below

Also here's the example from the post with the scale changed: https://makecode.com/_AYs4czfjLEbW

ezgif com-resize (11)


ladyada:
yay :) i think could be good for very dense gamemaps!

pelikhan:
Make the scale an enum with predefined values; then mount a block on the api to specify it.

jwunderl:
Yeah, merged dev into the branch and tested tmap collisions / drawing / events so it seems good to me now. I was only working on this for short periods so I was trying to avoid pulling all the dev branches until after microbit testing yesterday

pelikhan:
Lookin great!

jwunderl:
@pelikhan this should be fine to review / go into beta whenever you have time I think

ChaseMor:
The max velocity is tied to using 16x16 tile maps. So on slow computers that are computing a frame at 100 ms (The current max time step) or longer they can move a sprite up to 15 pixels. Basically, on slow machines, sprites will clip through the walls if moving fast enough.
https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/physics.ts#L28

riknoll:
@jwunderl and I were talking about switching to interpolation for the physics. The speed limit is getting increasingly annoying

hogan96:
> @jwunderl and I were talking about switching to interpolation for the physics. The speed limit is getting increasingly annoying

i think so as well

PR #856, opened 2019-06-03
support for 32 bit machines in VM

mmoskal:
also:
* optimize class member lookup
* free memory properly when deallocating image


mmoskal:
requires https://github.com/microsoft/pxt/pull/5628

PR #861, opened 2019-06-10
Pins ref doc update

ganicke:
- [x] Do some doc updates for SPI an I2C.
- [x] Fix pins card page.

RE: https://github.com/microsoft/pxt-adafruit/issues/864

### DIFF
--- a/libs/core/docs/reference/pins/spi-frequency.md
+++ b/libs/core/docs/reference/pins/spi-frequency.md
@@ -21,5 +21,3 @@ pins.spiFrequency(1000000);
 ## See also
 
 [spi mode](/reference/pins/spi-mode)
-
-[SPI handbook](https://developer.mbed.org/handbook/SPI)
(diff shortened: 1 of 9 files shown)


libs/core/docs/reference/pins/spi-frequency.md
```diff
@@ -8,8 +8,18 @@ pins.spiFrequency(1000000);
...
+```blocks
+pins.spiFrequency(1000000);
+```
 
 ## See also
 
-[SPI](https://developer.mbed.org/handbook/SPI)
+[spi mode](/reference/pins/spi-mode)
+
+[SPI handbook](https://developer.mbed.org/handbook/SPI)
```
pelikhan: We don’t use mbed anymore aside from Microbit. Remove

ganicke: ok...

PR #862, opened 2019-06-10
Better double->string conversion

mmoskal:
This no longer prints `1233.99999999999` instead of `1234`.

Fixes https://github.com/microsoft/pxt-microbit/issues/2019 - to be cherry-picked to master if everything works out

Tested on following numbers:
```
    13.7, 1234.56789, 1073741825,
    10737418250,
    107374182500,
    2000000001, 1, 0.1, 0.00001,
    1e101, 1.23456789e90
```

### DIFF
--- a/libs/base/core.cpp
+++ b/libs/base/core.cpp
@@ -1149,8 +1149,9 @@ TNumber neqq(TNumber a, TNumber b) {
 }
 
 
-// should not be more than 15, as 16 digits cannot be accurately represented
-// in 64 bit double and this code may crash
+// How many significant digits mycvt() should output.
+// This cannot be more than 15, as this is the most that can be accurately represented
+// in 64 bit double. Otherwise this code may crash.
 #define DIGITS 15
 
 void mycvt(NUMBER d, char *buf) {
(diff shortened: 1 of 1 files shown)


mmoskal:
also fixes https://github.com/microsoft/pxt-microbit/issues/1781

mmoskal:
this would need to run on hardware; I guess I could update the sim as well. This was my playground, translated directly to C++, could add to sim.

```typescript
declare namespace Math {
    //% shim=Math_::log10
    function log10(x: number): number;
}

function p10(k: number): number {
    if (k <= 0) return 1
    return 10 * p10(k - 1)
}

function conv(d: number): string {
    if (d < 0)
        return "-" + conv(-d)
    if (d == 0)
        return "0"

    let pw = Math.log10(d) | 0
    let e = 1

    if (d < 1e-6 || d > 1e21) {
        d /= p10(pw);
        e = pw;
        pw = 0;
    }

    const trg = 15
    let trailingZ = 0
    let dotAfter = pw + 1
    if (pw >= trg) {
        trailingZ = pw - trg + 1
        d /= p10(trailingZ);
    } else {
        d *= p10(trg - pw - 1)
    }
    d = Math.round(d)

    let r = ""

    if (dotAfter < 1) {
        r = "0."
        for (let i = 0; i < -dotAfter; ++i) {
            r += "0"
        }
    }

    for (let i = trg - 1; i >= 0; i--) {
        const q = p10(i)
        let k = 0
        while (d >= q) {
            d -= q
            k++
        }
        r += k
        if (d == 0 && (trg - i) >= dotAfter)
            break
        if ((trg - i) == dotAfter)
            r += "."
    }

    while (trailingZ-- > 0)
        r += "0"

    if (e != 1) {
        r += "e" + e
    }

    return r
}

let nums = [13.7, 1234.56789, 0.42,
    10737418250,
    107374182500,
    2000000001, 1, 0.1, 0.00001,
    1e101, 1.23456789e90]

game.onPaint(function () {
    let p = 2
    for (let n of nums) {
        screen.print(conv(n), 2, p, 3, image.font8)
        p += 10
    }
})
```

mmoskal:
I think I can use a uint64 instead of double for conversion, should be faster. Testing.

PR #873, opened 2019-06-19
Add support for alternative address for LIS3DH

mmoskal:
For this to work, the bootloader on pybadge needs to be updated with `ACCELEROMETER_TYPE = 0x30`; current bootloader has `0x32` which is correct for pygamer but not for pybadge (SA0 pin is connected to GND and 3V respectively). @dhalbert can we get new bootloader for pybadge with this correction?

@abchatra maybe this could go into the release? note that pygamer accelerometer should work already, this only fixes pybadge


mmoskal:
pybadge bootloader with fixed address for testing: 
[pybadge.zip](https://github.com/microsoft/pxt-common-packages/files/3307472/pybadge.zip)

mmoskal:
@dhalbert while you're at it, could you also update `0x18591ab9` (`BOOTLOADER_BOARD_ID`) to be a different random number for each board, so we have an easy way of telling them apart?

dhalbert:
Sorry, I missed the pings to me. I'll work on this very soon.

dhalbert:
Never mind, now I see https://github.com/adafruit/uf2-samdx1/pull/70.

PR #874, opened 2019-06-19
fix optional parameter for string.indexOf

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1088

### DIFF
--- a/libs/base/core.cpp
+++ b/libs/base/core.cpp
@@ -655,12 +655,19 @@ String substr(String s, int start, int length) {
 int indexOf(String s, String searchString, int start) {
     if (!s || !searchString)
         return -1;
-    auto lenA = s->getUTF8Size() - start;
-    auto lenB = searchString->getUTF8Size();
-    if (start < 0 || lenB > lenA)
-        return -1;
+
+    if (start < 0)
+        start = 0;
+
     auto dataA0 = s->getUTF8Data();
-    auto dataA = dataA0 + start;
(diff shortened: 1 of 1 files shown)


libs/base/core.cpp
```diff
@@ -655,13 +655,13 @@ String substr(String s, int start, int length) {
 int indexOf(String s, String searchString, int start) {
     if (!s || !searchString)
         return -1;
-    auto lenA = s->getUTF8Size();
+    auto lenA = s->getUTF8Size() - start;
```
pelikhan: start is the number of character, which might be greater than start in bytes.

jwunderl: Ah, so this would work in my case as I was using characters that would fit into a single byte, but break for e.g. chinese characters that would be split across a few bytes?

Is there a helper for length of a single unicode character or something similar for this? It looks like the current behavior is just doing a `memcmp` a few extra times for the partial characters, which won't work when we actually need the character counts here?

mmoskal: use `getUTF8DataAt(start)` - this gives pointer, you can subtract from that `getUTF8Data()` to get offset

`memcmp()` is fine, since no UTF8 character is a substring of another UTF8 character.

jwunderl:
Here's the test code I used, mainly copied from the issue: https://makecode.com/_YK7iw5UR702A

pelikhan:
@abchatra good for Microbit too

PR #875, opened 2019-06-21
Remove the metallic waveform

ChaseMor:
Removes support for metallic waveform.

Waveform 4 no longer maps to anything.

Fixes https://github.com/microsoft/pxt-arcade/issues/499

For simulator support, needs to be merged with https://github.com/microsoft/pxt/pull/5706


libs/mixer/melody.cpp
```diff
@@ -88,9 +83,7 @@ static gentone_t getWaveFn(uint8_t wave) {
     case SW_SAWTOOTH:
         return sawtoothTone;
     case SW_NOISE:
-        return fakeNoiseTone;
-    case SW_REAL_NOISE:
-        return realNoiseTone;
+        return noiseTone;
```
mmoskal: you could map 4 to noise as well, in case someone has a string with melody somewhere. or maybe not worth it...

ChaseMor: I chose not to so that we would have more flexibility and options in the future. The metallic waveform was pretty buggy to begin with when it came to melodies ([as the issue this fixes suggests](https://github.com/microsoft/pxt-arcade/issues/499)) so I don't think there are too many programs that use it. Plus, the real white noise doesn't support different frequencies so the melodies would break anyway.

PR #877, opened 2019-06-25
Text sprite object to manage state of text effects

shakao:
For character-level position-based text effects. Currently not exposed in blocks

### DIFF
--- a/libs/game/texteffects.ts
+++ b/libs/game/texteffects.ts
@@ -1,7 +1,7 @@
 namespace texteffects {
     export interface TextEffectState {
-        x_offset: number;
-        y_offset: number;
+        xOffset: number;
+        yOffset: number;
     }
 
     export enum TextEffectKind {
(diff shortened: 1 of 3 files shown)


libs/game/texteffects.ts
```diff
@@ -0,0 +1,106 @@
...
+            case texteffects.TextEffectKind.Shake:
+                return texteffects.shake;
+            default:
+                return null;
+        }
+    }
+ 
+    //% fixedInstances
+    export class TextEffect {
+        public getState: (index?: number, state?: TextEffectState) => any;
```
pelikhan: can we avoid "any"?

shakao:
eventually the textsprite will be used in the dialogs--we'll parse the string to create a sequence of text sprites, some with effects, and render those in the text blocks. i have some of that, but it's not clean enough to be merged in yet

PR #879, opened 2019-06-26
Physics interpolation

jwunderl:
closes https://github.com/microsoft/pxt-arcade/issues/1057; makes the max speed configurable (by default, bumps it up to 500 from 150)

Starting as a draft as it's a pretty decent size change, and there's a bit of clean up / sanity checking I still need to do.

Works about the same in most cases - a few fps slower in jumpy platformer which has the most sprites of any example I can remember (e.g. on pybadge, 18fps on the final level with the live site, 16 after this), and it scales a bit worse with huge numbers of sprites (e.g. > 50 on screen will be a bit more choppy, and high speeds will cause it to take more time as well). This can be mitigate a bit by changing the default values though (e.g. could bump up the step size parameter defaults a bit without hurting things too much)

(and I can fix the spacing of the ``Fx8`` functions before merging too, it's just hard to keep track of them when there are more than one or two collapsed into a single line so I split them up as they make the most sense to me)

new default max speed:

upped to 2k by changing the max explicitly:

and bouncing around at 500speed


libs/game/physics.ts
```diff
@@ -10,37 +10,55 @@ class PhysicsEngine {
...
 
-const MAX_DISTANCE = Fx8(15); // pixels
 const MAX_TIME_STEP = Fx8(100); // milliseconds
-const MAX_VELOCITY = Fx.div(MAX_DISTANCE, Fx.idiv(MAX_TIME_STEP, 1000));
-const NEG_MAX_VELOCITY = Fx.neg(MAX_VELOCITY)
-const GAP = Fx8(0.1);
+const SPRITE_CANNOT_COLLIDE = sprites.Flag.Ghost | sprites.Flag.Destroyed;
+const MIN_MOVE_GAP = Fx8(0.1);
+
+interface MovingSprite {
```
pelikhan: use a class, faster lookups

jwunderl: that actually seems to be enough to bring the speed back up to the current live site for jumpy platformer, wow (~16.5 fps to ~18.5 fps on the final level on meowbit) good call

libs/game/physics.ts
```diff
@@ -24,20 +24,19 @@ const MAX_TIME_STEP = Fx8(100); // milliseconds
...
-    // remaining x
-    dx: Fx8;
-    dy: Fx8;
-
-    // how much to move per step
-    xStep: Fx8;
-    yStep: Fx8;
+class MovingSprite {
+    constructor(
+        public sprite: Sprite,
```
pelikhan: do we support this?

jwunderl: yup!

libs/game/physics.ts
mmoskal: Object literal lookups are 5-10x slower than class lookups. Also object literals take maybe 3x memory.

PR #881, opened 2019-06-28
Fix crash in Segment::setLength manifesting in Object.keys()

mmoskal:
This also makes sure to interface with GC correctly in Object.keys() (though this wasn't the cause of the crash)

@jwunderl this fixes your 843


jwunderl:
Nice, thanks!

PR #883, opened 2019-06-30
Make small position adjustments respect tile map

jwunderl:
Make small adjustments in position respect the tile map; this makes it so behaviors like this don't cause the sprite to ignore the tile map:

Screen Shot 2019-06-30 at 12 51 33 PM

Also reorders the tile collision handler detection; it would overwrite the change if you tried to change position in the event handler, which is invoked from ``sprite.registerObstacle``, making the end position incorrect if you e.g. change the scene and move the sprite to the beginning of the level. (I think I want to change this to behave like sprite overlap events, with the 'run in parallel but drop when running' behavior, though - I'll add that change here in a second unless anyone strongly opposes it, as it feels very weird to have different behaviors between the two main types of 'collisions')

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -83,7 +83,6 @@ class Sprite implements SpriteLike {
     lifespan: number;
     private _image: Image;
     private _obstacles: sprites.Obstacle[];
-    private ongoingObstacles: sprites.Obstacle[];
 
     private updateSay: (dt: number, camera: scene.Camera) => void;
     private sayBubbleSprite: Sprite;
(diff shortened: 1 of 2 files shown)


libs/game/sprite.ts
```diff
@@ -673,9 +691,21 @@ class Sprite implements SpriteLike {
...
+
+            if (!handlerOccuring) {
+                collisionHandlers
+                    .filter(h => h.kind == this.kind())
+                    .forEach(h => {
+                        this.ongoingObstacles.push(other);
+                        control.runInParallel(() => {
+                            h.handler(this);
+                            this.ongoingObstacles.removeElement(other);
+                        });
```
pelikhan: Clear array if done

libs/game/sprite.ts
```diff
@@ -673,9 +691,21 @@ class Sprite implements SpriteLike {
...
+            if (!this.ongoingObstacles) this.ongoingObstacles = [];
+            const handlerOccuring = this.ongoingObstacles
+                .some(obs => obs.x === other.x && obs.y === other.y);
+
+            if (!handlerOccuring) {
+                collisionHandlers
+                    .filter(h => h.kind == this.kind())
+                    .forEach(h => {
+                        this.ongoingObstacles.push(other);
+                        control.runInParallel(() => {
```
pelikhan: We should avoid runInParallel, it allocates news fibers and messes up scenes

jwunderl: Maybe a ``scene.runInParallel``, like ``game.forever``?

PR #866, opened 2019-06-17
Add gcPreAllocateBlock()

mmoskal:
This needs separate porting to `dev`, so do not merge to `master`

PR #884, opened 2019-07-02
Cherry-pick Add gcPreAllocateBlock() into master

darzu:
See https://github.com/microsoft/pxt-common-packages/pull/866/files which was completed against stable6.9


pelikhan:
@mmoskal do we need this fix?

PR #885, opened 2019-07-10
make console log take any to match browser

jwunderl:
make ``console.log`` accept ``any`` like the dom version. Possibly noteworthy that ``text.shadowOptions.toString=true`` makes the blocks version already behave like this, except the string coercion shows up in the javascript after decompiling instead of just being part of log

@mmoskal are there any potential perf / object size issues that would block this?

### DIFF
--- a/libs/base/console.ts
+++ b/libs/base/console.ts
@@ -53,10 +53,10 @@ namespace console {
      */
     //% weight=90
     //% help=console/log blockGap=8
-    //% blockId=console_log block="console|log %text"
-    //% text.shadowOptions.toString=true
-    export function log(text: string): void {
-        add(ConsolePriority.Log, text);
+    //% blockId=console_log block="console log $value"
+    //% value.shadow=text
+    export function log(value: any): void {
+        add(ConsolePriority.Log, value + "");
     }
 
     /**


jwunderl:
Yeah, I noticed even array didn't have a ``toString``, and adding it to pxt-core.d.ts / pxt-helpers didn't make it auto-invoke when coercing it to be a string. 

Not that much work to support pretty printing objects now though if we really wanted, it's a bit ugly as I wrote it though?

https://makecode.com/_dfTAviieFLgM

```typescript
function log(obj: any) {
    if (typeof obj == "string") {
        console.log(obj);
    } else if (typeof obj == "number") {
        console.log("" + obj)
    } else if (Array.isArray(obj)) {
        console.log((obj as Array<string>).join(","));
    } else if (obj.toString) {
        console.log(obj.toString());
    } else {
        const keys = Object.keys(myObj);
        console.log(`{
    ${
            keys
                .map(key => key + ": " + obj[key])
                .join(",\n\t")}
}`);
    }
}
```

mmoskal:
I guess we could have `control.inspect(v:any): string` like the one your wrote and then have `console.log` use that; it would be probably smart to limit the string length in that one though (i.e., if an array with 1000 elements is passed)

jwunderl:
Hm, that might be nice then. I'll merge this change as is and write that up then, as they seem like distinct changes to me

PR #886, opened 2019-07-11
Console inspect and pretty printing for objects / arrays in console.log

jwunderl:
re: discussion in https://github.com/microsoft/pxt-common-packages/pull/885

~Had to work around minor bug: https://github.com/microsoft/pxt-arcade/issues/515~

Noticed that the default toString for objects is ``[Object]`` on pygamer in arcade, rather than ``[object Object]``; could fix that I suppose, as it looks like it's defined here https://github.com/microsoft/pxt-common-packages/blob/68c76b90636c9dcd9ec3de0774a06e5d58ec0316/libs/base/core.cpp#L408

Also, the ``console overlay`` doesn't handle new lines nicely (it just offsets by a single line per console.log rather than by how many lines it takes up), will fix that separately

here's the test I used: https://makecode.com/_9vd7opftoaqP

```typescript
game.consoleOverlay.setVisible(true)
const myObj = {
    a: "a",
    b: "b",
    c: "c",
    d: "d",
    e: "e",
    s: sprites.create(img`2`)
}

const myLargeObj = {
    a: "a",
    b: "b",
    c: "c",
    d: "d",
    e: "e",
    f: "f",
    g: "g",
    h: "h",
    i: "i",
    j: "j",
    k: "k",
    l: "l",
    m: "m",
    n: "n",
    o: "o",
    p: "p",
    q: "q",
    r: "r",
    s: sprites.create(img`2`),
    t: "t",
    u: "u",
    v: "v",
    w: "w",
    x: "x",
    y: "y",
    z: "z"
}
let myLargeArr = [0]
for (let i = 1; i < 100; i++) myLargeArr.push(i);

console.log("hello")
console.log([1, 2, 3])
console.log(sprites.create(img`1`))
console.log(1)

console.log(myObj)
console.log(myLargeObj)
console.log(myLargeArr)
```

with output

```
sim.js:3449 l>hello
sim.js:3449 l>1,2,3
sim.js:3449 l>2(79,59)->(0,0)
sim.js:3449 l>1
sim.js:3449 l>{
    a: a,
    b: b,
    c: c,
    d: d,
    e: e,
    s: 0(79,59)->(0,0)
}
sim.js:3449 l>{
    a: a,
    b: b,
    c: c,
    d: d,
    e: e,
    f: f,
    g: g,
    h: h,
    i: i,
    j: j,
    k: k,
    l: l,
    m: m,
    n: n,
    o: o,
    p: p,
    q: q,
    r: r,
    s: 1(79,59)->(0,0),
    t: t
    ...
}
sim.js:3449 l>0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19...
```

### DIFF
--- a/libs/base/console.ts
+++ b/libs/base/console.ts
@@ -74,18 +74,19 @@ namespace console {
     /**
      * Convert any object or value to a string representation
      * @param obj value to be converted to a string
+     * @param maxElements [optional] max number values in an object to include in output
      */
-    export function inspect(obj: any): string {
+    export function inspect(obj: any, maxElements = 20): string {
         if (typeof obj == "string") {
             return obj;
         } else if (typeof obj == "number") {
             return "" + obj;
         } else if (Array.isArray(obj)) {
             const asArr = (obj as Array<string>);
-            if (asArr.length <= 20) {
+            if (asArr.length <= maxElements) {
                 return asArr.join(",");
             } else {
-                return `${asArr.slice(0, 20).join(",")}...`;
+                return `${asArr.slice(0, maxElements).join(",")}...`;
             }
         /** TODO: should handle objects with toString like this:
          * } else if (obj.toString) {
@@ -101,9 +102,9 @@ namespace console {
             }
 
             let keys = Object.keys(obj);
-            const snipped = keys.length >= 20;
+            const snipped = keys.length > maxElements;
             if (snipped) {
-                keys = keys.slice(0, 20);
+                keys = keys.slice(0, maxElements);
             }
 
             return`{
--- a/libs/base/console.ts
+++ b/libs/base/console.ts
@@ -112,8 +112,8 @@ namespace console {
     ${
         keys
             .map(key => key + ": " + obj[key])
-            .join(",\n\t")
-        + (snipped ? "\n\t..." : "")
+            .join(",\n    ")
+        + (snipped ? "\n    ..." : "")
     }
 }`;
         }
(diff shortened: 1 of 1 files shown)


mmoskal:
testing for `o.toString` wouldn't work in javascript, since it's defined on `object.prototype`, I think

jwunderl:
Ah, good point, too used to the pxt editor version where it's not implicitly defined on objects. Guess there isn't a non 'try it and see if it works' approach for testing against toString then

libs/base/console.ts
```diff
@@ -71,6 +71,46 @@ namespace console {
...
+
+            let keys = Object.keys(obj);
+            const snipped = keys.length > maxElements;
+            if (snipped) {
+                keys = keys.slice(0, maxElements);
+            }
+
+            return `{${
+                keys.reduce(
+                    (prev, currKey) => prev + `\n    ${currKey}: ${obj[currKey]}`,
```
pelikhan: call inspect again?

jwunderl: Currently, any member of the object without a toString will end up as `[object Object]`. I didn't want to make it recursive right away as it seems like it might be too much info to show at once. I think it works nicely in chrome cause you can expand / collapse at will, but non-trivial objects might be really hard to parse if it shows everything right away - I suppose it could just have depth for how far to recurse as a parameter though

Also have to deal with cases like if a node containing a reference back to it's parent if we made it recurse.

Not ideal of course, can look into making this more descriptive - for now I figured showing the  first 'layer' of properties is helpful for giving quick info on what's in the object

mmoskal: I think this PR is an improvement over current state. I guess the real solution would need to keep track of the length of the string so far and stop printing when it exceeds some limits, but this is generally tricky. I would say good enough for now

PR #887, opened 2019-07-11
handle multi line strings in console overlay

jwunderl:
splits up lines so multi line strings fit nicely into the console overlay; e.g. this:

instead of this:

this was annoying with https://github.com/microsoft/pxt-common-packages/pull/886, as the pretty printing wasn't that pretty due to everything overlapping

Also add optional parameter to to ``.setVisible``, so that people can set the color if they want to

https://makecode.com/_1y3W2ceD8C3J

### DIFF
--- a/libs/game/console.ts
+++ b/libs/game/console.ts
@@ -4,7 +4,7 @@ namespace game.consoleOverlay {
     const marginx = 4;
     const marginy = 2;
     const consoleFont = image.font5;
-    const consoleLines = Math.floor(screen.height / (consoleFont.charHeight + marginy)) - 1;
+    const MAX_CONSOLE_LINES = Math.floor(screen.height / (consoleFont.charHeight + marginy)) - 1;
     console.addListener(listener);
 
     export function isVisible() {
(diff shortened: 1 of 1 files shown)


libs/game/console.ts
```diff
@@ -15,31 +15,40 @@ namespace game.consoleOverlay {
...
+        if (col !== undefined)
+            consoleColor = col;
     }
 
     function listener(priority: ConsolePriority, text: string) {
         if (!consoleStrings)
             return;
-        else if (consoleStrings.length < consoleLines)
+
+        let newLineIndex = text.indexOf("\n");
```
pelikhan: why are we not using "split"?

libs/game/console.ts
```diff
@@ -15,31 +15,40 @@ namespace game.consoleOverlay {
...
+        if (consoleStrings.length > MAX_CONSOLE_LINES) {
+            consoleStrings.splice(0, consoleStrings.length - MAX_CONSOLE_LINES);
         }
     }
 
     export function draw() {
         if (!consoleStrings || scene.systemMenu.isVisible()) return;
-        const height = consoleFont.charHeight + marginy ;
-        const top = 2 + (game.stats ? height : 0);
+        const lineHeight = consoleFont.charHeight + marginy;
```
pelikhan: unrelated change

libs/game/console.ts
jwunderl: Oops, must have just blanked for a bit, I did this right before leaving for the day. fixed

libs/game/console.ts
```diff
@@ -15,27 +15,29 @@ namespace game.consoleOverlay {
...
             return;
-        else if (consoleStrings.length < consoleLines)
-            consoleStrings.push(text);
-        else {
-            for (let i = 1; i < consoleStrings.length; ++i) {
-                consoleStrings[i - 1] = consoleStrings[i];
-            }
-            consoleStrings[consoleStrings.length - 1] = text;
+
+        text.split("\n")
```
pelikhan: is fast to do array.concat?

jwunderl: concat returns a new array, so it would have to push all the old strings too instead of just the new ones https://github.com/microsoft/pxt/blob/master/libs/pxt-common/pxt-helpers.ts#L193

PR #898, opened 2019-07-24
Adding palette/colors/storyboard packages

pelikhan:
Centralizing color utilities scatered accross our packages and adding a very lightweight palette project (as a separate package to avoid overloading arcade)
- [x] refactor color compulation routines into "colors" package (from light, palette)
- [x] define a "ColorBuffer" class to be used in  various settings, including light/palette
- [x] lightweight platte utilities
- [x] lightweight storyboard project (scene orchestration)

## storyboard

The storyboard allows to register functions as "scenes". It maintains a stack of such "frames" and allows to push entire games and pop them without running out of memory. It also brings in the concept of boot sequence which runs at the start of a storyboard.

There is a few niceties that still need to be added in the future: smoother background color transiation in boot sequence, transitions between games (using screen effects?)...


jwunderl:
Ah, so with this I should clean up my pxt-color extension to just have a dependency on these then?

pelikhan:
Yes and add missing APIs in here if needed

libs/color/colors.ts
```diff
@@ -0,0 +1,213 @@
...
+    Pink = 234
+}
+
+/**
+ * Color manipulation
+ */
+//% advanced=1
+namespace color {
+    /**
+ * Converts red, green, blue channels into a RGB color
```
jwunderl: spacing

libs/color/colors.ts
```diff
@@ -0,0 +1,213 @@
...
+     * Fade the color by the brightness
+     * @param color color to fade
+     * @param brightness the amount of brightness to apply to the color, eg: 128
+     */
+    //% blockId="colorsfade" block="fade %color=neopixel_colors|by %brightness"
+    //% brightness.min=0 brightness.max=255
+    //% help="light/fade"
+    //% group="Color" weight=18 blockGap=8
+    //% blockHidden=true
+    export function fade(color: number, brightness: number): number {
```
jwunderl: ``fade`` would be overloaded a bit with my extension (which has fades between different color palettes), if this needs to stay the same name for backwards compat though I can figure something out / disambiguate somehow

mmoskal:
```
 export function fade(color: number, brightness: number, otherColor = 0): number 
```
?

pelikhan:
don't change anything until this package hits master!

libs/color/colors.ts
pelikhan: this is fade between 2 colors, not color buffers.

pelikhan:
> add my a version of my package to common-packages by the end of this weekend and fix any minor linting/etc then.

I want to keep the "color" package smallish so let's not bulk add stuff to it.

jwunderl:
Of course, I'd be porting it into a package in common packages and leaving my extension unchanged in the current repo until master is bumped.

I'd add as much of the extension to it's own package (e.g. Color Fades), just adding the things to the other packages if they're strictly necessary (though looking at the changes again I can't think of anything right away that I'd need to add to these packages, shouldn't be much)

libs/color/colors.ts
jwunderl: Maybe I just change my package to have the class ``PaletteFade`` instead of ``Fade`` then when merging in, and rename the necessary functions, easy change and more clear what the class is for anyway

libs/color/colorbuffer.ts
```diff
@@ -0,0 +1,100 @@
...
+            this.buf = buf;
+            this.layout = layout || ColorBufferLayout.RGB;
+        }
+
+        get stride() {
+            return this.layout == ColorBufferLayout.RGB ? 3 : 4;
+        }
+
+        get length() {
+            return (this.buf.length / this.stride) | 0;
```
mmoskal: `Math.idiv` ?

PR #899, opened 2019-07-27
add support for on-device storage

mmoskal:
TODO:
* [ ] think about scoping file names per-application; maybe clear the non-scoped stuff preemptively on start
* [ ] move flash interfaces to CODAL
* [ ] think about low-level end-user APIs
* [ ] (in 16 bit mode) make sure no 32 bit word is written more than twice for nordic

cc @jwunderl @pelikhan


libs/core/usb.cpp
```diff
@@ -47,7 +47,7 @@ static const DeviceDescriptor device_desc = {
 static void start_usb() {
     // start USB with a delay, so that user code can add new interfaces if needed
     // (eg USB HID keyboard, or MSC)
-    fiber_sleep(100);
+    fiber_sleep(1000);
```
mmoskal: review this

kevinjwalters:
What storage would be used on an Adafruit Pygamer or Kittenbot Meowbit, for example?

mmoskal:
Superseded by https://github.com/microsoft/pxt-common-packages/pull/923

PR #901, opened 2019-07-29
Adding the melody editor block to the music drawer

kimprice:
A new category was created in the music drawer called "melody" where the new block will reside. The block uses `playUntilDone()` and therefore is blocking. This allows multiple melodies to be stacked to create longer melodies. The picture below is an example of how to play a melody in the background of a game and conditionally change the type of music.
musicInBackground

melodyCategoryHigher


pelikhan:
This block should be near the top of the drawer. All this work to hide it at the bottom.

libs/mixer/melody.ts
```diff
@@ -110,6 +110,60 @@ namespace music {
...
+    //% blockNamespace=music
+    //% weight=75 blockGap=8
+    //% group="Melody"
+    //% melody.shadow="melody_editor"
+    //% tempo.min=40 tempo.max=500
+    //% tempo.defl=120
+    export function playMelodyFromEditor(melody: string, tempo: number) {
+        // add tempo to string so it is reflected in simulation
+        // creates format like "C5-174 B4 A G F E D C "
+        let formattedMelody = melody.substr(0, melody.indexOf(' ')) + "-" + tempo + melody.substr(melody.indexOf(' '), melody.length);
```
riknoll: Break up this expression, maybe move the indexOf result into a variable

kimprice:
> This block should be near the top of the drawer. All this work to hide it at the bottom.

Yea I think it needs to be moved higher as well. The whole drawer may need to be rearranged, though. There are some blocks that are not compatible as well. Like the set tempo block does not change the tempo of the melody from the melody editor.

libs/mixer/melody.ts
```diff
@@ -110,6 +110,60 @@ namespace music {
 
     let playToneID = 0
 
+    /**
+     * Play a melody from the melody editor.
```
riknoll: This comment should include a breakdown of the melody format; otherwise it might be tough to use this from javascript. Also, add descriptions for both of the parameters.

libs/mixer/melody.ts
```diff
@@ -110,6 +110,60 @@ namespace music {
...
+    //% blockNamespace=music
+    //% blockHidden = true
+    //% weight=75 blockGap=8
+    //% group="Melody" duplicateShadowOnDrag
+    //% melody.fieldEditor="melody"
+    //% melody.fieldOptions.decompileLiterals=true
+    //% melody.fieldOptions.decompileIndirectFixedInstances="true"
+    //% melody.fieldOptions.onParentBlock="true"
+    export function melodyEditor(melody: string): string {
+        let notes: string[] = melody.split(" ");
```
riknoll: Why not just move this parsing code into the other function? Typically when we have a function for a field editor we just have it do nothing and put all of the logic in the actual API. We have a special comment annotation `SHIM=TD_ID` which will make it so that the blocks compiler/decompiler will ignore this function and not emit it.

kimprice: Thanks, it looks much cleaner now!

libs/mixer/melody.ts
kimprice: I rewrote it this since I moved the parsing into here and it's much easier to read now. :)

libs/mixer/melody.ts
```diff
@@ -110,6 +110,65 @@ namespace music {
 
     let playToneID = 0
 
+    /**
+     * Play a melody from the melody editor.
+     * @param melody - string of up to eight notes [C D E F G A B C5] separated by spaces, 
+     * which will be played one at a time, eg: "E D G F B A C5 B "
```
riknoll: Don't use eg:, it actually has special meaning in our comment parser. The new syntax is to do `melody.defl="whatever"`

libs/mixer/melody.ts
```diff
@@ -110,6 +110,65 @@ namespace music {
...
+     */
+    //% block="play melody $melody at tempo $tempo|(bpm)" blockId=playMelody
+    //% blockNamespace=music
+    //% weight=85 blockGap=8
+    //% group="Melody"
+    //% melody.shadow="melody_editor"
+    //% tempo.min=40 tempo.max=500
+    //% tempo.defl=120
+    export function playMelody(melody: string, tempo: number) {
+        let notes: string[] = melody.split(" ");
```
riknoll: This will be a bit more resilient to whitespace if you filter out the empty strings
```suggestion
        let notes: string[] = melody.split(" ").filter(n => !!n);
```

chino:
Not sure where to report this but I seem to get "calling non-function" when trying to call play melody from a forever block.

riknoll:
@chino please open an issue in the relevant editor repo ([microbit](https://github.com/microsoft/pxt-microbit/issues), [arcade](https://github.com/microsoft/pxt-arcade/issues), [adafruit](https://github.com/microsoft/pxt-adafruit/issues)) and include a share link with the code that is giving you the error. You can create a share link for a project like this:



Thanks!

PR #906, opened 2019-07-31
New animation extension

dannytech:
This PR would replace the existing animation extension with a new one containing some major functionality changes. The overall objectives were to reduce the amount of blocks needed to run an animation to one and add movement animations (animation sprite position over time).

Changes:

* The old animation blocks are now hidden. They have been replaced with two blocks, one for image animations and one for movement animations. There is another block that returns SVG path strings to be consumed by the movement animation block.
* Image animations and movement animations can run in parallel.
* Animations are now tied to a sprite when they are created and are immediately run after being created, removing the need for extra blocks to perform those actions.
* Animations can now be configured as either blocking or non-blocking..

Future changes:

* Due to the number of parameters, the block wraps inputs into lines. This has the unfortunate effect of moving the "ms" text after the time picker to the next line.
* Loop and stop blocks will be added to remove the need for manually looping and allow running animations to be interrupted.


libs/animation/animation.ts
```diff
@@ -1,13 +1,52 @@
...
 namespace animation {
-    //Handles all the updates
-    let animations: Animation[];
+    export interface AnimationState {
+        movementAnimation?: MovementAnimation,
+        imageAnimation?: ImageAnimation
+    }
+
+    // Stores the animations for the current scene
+    let animations: AnimationState;
```
pelikhan: animations should be stored in the scene.state.

libs/animation/animation-old.ts
```diff
@@ -0,0 +1,178 @@
+/*
+    Animation library for sprites
+*/
+//% color="#03AA74" weight=78 icon="\uf021"
+namespace animation {
+    //Handles all the updates
+    let animations: OldAnimation[];
```
pelikhan: IMO, this should be a new animation library (other name) which we keep the old one untouched for legacy.

libs/animation/animation.ts
```diff
@@ -1,13 +1,52 @@
+enum AnimationPath {
```
pelikhan: use fixedInstances instead of enums to make the animation system extensible.

libs/animation/animation.ts
```diff
@@ -1,13 +1,52 @@
...
+    export interface AnimationState {
+        movementAnimation?: MovementAnimation,
+        imageAnimation?: ImageAnimation
+    }
+
+    // Stores the animations for the current scene
+    let animations: AnimationState;
 
+    // Preserves animations when switching back and forth between scenes
     let animationStateStack: {
```
pelikhan: you can store state in the scene itself: 

    let state = scene.state["__animations"] as YOURDATA

libs/animation/animation.ts
```diff
@@ -26,148 +65,483 @@ namespace animation {
...
-    //% enumName="ActionKind"
-    //% enumMemberName="action"
-    //% enumPromptHint="e.g. Walking, Idle, Jumping, ..."
-    //% enumInitialMembers="Walking, Idle, Jumping"
-    //% weight=10
-    export function _actionEnumShim(arg: number) {
-        // This function should do nothing, but must take in a single
-        // argument of type number and return a number value.
-        return arg;
+    export interface Animation {
```
pelikhan: use classes instead of interfaces

pelikhan:
Please add usage examples in test.ts.

pelikhan:
Could you also add screenshot of all the blocks?

libs/animation/animation.ts
dannytech: Is scene.state a very new feature? I'm not seeing it anywhere.

pelikhan: It is a property on the current scene (game.currentScene())

dannytech: I'm not seeing a state property there. Here's where I'm looking: https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/scene.ts.

Are you suggesting I just add a property to the scene class to store animations?

libs/animation/animation-old.ts
dannytech: We discussed this before, and decided to keep everything in the same library. I've tested for backwards compatibility and the old blocks will still work but aren't able to be added. We also can't re-namespace it or the old blocks would be destroyed.

libs/animation/animation.ts
pelikhan: https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/scene.ts#L163

dannytech: That property is read-only (no setter).

pelikhan: You can did fields tonit. ...data[“dhrveihs”] = ...

jwunderl: `game.currentScene().data["animationState"] = yourValue`

dannytech:
image

Here's what the blocks look like currently. The "ms" wraps to the next line, unfortunately, but @riknoll and @shakao suggested that the best solution would be to have a flag on the timePicker to show "ms" within the picker. If you have another suggestions, that would be useful.

In regards to the test.ts file, I assume this is just to test that decompilation works properly? Let me know if the current contents work.

pelikhan:
it would be good to flappy duck using this api.

parameter order seems reversed on the animate block, should be consistent with follow.

pelikhan:
We should not have an animation api that waits. It will hog the game update loop.

dannytech:
@pelikhan Here's the updated look based on your feedback. Changes are removing the blocking (wait) option, switching from an interval for movement animations to a period, and fixing the order of parameters to be more consistent. We're working on a separate mechanism for looping and stopping animations, which will be in a later PR.

image

pelikhan:
You need to move it to another package. You did renaming of the TypeScript classes that would break published scripts.

abchatra:
@pelikhan, We (@riknoll , @shakao & me) discussed this yesterday. Plan is to just add new blocks to the existing extension and keep the existing blocks. New blocks will be at the top and old blocks will be deprecated at a phased manner. 3 steps in future. 
1. Keep it there for now at the bottom of the category
2. Mark deprecated (text or UI cue)
3. Remove from the animation extension.

pelikhan:
Sounds like a plan.

2.5: update all known content using the old blocks (after release)

abchatra:
Yes, we should update all the existing our content use nicer apis.

libs/animation/animation-old.ts
```diff
@@ -142,24 +142,24 @@ namespace animation {
...
+        return new Animation(action, interval);
     }
 
     /**
      * Attach an animation to a sprite
      */
     //% blockId=attachAnimation
     //% block="attach animation $set=variables_get(anim) to sprite $sprite=variables_get(mySprite)"
-    //% blockHidden=true deprecated
+    //% group="Old blocks (DEPRECATED)"
```
abchatra: Just mark them as 'advanced' instead of "Old blocks (DEPRECATED)"

abchatra: We will move to deprecated at a later point of time

libs/animation/animation.ts
```diff
@@ -1,173 +1,624 @@
...
+                }
+                case "q": { // q dx1 dy1 dx2 dy2
+                    const p1 = new Point(p0.x + args[0], p0.y + args[1]);
+                    const p2 = new Point(p0.x + args[2], p0.y + args[3]);
+                    node = new QuadraticCurveTo(p0, p1, p2);
+                    break;
+                }
+                case "T": { // T x2 y2
+                    let lastControlPoint: Point;
+                    if (metadata.lastNode instanceof QuadraticCurveTo) {
```
jwunderl: Could there be a ``lastControlPoint()`` method on ``PathNode`` to describe this behavior / not repeat it for TtSs? I would imagine it just returning undefined by default and being overridden in ``QuadraticCurveTo`` and ``CubicCurveTo``

libs/animation/animation.ts
```diff
@@ -1,173 +1,624 @@
...
-                    if (sprite.image !== newImage) {
-                        sprite.setImage(newImage);
+        public static parse(pathStart: Point, pathString: string): Path {
+            let path: Path = new Path();
+            let p0: Point = pathStart;
+
+            // This implementation of SVG parsing does not support the A/a commands, nor does it support exponents in arguments
+            const digits = "0123456789";
+            const separators = ", \t\n\r\f\v";
+            const commands: {
```
jwunderl: Might be worth pulling this out into a helper function that just returns based off a switch statement, obj literals are much slower and take up much more space on hw

libs/animation/animation.ts
```diff
@@ -1,173 +1,624 @@
...
+                //   - if there's sufficient data to make a node during this step, create it and continue
+                // - else if it's a plus/minus sign, and if it's the start of a new argument, add it to allow for positive/negative numbers
+                // - if it's the end of the string, complete the current argument before proceeding to the next step
+                // - if there's sufficient data to make a node after all of these steps, create it
+                if (digits.indexOf(char) > -1) { // Parses number arguments
+                    currentArg += char;
+                } else if (separators.indexOf(char) > -1 && currentArg) { // Terminates number arguments
+                    args.push(parseInt(currentArg));
+                    currentArg = "";
+                } else if (Object.keys(commands).indexOf(char) > -1) { // Parses command arguments
```
jwunderl: This might change re: the last comment, but if not it seems preferable to pull out the ``Object.keys(commands)`` here into a const to avoid recreating the same array over and over again. If that last comment is implemented I would imagine you'd just return -1 in the default case there and handle based of that, though

libs/animation/test.ts
```diff
@@ -0,0 +1,55 @@
...
+    500
+)
+animation.runMovementAnimation(
+    mySprite,
+    animation.animationPresets(animation.flyToCenter),
+    500
+)
+animation.loopImageAnimation(
+    mySprite,
+    [img`
```
jwunderl: Very minor, but maybe add one extra frame to each of these so they change, and also create a sprite for ``mySprite``? Ideally you'd be able to copy and paste this into the editor with the animation extension loaded and have something work

libs/animation/animation.ts
```diff
@@ -1,173 +1,624 @@
...
+            this.lastNode = -1;
         }
 
-        _init() {
-            if (!animations) {
-                animations = [];
-                game.eventContext().registerFrameHandler(scene.ANIMATION_UPDATE_PRIORITY, () => {
-                    animations.forEach(anim => anim.update());
-                });
+        private static generateNode(p0: Point, command: string, args: number[], metadata: { pathStart: Point, lastNode: PathNode }): PathNode {
```
jwunderl: don't think it actually matters, but would it be reasonable to switch ``metaData`` to be a tuple instead and then you can just destructure it?

```suggestion
        private static generateNode(p0: Point, command: string, args: number[], metadata: [Point, PathNode]): PathNode {
            const pathStart, lastNode = metadata;
```

Would need to change bits in the rest of the function. Just a quick thought, don't know if it's actually a good idea at all

libs/animation/animation.ts
dannytech: Could you clarify which parts would be pulled into their own helper function? What exactly would the function be accomplishing? Or would it be multiple, like isDigit, isSeparator, etc.?

jwunderl: something like 

```typescript
function commandToArgCount(command: string): number {
    switch (command) {
        case "M":
        case "m":
            return 4;
        ...
        default: return -1
    }
}
```

just to replace commands here

riknoll:
@pelikhan do you want to take another look?

libs/animation/animation.ts
```diff
@@ -1,173 +1,648 @@
...
+                }
+                case "q": { // q dx1 dy1 dx2 dy2
+                    const p1 = new Point(p0.x + args[0], p0.y + args[1]);
+                    const p2 = new Point(p0.x + args[2], p0.y + args[3]);
+                    node = new QuadraticCurveTo(p0, p1, p2);
+                    break;
+                }
+                case "T": { // T x2 y2
+                    let lastControlPoint: Point = lastNode.getLastControlPoint();
+                    if(!lastControlPoint) break;
```
jwunderl: spacing on these lines (``if (!lastControlPoint)``

dannytech: Good catch, fixed!

abchatra:
Is there a link to build which I can try out?

riknoll:
@abchatra here you go: https://arcade.makecode.com/app/03e068b0cc4c98c733bb23c9f7a23856fb75a539-42185836d1#editor

jwunderl:
Here's jumpy platformer mostly switched over, but still buggy: https://makecode.com/_9qD33H6Ru90p

Main issue with converting that is that the current animations has a pretty nice behavior that jumpy platformer was using, where if you activate the same animation that is currently running it will just continue the animation that was running. Without that it's a bit hard to do the different sequences -- e.g. the ``on game update`` that controls the hero sprite's animation; I guess I would need it to manage the state separately, only activate the animation when the state has changed, and stop all animations before doing so?

Also, should ImageAnimations maybe clear any other ImageAnimations on the sprite when they are started? I suppose it makes sense to have a main looping animation and then a single one that runs as well, but if you add two looping animations to the same sprite the first one will be running in the background doing nothing until you explicitly remove all animations on the sprite.

abchatra:
This block is missing:
image

abchatra:
make all as the default.  Currently frame is default. 
image

dannytech:
@jwunderl Working on a feature to clear any existing animations of the same type and on the same sprite as the new animation. Should I implement a check to see if that same animation is already running or just barrel on and overwrite it?

@abchatra As for the block containing animation states, should this really be there? Is it used anywhere but inside a create/activate block?

abchatra:
You can do something like this
image

We should keep all the legacy blocks for now.

jwunderl:
If possible, making it behave like the current one would be good in my opinion - if the animation is currently running with the same contents don't do anything, but if it's a different animation clear the old one and start the new one (partially because I think that will make the link I shared 'just work', and make upgrading old games much easier / more consistent). I don't think this one's necessarily a blocking thing for this PR though, can be figured out later as well as needed

PR #907, opened 2019-07-31
Esp32 refactoring

pelikhan:
- [ ] expose pins in configuration
- [ ] move request code to net
- [ ] refactor to allow Serial impl for esp32


libs/base/configkeys.h
```diff
@@ -100,6 +100,16 @@
 
 #define CFG_ACCELEROMETER_SPACE 81
 
+#define CFG_PIN_ESP32_MOSI 82
```
mmoskal: should this be `CFG_PIN_WIFI_MOSI` etc?

libs/esp32-game/settings.ts
```diff
@@ -0,0 +1,57 @@
+
+function wifiSystemMenu() {
+    const wifi = esp32spi.defaultController();
+    if (!wifi) {
+        game.splash("wifi not configured");
```
mmoskal:
```suggestion
        game.splash("WiFi module not configured");
```

PR #908, opened 2019-08-01
explicit random seed

mmoskal:


libs/base/eventcontext.ts
```diff
@@ -77,7 +83,12 @@ namespace control {
             control.enablePerfCounter("all frame callbacks")
 
             let loopStart = control.millis()
-            this.deltaTimeMillis = loopStart - this.prevTimeMillis;
+            if (_animationTime) {
+                this.deltaTimeMillis = 30;
```
pelikhan: What’s the meaning of 30?

mmoskal: This is what we typically get without this (ie 31 fps or so)

pelikhan:
Do we want this one? @mmoskal

mmoskal:
It didn't really make it deterministic, so I think let's not complicate.

PR #918, opened 2019-08-08
Allow game.askForNumber prompt to use screen's width

alex-kulcsar:
Allow `game.askForNumber` prompt to span the width of the screen, instead of restricting it to the number pad's width.

Before:
before

After:
after


alex-kulcsar:
Commits are unverified; will resubmit shortly.

PR #919, opened 2019-08-08
Allow game.askForNumber prompt to span screen width

alex-kulcsar:
Modify @ChaseMor's commit to allow `game.askForNumber` prompt to span the width of the screen, rather than restrict it to the width of the number pad.

Before:
before

After:
after

### DIFF
--- a/libs/game/numberprompt.ts
+++ b/libs/game/numberprompt.ts
@@ -88,9 +88,15 @@ namespace game {
     //% whenUsed=true
     const INPUT_TOP = NUMPAD_TOP - INPUT_HEIGHT - NUMPAD_INPUT_MARGIN;
 
+    // Pixels kept blank on left and right sides of prompt
+    //% whenUsed=true
+    const PROMPT_MARGIN_HORIZ = 3;
+
     // Dimensions of prompt message area
     //% whenUsed=true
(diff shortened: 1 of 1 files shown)


riknoll:
Thanks Alex!

alex-kulcsar:
Out of curiosity, what does the `whenUsed=true` annotation do?

jwunderl:
whenUsed indicates that it's something that can be dropped if there is nothing directly calling it - basically helping out making the [tree shaking](https://medium.com/@netxm/what-is-tree-shaking-de7c6be5cadd)'s job easier. This is particularly useful for things like [particle effects](https://github.com/microsoft/pxt-common-packages/blob/6ac149ab3ae78f1b382742dfb016ca1485589e96/libs/game/particleeffects.ts#L165), where it's possible that the tree shaking might not be able to recognize that it can get rid of the factories / etc used by the global constant otherwise. (also minor note, the ``=true`` can be dropped as seen in the particle effects example)

I'm not actually certain it will do anything in this context - it seems very possible that the constants would be safely dropped already, but I can't say that for sure without just testing it as I haven't really interacted much with that part of the code. No harm in adding it in this case, though :)

PR #920, opened 2019-08-09
make sprite collisions register events with all tiles touched

jwunderl:
Currently, only the top most / left most wall that a sprite is touching has it's collision event trigger, which feels weird as the sprite is hitting multiple tiles but only counting collisions for one of them seemingly arbitrarily.

This defers the wall collision handling until after each tile the sprite would be touching has been checked, and then runs the collision events for each of them, instead of backing out after the first one.

fixes https://github.com/microsoft/pxt-arcade/issues/1185 (except the sub issue of a request for non-bounding box based tile collisions, which I don't think will be feasible for now at least.)

 (example is slightly modified from the issue this fixes)

Screen Shot 2019-08-08 at 10 26 30 PM

worth a note this will cause a minor change in the block above's behavior (will return the rightmost collided tile instead of the leftmost), but that behavior wasn't defined / generally understood before anyways. I think that one might be good to change in the future, maybe just to be an option on the ``is hitting wall`` block that takes a parameter for the type of wall (e.g so you can check if it's hitting a lava tile); I don't think there's really anything valuable you can do with the block currently (without going to ts), and haven't seen it used (that I remember).


libs/game/sprite.ts
```diff
@@ -732,10 +732,9 @@ class Sprite extends sprites.BaseSprite {
     }
 
     registerObstacle(direction: CollisionDirection, other: sprites.Obstacle) {
-        if (other == undefined) return;
```
riknoll: why this?

jwunderl: Ahh I should have mentioned this one, [``TileMap.getObstacle``](https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/tilemap.ts#L249) can no longer return undefined after the physics interpolation changes, so this is now a redundant check. I just noticed that it while moving stuff around above.

riknoll: Cool! Just checking

PR #922, opened 2019-08-12
remove id field conflicting with base class

jwunderl:
re: https://forum.makecode.com/t/possibly-bogus-compilation-errors-from-beta-beta-makecode-arcade/431; just checked my old branch and I had removed it there but not pushed by accident :(

issue is that duplicate fields with the base class are failing to compile for hardware but succeed in sim; here ParticleSource had the id field redeclared accidentally

### DIFF
--- a/libs/game/particles.ts
+++ b/libs/game/particles.ts
@@ -57,8 +57,6 @@ namespace particles {
          * be culled before a source with a higher priority.
          */
         priority: number;
-
-        id: number;
         _dt: number;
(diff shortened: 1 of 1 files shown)


jwunderl:
Kinda valuable with getters / setters maybe if that's the same code path? but that's buggy right now https://github.com/microsoft/pxt-arcade/issues/1195. I can't think of a reason to non-accidentally have that besides that though.

PR #923, opened 2019-08-12
Settings storage in flash

mmoskal:
Still WIP, but please [take a look at the APIs](https://github.com/microsoft/pxt-common-packages/blob/flash_settings/libs/settings/settings.ts)

I'm not sure we want blocks for these low-level APIs - I think we should rather have high score and possibly save-game APIs.

* [ ] move low-level flash driver to codal
* [x] store volume in settings
* [x] store high score in settings
* [x] implement for Linux
* [x] use program name, not hash as the scope
* [x] panic when second erase within 10s of previous


kevinjwalters:
What storage would be used on an Adafruit Pygamer or Kittenbot Meowbit, for example?

Does this need documentation that describes any longevity issues and an interface that rate limits writes to prevent/slow down tight loops on hardware where premature death of the device from write exhaustion is an issue?

Is there an issue ticket associated with this work?

mmoskal:
There isn't a ticket, feel free to discuss here.

The internal flash is used on both SAMD51 and STM32F4. The flash is filled out sequentially, and then the entire sector is erased, meaning you can do hundreds of writes between erases. However, indeed if someone just writes in a loop, then after about half hour we would reach 10k erase cycles and both MCUs might start failing.

I'm thinking of clearing storage and panic'ing if two erases happen within 10s.

riknoll:
APIs look good! I don't know if this is intended to be used externally, but if it is then it might be good to have set/read APIs for `number[]` in addition to `Buffer`.

mmoskal:
Will add, should be trivial using `msgpack` (we should btw extend `msgpack` with all JSON stuff). `readNumberArray` ?

riknoll:
Sounds good!

mmoskal:
Should be good to go. Tested on SAMD51, STM32F4 and SAMD21. Flash drivers will move to codal at a later date.

Test script: https://makecode.com/_VdYCyr700Dm6

PR #925, opened 2019-08-14
Add Image.equals(otherImage)

jwunderl:
needed to be able to compare images to make new animations be able to continue properly when you play the same one over again

I think https://github.com/microsoft/pxt/pull/5791 might be causing an issue compiling (to meowbit at least, I would assume all hw) now? Had to checkout a commit before to test this on hw, just giving a generic error: ``error: main.ts(1,1): error TS9200: Assertion failed``. Fun fact though, compilation is fast enough that it's actually pretty hard to read output.txt when it gets overwritten; need to stop the auto-re-running after pressing download so that it won't clobber the output I guess.

### DIFF
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -412,8 +412,11 @@ void _mapRect(Image_ img, int xy, int wh, Buffer c) {
     mapRect(img, XX(xy), YY(xy), XX(wh), YY(wh), c);
 }
 
-//%
-bool _equals(Image_ img, Image_ other) {
+//% argsNullable
+bool equals(Image_ img, Image_ other) {
+    if (!other) {
+        return false;
+    }
     auto len = img->length();
     if (len != other->length()) {
         return false;
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -414,20 +414,11 @@ void _mapRect(Image_ img, int xy, int wh, Buffer c) {
 
 //%
 bool _equals(Image_ img, Image_ other) {
-    auto imgData = img->data();
-    auto otherData = other->data();
     auto len = img->length();
-
     if (len != other->length()) {
         return false;
     }
-
-    for (int i = 0; i < len; i++) {
-        if (imgData[i] != otherData[i]) {
-            return false;
-        }
-    }
-    return true;
+    return 0 == memcmp(img->data(), other->data(), len);
 }
 
 /**
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -412,6 +412,24 @@ void _mapRect(Image_ img, int xy, int wh, Buffer c) {
     mapRect(img, XX(xy), YY(xy), XX(wh), YY(wh), c);
 }
 
+//%
+bool _equals(Image_ img, Image_ other) {
+    auto imgData = img->data();
+    auto otherData = other->data();
+    auto len = img->length();
+
+    if (len != other->length()) {
+        return false;
+    }
+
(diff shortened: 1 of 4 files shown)


mmoskal:
Definitely, download shouldn't mean run. You can do ?dbg=1 or PXT_DEBUG=1 in cli to print out stack trace on such "Assertion failed" messages.

How do I repro the failure though? (I didn't try yet)

libs/screen/image.cpp
```diff
@@ -412,6 +412,15 @@ void _mapRect(Image_ img, int xy, int wh, Buffer c) {
     mapRect(img, XX(xy), YY(xy), XX(wh), YY(wh), c);
 }
 
+//%
+bool _equals(Image_ img, Image_ other) {
```
mmoskal: just call it `equals()`, put it in ImageMethods namespace, and skip the helper - it will be added to the interface. The reason _mapRect() is done with helper, is because it has more than 4 args

jwunderl:
It was failing for me with anything i tried with master of all repos, i think I had ``effects.starField.startScreenEffect()`` as one particular case though.

libs/screen/image.cpp
jwunderl: I might have got in my own head a bit on this, I was doing that to have a null / undefined check for the ``other`` param in the ts.  Is there a macro somewhere to check that in cpp, I'm not seeing how to compare to / check against the tagged values right away.

mmoskal: Currently, if you pass null/undefined to a C++ method, it will 980. I'm thinking we should have some annotation to allow null to be passed to C++.

mmoskal: Oh, this only applies for custom reference types. If C++ takes integer it will get `0` for `null`, if it takes a string it will get `"null"`

jwunderl: Ah. So is there a better way to structure this then for now, if I want to just return false on ``validImage.equals(undefined)`` instead of 980? That's what I was trying to oavoid with the helper / ts split, definitely feels a bit wonky / a bit fragile as is though. I guess the annotation would just convert ``null`` / ``undefined`` for custom types to ``nullptr`` so it can be checked in the function itself?

mmoskal: Correct, and correct.

mmoskal: let me see if I can update pxt to support null today

jwunderl:
Hm, I tried this on your branch from https://github.com/microsoft/pxt/pull/5821 and it's failing with 980/989's on my pygamer still; doesn't look like it's being read as nullable as I don't see any of the comments / etc showing up in the assembly. Here's the quick test I'm using;

```typescript
const im1 = img`121`
const im2 = img`121`
const im3 = img`131`
const im4 = img`1212`
const im5 = img`12`

game.splash("true " + im1.equals(im2))
game.splash("false 1" + im1.equals(im3))
game.splash("false 2" + im1.equals(im4))
game.splash("false 3" + im1.equals(im5))
game.splash("nulls" + im1.equals(null)) // << fails 989
game.splash("nulls" + im1.equals(undefined)) // << fails 980
```

@mmoskal Did I put the annotation on the wrong function or something of the sort?

mmoskal:
Shims.d.ts need to be redone in arcade

jwunderl:
@mmoskal this should be good to go now right? It looks like the buffers will still always be the same between two images that are 'equal'  after https://github.com/microsoft/pxt-common-packages/pull/930, just a few more things that will be compared in the header with the memcmp?

mmoskal:
Yup. Good to go.

PR #927, opened 2019-08-16
Move lowTimer on SAMD21 from TC3 to TCC0 to avoid conflict with DAC

mmoskal:
Note that the JACDAC timer on TC4-5 will conflict with PWM on A6, A7.

A2, A3 have PWM on TCC1, so should be fine.

@jamesadevine can jacdac use 16 bit timer? TCC2 or TC5 would be good.


jamesadevine:
I think that should be fine, the stack never waits more than the 160 us in many cases

PR #929, opened 2019-08-16
Move timers to accomodate jacdac, IR, PWM

mmoskal:
now PWM, jacdac, and IR should all work at once

Still, the A6/A7 should be able to do PWM (with TC4), but it seems to mess up TC5; not a regression there though.

### DIFF
--- a/libs/core---samd/platform.cpp
+++ b/libs/core---samd/platform.cpp
@@ -8,9 +8,11 @@ namespace pxt {
 
 #ifdef CODAL_JACDAC_WIRE_SERIAL
 // TC3 is used by DAC on both D21 and D51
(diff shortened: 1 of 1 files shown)


mmoskal:
well, there are these hardware counters/timers and we cannot have two things (among them sound DAC, jacdac, system timer, and PWM for specific pins) use them at once. Ideally we should be smarter about allocating them, but it's generally difficult...

mmoskal:
anyway, after this change A1/A2 seem to work correctly as PWM (eg servoWrite)

mmoskal:
sounds also works, not sure about jacdac, but I don't think it's critical for cpx release

PR #934, opened 2019-08-20
Add Button::pressureLevel()

mmoskal:
This adds `pressureLevel()` method on controller buttons and hardware buttons.

It also adds `control.deviceDalVersion()` on Linux (returning `"linux"`), changes its output from `"0.0.0"` on sim to `"sim"`.

Finally, it adds `controller.player1` to existing `controller.player2,3,4`

Test program:
```typescript
let mySprite2 = sprites.create(img`1 1`)
controller.player1.analog = true
controller.moveSprite(mySprite2, 300, 300)
game.onPaint(function () {
    let x = controller.right.pressureLevel() - controller.left.pressureLevel()
    let y = controller.down.pressureLevel() - controller.up.pressureLevel()
    screen.print(`x:${x} y:${y}`, 0, 0)
})
```


pelikhan:
sim?

libs/game/controller.ts
```diff
@@ -446,53 +470,50 @@ namespace controller {
...
+            } else {
+                svx = (this.right.isPressed() ? 256 : 0) - (this.left.isPressed() ? 256 : 0)
+                svy = (this.down.isPressed() ? 256 : 0) - (this.up.isPressed() ? 256 : 0)
+            }
+
+            let ssvx = svx
+            let ssvy = svy
+
+            const sq = svx * svx + svy * svy
+            const max = 256 * 256
```
riknoll: What is max referring to here? How can sq be greater than max?

riknoll: Also, could we make these variable names a little clearer?

libs/game/controller.ts
```diff
@@ -122,14 +125,29 @@ namespace controller {
...
             return this._pressed;
         }
 
+        /**
+         * Indicates how hard the button is pressed, 0-512
+         */
+        //% weight=99 blockGap=8 help=controller/button/pressure-level
+        //% blockId=key_pressure_level block="%button **button** pressure level"
+        //% group="Single Player"
+        pressureLevel() {
```
riknoll: We should discuss before adding more blocks, especially hardware-dependent ones. Can we remove the block definition for now? This is only useful for pygamer, right?

libs/game/controller.ts
riknoll: Thanks!

libs/game/controller.ts
mmoskal: removed block; it also works on Adafruit Joy Bonnet, and I assume it will also work with the new sim

but I guess we would actually rather have the block for `player1.analog = true`

kevinjwalters:
Is there support in `Blocks` for this? Controller extension?

PR #941, opened 2019-08-22
Add block to make one sprite follow another

jwunderl:
add a block for making a sprite follow another one

~Had a version where turnRate was in degrees, but it slowed things down pretty significantly; this version the turnRate is very much arbitrary, but it has a pretty decent range of behaviors as written between 'orbit around player a bit before hitting' at 0.1 and 'immediately turn towards player' at ~ 5+; not sure that behaviors worth exposing, maybe better to just set it as a constant for now and revisit / add a parameter later. I guess it could also just use an enum (TurnSpeed.VerySlow, ...) and still be able to adjust it later~ dropping turnRate from the block, leaving as a javascript only option and we can adjust it later if we want to give it some more meaningful value

Couple quick examples

### pizza bouncing a pineapple away

https://makecode.com/_JbWix6U3e2du

arcade-screenshot (1)

### duck and soap

https://makecode.com/_PDCiFA6hm1kj

arcade-screenshot (2)

### circling the drain

https://makecode.com/_4vC95zhtYbau

arcade-screenshot (3)

### DIFF
--- a/libs/game/scene.ts
+++ b/libs/game/scene.ts
@@ -33,9 +33,10 @@ namespace scene {
 
     // frame handler priorities
     export const CONTROLLER_PRIORITY = 8;
+    export const UPDATE_CONTROLLER_PRIORITY = 13;
+    export const FOLLOW_SPRITE_PRIORITY = 14;
     export const PHYSICS_PRIORITY = 15;
     export const ANIMATION_UPDATE_PRIORITY = 15;
-    export const UPDATE_CONTROLLER_PRIORITY = 13;
     export const CONTROLLER_SPRITES_PRIORITY = 13;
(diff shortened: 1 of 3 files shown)


pelikhan:
Flocking behavior?

jwunderl:
You mean like [boids](https://www.red3d.com/cwr/boids/)? Could be interesting to add, yeah

pelikhan:
We need a plan to componentize these mini “engines” or we’ll just have tooooo many blocks. We need better discovery of extensions to scale complexity.

jwunderl:
I think following is common enough behavior in games that it might be valid to add in by default personally, but yeah, probably need to review the categories and prefer to move things to packages.

Either way. Would you just think of this as having its own package then?

PR #952, opened 2019-08-28
missing blocks to print text on image

pelikhan:


abchatra:
Why are adding these blocks? I am trying to limit the number of blocks we add.

pelikhan:
We already have blocks for images under "Advanced". We are missing ways to draw text.

jwunderl:
Looks like the blocks are messed up a bit:

Screen Shot 2019-09-05 at 5 37 51 PM

should be under `Images` not `Image` and should have the image they apply to, right?

PR #957, opened 2019-09-04
Add 'play Melody' reference page

ganicke:
Add a page for the new **playMelody()** function.

RE: https://github.com/microsoft/pxt-adafruit/issues/1118, https://github.com/microsoft/pxt-arcade/issues/1276

### DIFF
--- a/libs/mixer/docs/reference/music/play-melody.md
+++ b/libs/mixer/docs/reference/music/play-melody.md
@@ -1,0 +1,38 @@
+# play Melody
+
+Play a short melody of notes composed in a string.
+
+```sig
+music.playMelody("", 120);
+```
+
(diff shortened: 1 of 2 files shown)


libs/mixer/docs/reference/music/play-melody.md
```diff
@@ -0,0 +1,38 @@
...
+
+``"E B C5 A B G A F "``
+
+The melody is shown in the ``||music:play melody||`` block as note symbols which also appear in the Melody Editor.
+
+```block
+music.playMelody("E B C5 A B G A F ", 120);
+```
+
+The medlodies are most often created in the Melody Editor from the block so that valid notes are chosen and the correct melody length is set.
```
jwunderl:
```suggestion
The melodies are most often created in the Melody Editor from the block so that valid notes are chosen and the correct melody length is set.
```

ganicke:
@alex-kulcsar it goes in here. You can suggest some adds/mods if you want.

PR #958, opened 2019-09-05
Adjust block colors to align with WCAG 2.0 Level AA standard

alex-kulcsar:
Tweak colors of blocks to align them with WCAG 2.0 Level AA standards to address accessibility concerns.

Additional details can be found in https://github.com/microsoft/pxt-arcade/pull/1288


jwunderl:
Thanks again!

PR #959, opened 2019-09-05
Changing the shape of animation apis

abchatra:
- Remove loop api and make it a part of single api
- Change the wordings
- Change the path animation default to 2 second
- Rename animation-old.ts to legacy.ts to reflect how the api is categorized. 

**Old:**
image

**New:**
image

**Usage:**
image


abchatra:
Fixes https://github.com/microsoft/pxt-arcade/issues/1280

PR #961, opened 2019-09-09
Revert "make clear particle effects allow sprites in blocks"

jwunderl:
This reverts commit d7f76f0a8f91cead5c0388e73b6a31c62ab613ae from #960.

Seems to be exposing a weird compilation bug when targeting hardware, reverting until that is identified / fixed. Will unfix https://github.com/microsoft/pxt-arcade/issues/1254 for now


riknoll:
This seems to work now... I think the compiler bug may have been fixed

PR #973, opened 2019-09-19
Esp32 cleanup

pelikhan:
Cleaning up logging, updated sample.


libs/esp32/atcontroller.ts
```diff
@@ -0,0 +1,132 @@
...
+            let errorCode: number = 0;
+            let line = "";
+            const lines: string[] = [];
+            do {
+                line = this.ser.readLine();
+                if (line == "OK")
+                    status = ATStatus.Ok;
+                else if (line == "ERROR")
+                    status = ATStatus.Error;
+                else if (line.substr(0, "ERR CODE:".length) == "ERR CODE:")
```
mmoskal: we should really add `String.startsWith()`...

sabas1080:
why not add support to BLE in NINA firmware? https://github.com/arduino-libraries/ArduinoBLE

libs/esp32/atcontroller.ts
```diff
@@ -0,0 +1,132 @@
+namespace esp32 {
```
pelikhan: ignored for now.

PR #975, opened 2019-09-25
Add pin for USB power detection

mmoskal:
Also adds pins from #973 to avoid collisions


pelikhan:
patcher needs to be updated

mmoskal:
Already updated

PR #976, opened 2019-09-26
Don't trigger 'any' events twice

mmoskal:


mmoskal:
fixes https://github.com/microsoft/pxt-arcade/issues/1126

PR #979, opened 2019-09-30
Update 'animation' ref pages

ganicke:
Add the pages for the new style animation APIs.

RE: https://github.com/microsoft/pxt-arcade/issues/1350

Notes:

* **stopAnimation()** seems to have trouble decompiling back into blocks

stop-animation-decomp

*  why is ``sprite`` set as the second param in **stopAnimation()**?

### DIFF
--- a/libs/animation/docs/reference/animation/run-movement-animation.md
+++ b/libs/animation/docs/reference/animation/run-movement-animation.md
@@ -1,0 +1,78 @@
+# run Movement Animation
+
+Apply a movement animation effect to a sprite.
+
+```sig
+animation.runMovementAnimation(null, "")
+```
+
+A movement animation is motion effect applied to a sprite. The image in the
+sprite will appear to move with the type of motion that you select. You have several
+movement types to choose from and you can select them from the list in the block.
+
+The animation runs for the amount of time you set for it. When the animation finishes, the sprite returns to its original position. You can make the animation repeat by setting the **loop** parameter to **ON**.
+
+## Parameters
+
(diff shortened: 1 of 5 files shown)


libs/animation/docs/reference/animation/run-movement-animation.md
```diff
@@ -0,0 +1,78 @@
...
+
+```sig
+animation.runMovementAnimation(null, "")
+```
+
+A movement animation is motion effect applied to a sprite. The image in the
+sprite will appear to move with the type of motion that you select. You have several
+movement types to choose from and you can select them from the list in the block.
+
+The animation runs for the amount of time you set for it. When the animation finishes, the sprite returns to its original position. You can make the animation repeat by setting the **loop** parameter to **ON**.
```
dannytech: The sprite will only return to its original position if the animation is in looping mode. This may not actually be the desired behavior, to be honest. Some animations, such as wave, were designed so in looping mode it would start from the ending position of the previous animation. However, this does not occur because the path is parsed into a set of absolute values at the instantiation of a MovementAnimation, and are not parsed again into a new set when looping. It might make sense to redesign this behavior, either caching the path string and simply parsing again with a different starting point, or implementing a `shift` function that simply takes a vector of the difference between start and end and generates a new path with all the points shifted.

ganicke: Thanks Danny. It seems that "loop" now will both return the sprite to its original location and restart the animation too.

PR #982, opened 2019-10-02
reorg core---linux to use in EV3

mmoskal:
Also cleans up serial number story


libs/core---linux/linux.cpp
```diff
@@ -102,12 +102,16 @@ Event *mkEvent(int source, int value) {
 volatile bool paniced;
 extern "C" void drawPanic(int code);
 
+void tryLockUser() {
+    pthread_mutex_trylock(&execMutex);
```
mmoskal: probably makes sense to return value from trylock (to see if it succeeded)

PR #986, opened 2019-10-08
use identifier for A/B buttons

pelikhan:
A,B already used in melodies. Leads to confusing translations.

### DIFF
--- a/libs/game/controller.ts
+++ b/libs/game/controller.ts
@@ -9,9 +9,9 @@ enum ControllerButtonEvent {
 
 
 enum ControllerButton {
-    //% block="A"
+    //% block="{id:controller}A"
     A = 5,
-    //% block="B"
+    //% block="{id:controller}B"
     B = 6,
     //% block="left"
     Left = 1,


pelikhan:
Example: https://github.com/microsoft/pxt/blob/6dd7417bfd645d3714c4da81da89893a2c6af013/webapp/src/blocksSnippets.ts#L105

ganicke:
Is something else needed to make this work correctly? -

image

riknoll:
Yeah @pelikhan this is how it shows up in beta. We need a fix...

pelikhan:
It cannot be “A” because “A” is already translated to “La” in French. Somehow this {id:...} should have been eaten by lf

riknoll:
Maybe it doesn't get eaten for English?

jwunderl:
Worth noting single player `A` is getting the flag removed properly; maybe it's only being removed for the `this` ref on methods currently and not other params? The button is being passed in as a param to the multiplayer `onEvent`

pelikhan:
Might be a code path that needs a rlf call.

jwunderl:
Oh.. This uses and enum and the other one's a fixed instance, the normal ``on a pressed`` just didn't get changed: https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/controlleroverrides.ts#L2

Looks like we don't use rlf directly when translating blocks, which is why these didn't cleared: https://github.com/microsoft/pxt/pull/6303

PR #990, opened 2019-10-10
moving to execption in esp32 code

pelikhan:
- [x] don't crash unless config issue.
- [x] top level friendly function like request.get swallow exceptions.


ladyada:
fyi we just updated the esp32 firmware we're using to support client-side ssl cert/key auth
https://github.com/adafruit/nina-fw/releases/tag/1.4.0

PR #991, opened 2019-10-17
fix some projectile + tilemap interactions

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1393
fixes issue reported on discord: https://makecode.com/_bhqJj74qjY0J has sprites moving only vertically, but they should have vx too (they were hitting the bounds of the tmap)
fixes https://github.com/microsoft/pxt-arcade/issues/1104
fixes https://github.com/microsoft/pxt-arcade/issues/1323

also fixes a few minor things in projectiles / sprites
* actually use the velocity that is chosen when randomly picking
* only set position once, to get rid of some unnecessary tilemap interactions
* defer setting position to ~sprites.create~ and Tile.place (same reason)

If a sprite is created or is moved more than a single step away, and the new position for the sprite is a wall, this disables tilemap interactions with that sprite (clipping and preventing movement) until the sprite 'unclips' itself by moving out of the way.

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -165,8 +165,8 @@ class Sprite extends sprites.BaseSprite {
     constructor(img: Image) {
         super(scene.SPRITE_Z);
 
-        this._x = Fx8(screen.width - img.width >> 1);
-        this._y = Fx8(screen.height - img.height >> 1);
+        this._x = Fx.zeroFx8;
+        this._y = Fx.zeroFx8;
         this._lastX = this._x;
         this._lastY = this._y;
         this.vx = 0
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -165,8 +165,8 @@ class Sprite extends sprites.BaseSprite {
     constructor(img: Image) {
         super(scene.SPRITE_Z);
 
-        this._x = Fx.zeroFx8;
-        this._y = Fx.zeroFx8;
+        this._x = Fx8(screen.width - img.width >> 1);
+        this._y = Fx8(screen.height - img.height >> 1);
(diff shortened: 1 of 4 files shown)


libs/game/sprite.ts
```diff
@@ -165,8 +165,8 @@ class Sprite extends sprites.BaseSprite {
     constructor(img: Image) {
         super(scene.SPRITE_Z);
 
-        this._x = Fx8(screen.width - img.width >> 1);
-        this._y = Fx8(screen.height - img.height >> 1);
+        this._x = Fx.zeroFx8;
```
pelikhan: this is a breaking change

riknoll: Not sure there is a good reason to break this; I realize it probably won't affect many people but why change it?

libs/game/tilemap.ts
```diff
@@ -257,5 +255,24 @@ namespace tiles {
                 index
             );
         }
+
+        public isOnWall(s: Sprite) {
+            const hbox = s._hitbox
+
+            const left = Fx.toIntShifted(Fx.add(hbox.left, Fx.oneFx8), this.scale);
```
riknoll: Why do you add 1 to the left and top?

jwunderl: Ahh, that was to fix an off by one bug, but I ended up fixing the issue causing it / rendering the offset moot. Removed~

libs/game/sprite.ts
jwunderl: Was just moving the initialization to the factories / `sprites.create` / `sprites.createProjectile` to keep them consistent there; good point though, no actual reason to break in case some person has extended this. Will put this back

PR #994, opened 2019-10-18
Keyboard buffered

pelikhan:
Keeps track of the down state of a key so that we don't span the computer with incorrect up/down events.

### DIFF
--- a/libs/keyboard/keyboard.ts
+++ b/libs/keyboard/keyboard.ts
@@ -8,7 +8,11 @@ namespace keyboard {
         private _keysDown: number[];
         private _mediasDown: number[];
         private _functionsDown: number[];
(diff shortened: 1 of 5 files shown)


libs/keyboard/keyboard.ts
```diff
@@ -3,5 +3,154 @@
...
+                    }
+            }
+
+            return false;
+        }
+
+        clear() {
+            // send up command for all down keys and clear
+            if (this._keysDown) {
+                this._keysDown.forEach(c => __key(c, KeyboardKeyEvent.Up));
```
mmoskal: How does for each interact with modifying the array?

pelikhan: using pop now

PR #997, opened 2019-10-21
add 'see through' scene flag

jwunderl:
Make it so you can set a scene as 'see through' - if set, it will render the scene below it as the background instead of using a normal background.

This lets you create scenes 'on top' of other scenes (menus, splash screens, other temporary overlays like tutorials) without explicitly managing a snapshot of the image from the previous scene as the background image.

### DIFF
--- a/libs/game/scene.ts
+++ b/libs/game/scene.ts
@@ -197,7 +197,7 @@ namespace scene {
             this.flags |= scene.Flag.IsRendering;
 
             control.enablePerfCounter("render background")
-            if (this.flags & scene.Flag.SeeThrough && this.previousScene) {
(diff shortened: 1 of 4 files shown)


libs/game/scene.ts
```diff
@@ -189,34 +188,33 @@ namespace scene {
...
-        }
+        render() {
+            // bail out from recursive or parallel call.
+            if (this.flags & scene.Flag.IsRendering) return;
+            this.flags |= scene.Flag.IsRendering;
 
-        private renderCore() {
             control.enablePerfCounter("render background")
-            this.background.draw();
+            if (this.flags & scene.Flag.SeeThrough && this.previousScene) {
```
pelikhan: use parenthesis

PR #1000, opened 2019-10-24
make sprite.follow(sprite) follow more accurately

jwunderl:
and make the turnRate variable meaningful / scale properly

Now the default behavior will handle momentum a bit more accurately; from a dead stop (with turnRate = 100) it will hit max speed after .5 seconds, if you double the turnRate (200) it will take .25 seconds, etc.

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -784,7 +784,9 @@ class Sprite extends sprites.BaseSprite {
      *
      * @param target the sprite this one should follow
      * @param speed the rate at which this sprite should move, eg: 100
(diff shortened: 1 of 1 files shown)


riknoll:
I think this is a breaking change, since we shipped the old behavior. @abchatra thoughts?

abchatra:
breaking how? We just released follow like last month.

riknoll:
The second parameter is now scaled differently. It used to have a default of 3 and now the default is 100. right @jwunderl ?

jwunderl:
Yup, I thought we had left that parameter off the block initially partially to make it so we could play with the scaling later though

PR #1013, opened 2019-11-21
Adding a camera sprite flag and getters for the camera position

riknoll:
Adding two new blocks for getting camera left/top and a new sprite flag for HUD elements.

This is in response to https://forum.makecode.com/t/overlay-help/807/3 and https://forum.makecode.com/t/hud-and-inventory-for-large-tile-maps/644


riknoll:
Good point @jwunderl, I've updated it to act like ghost and added documentation

PR #1024, opened 2019-12-09
Tilemap block definitions

shakao:
image

TODO: 
- Go over all examples and ensure they're correct
- Fix bug with using 'location' argument in sprite overlaps tile block

### DIFF
--- a/libs/game/docs/reference/sprites/sprite/tile-kind-at.md
+++ b/libs/game/docs/reference/sprites/sprite/tile-kind-at.md
@@ -1,0 +1,79 @@
+# tile Kind At
+
+Check if the type of tile next to the sprite is of a specific kind
+
+```sig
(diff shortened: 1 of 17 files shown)


riknoll:
Would `array of all _____ tile locations` be clearer? Or is that too long?

shakao:
updated blocks:
image

pelikhan:
I think "tile map" should be used consistently across all blocks. Map alone has many meanings.

shakao:
yeah good point, better to be consistent
image

riknoll:
Hmmm..... "tile map" or "tilemap"?

riknoll:
I thought we were doing the tileset field editor as a shadow block?

pelikhan:
also use ``blockGap`` to harmonize the gap

PR #1025, opened 2019-12-09
Pull legacy 'music' docs into 'mixer'

ganicke:
Bring over the older `music` reference pages to match corresponding APIs in `mixer`.

Fixes https://github.com/microsoft/pxt-arcade/issues/1460

### DIFF
--- a/libs/mixer/docs/reference/music/composing-sounds.md
+++ b/libs/mixer/docs/reference/music/composing-sounds.md
@@ -1,0 +1,71 @@
+# Composing sounds
+
+Composing some sound, or maybe some music, is done by putting tones to together, one after another.
+
+## Musical notes
+
+A _note_ is a tone that is recognized as part of music. A note has a name like '**C**'. A note is
+played for an amount of time called its _duration_. 
+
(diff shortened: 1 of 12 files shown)


libs/mixer/docs/reference/music/composing-sounds.md
```diff
@@ -0,0 +1,71 @@
...
+"E3:3 R:1 D#:3 R:1 D:4 R:1 C#:8"
+```
+
+What you see is not some alien language but a bunch of notes with their duration. The form of a single
+note is **note : duration** or ``C:2``. This means play the '**C**' note for **2** beats of time.
+The notes are placed one after the other with a _space_ between them, like ``"B:2 C#:6"``. If you want
+a note to play for **4** beats, you don't need to use any duration number (a note with 4 beats is called a _whole_ note).
+Just say something like ``"E"`` with no colon (leave out the ``':'``) and no duration number.
+
+You might notice that the sound string has an ``R:1`` in it. The '**R**` means _rest_ and to rest for one beat.
```
jwunderl:
```suggestion
You might notice that the sound string has an ``R:1`` in it. The `**R**` means _rest_ and to rest for one beat.
```

PR #1026, opened 2019-12-10
按钮响应问题

zidong0822:


libs/buttons/buttons.cpp
```diff
@@ -212,9 +212,10 @@ PressureButton *getButtonByPin(int pin, int flags) {
         }
 
         auto pull = PullMode::None;
-        if ((flags & 0xf0) == 0x10)
+        //硬件按钮从低电平改为高电平后，需要修改此处的条件判断。
```
pelikhan: could you explain this change?

mmoskal:
Generally, to configure BTN_A as pin 7 use: `BTN_A = 0x0007` (which is equivalent to `BTN_A = 0x210007`) when the button is connected to GND and use `BTN_A = 0x110007` when button is connected to 3V, see https://arcade.makecode.com/hardware/adding#cf2

PR #1027, opened 2019-12-11
fix category recoloring issue by fixing dependency from game to "device" pkg instead

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/1489
along with this PR: https://github.com/microsoft/pxt-arcade/pull/1528

### DIFF
--- a/libs/animation/pxt.json
+++ b/libs/animation/pxt.json
@@ -12,6 +12,6 @@
     ],
     "public": true,
     "dependencies": {
-        "game": "file:../game"
+        "device": "file:../device"
     }
 }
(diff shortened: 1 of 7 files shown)


riknoll:
I think this should happen in pxt-arcade, not common-packages. You can override the pxt.json in the libs folder over there. I guess alternatively we could just make the arcade colors the default colors but that might just be punting the issue. As this PR stands, I think it will break pxt-brainpad because it does not have a libs/device

darzu:
Shoot this is failing to build because it doesn't know about "device".

It's not great to have a dependency from pxt-common-packages on pxt-arcade anyway. Not sure what to do about this yet.

pelikhan:
Why is this happening?

PR #1029, opened 2019-12-11
fix arcade colors in common packages

darzu:
Use the colors from pxt-arcade.

Fixes: https://github.com/microsoft/pxt-arcade/issues/1489
Sister PR: https://github.com/microsoft/pxt/pull/6320


darzu:
After:
image

PR #1034, opened 2019-12-12
Disable screen brightness setting on Kitronik

mmoskal:
This makes the API no-op and removes buttons from the menu.

This also removes buttons from the menu in simulator. Any objections to that?


mmoskal:
Oh, the backstory is that Kitronik board has backlight on PA06 which doesn't support PWM

PR #1038, opened 2019-12-13
Trying out software cap-sense

mmoskal:


libs/core/pinsDigital.cpp
```diff
@@ -101,6 +101,19 @@ void onEvent(DigitalInOutPin pin, PinEvent event, Action body) {
...
+/**
+ * Measure pin capacitance.
+ */
+//%
+int readCapacitance(DigitalInOutPin pin) {
+    // on STM32F4 this requires 2M pulldown
+    // other chips to be tested
+    pin->setDigitalValue(1);
+    pin->getDigitalValue(PullMode::None);
+    sleep_us(50);
```
pelikhan, line 113: Should this be a bootloader variable?

mmoskal: probably, let me actually see if I can do this in TS

mmoskal:
Can be done in TS, probably need to drop this PR.

```typescript
let v = 0
let t = 0

function cap(p:PwmPin) {
    p.digitalWrite(true)
    p.setPull(PinPullMode.PullNone)
    p.digitalRead()
    control.waitMicros(50)
    return p.analogRead()
}

game.onPaint(() => {
    let n= control.millis()
    if (n-t>400) {
        t=n
        v=cap(pins.P0)
    }
    screen.print("v:" + v, 3, 3, 3)
})
```

PR #1039, opened 2019-12-13
Adding decompiler annotation to tile arguments

riknoll:
The annotation we always forget to add to all image arguments. This is part of the fix for https://github.com/microsoft/pxt-arcade/issues/1575


libs/game/tilemap.ts
```diff
@@ -468,6 +468,7 @@ namespace tiles {
      */
     //% blockId=mapsettileat block="set %loc=mapgettile to %tile"
     //% tile.shadow=tileset_tile_picker
+    //% tile.decompileIndirectFixedInstances=true
```
darzu: Side note: do we document all these flags anywhere?

PR #1040, opened 2019-12-16
Add extension for color-coded (legacy) tilemap, add getTileAt API in …

shakao:
…javascript

Fixes https://github.com/microsoft/pxt-arcade/issues/1576
Fixes https://github.com/microsoft/pxt-arcade/issues/1584


libs/game/docs/reference/tiles/get-tile.md
```diff
@@ -1,9 +1,9 @@
-# get Tile
+# get Tile Location
```
jwunderl: Name of this file should probably change to match (looks like it was already intended to from the links below)

libs/game/docs/reference/tiles/get-tile-location.md
```diff
@@ -1,9 +1,9 @@
-# get Tile
+# get Tile Location
 
 Get a tile location from the tilemap.
 
 ```sig
-tiles.getTile(0, 0)
+tiles.getTileLocation(0, 0)
```
pelikhan: since we are making this a legacy api, why are we renaming this?

libs/color-coded-tilemap/README.md
```diff
@@ -0,0 +1,3 @@
+# LEGACY Tilemap
```
pelikhan: color coded tilemap; don't call it legacy until we fix the course.

pelikhan: "Blocks for the color-coded tilemap"

libs/color-coded-tilemap/tilemap.ts
```diff
@@ -0,0 +1,160 @@
+//% blockGap=8
+namespace scene {
+    /**
+     *  LEGACY: package for old tilemap blocks, to support existing curriculum.
```
pelikhan: Color coded tilemap

pelikhan: is this comment slurped into the jsdoc of setTileMap?

pelikhan:
I think we should keep the existing docs for the color-coded-tilemap since the blocs might be used and the user might try to get help for them.

shakao:
removed "legacy" from descriptions, and added the package to existing docs! user will be able to right-click -> Help and see the same docs as before

PR #1044, opened 2019-12-19
Make buffer writable for tilemap data

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1625

Obviously this has memory tradeoffs. I was hoping to only make it writable if the user calls the set tile API, but I can't see a way to check it in the TS. @mmoskal is this okay do you think? Tilemaps are one byte per pixel.

### DIFF
--- a/libs/game/tilemap.ts
+++ b/libs/game/tilemap.ts
@@ -155,6 +155,10 @@ namespace tiles {
         setTile(col: number, row: number, tile: number) {
             if (this.isOutsideMap(col, row)) return;
 
+            if (this.data.isReadOnly()) {
+                this.data = this.data.slice();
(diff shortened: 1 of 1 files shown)


mmoskal:
It would be really easy to expose the isReadOnly method on buffer - it already exists in C++. Also `const b = data.slice()` should work.

Actually, gimme sec.

mmoskal:
Added method, unfortunately needs to also go in pxt-core for sim.

Maybe it's not worth it for this release, I don't know...

riknoll:
I'll port it over to stable tomorrow. Thanks!

riknoll:
Updated, now with the new API!

PR #1057, opened 2020-01-08
Include new codes is error codes page

ganicke:
Should there be a note for `000`?

RE: https://github.com/microsoft/pxt-arcade/issues/1688, https://github.com/microsoft/pxt-arcade/issues/1474


abchatra:
>> Should there be a note for 000?

I think its alright, user-generated panic code can be any number.

PR #1058, opened 2020-01-09
Migrate radio to common packages

pelikhan:
Moving radio code from microbit to common packages.
Requires https://github.com/microsoft/pxt/pull/6431

Note that I have left deprecated APIs in microbit for backward compat. 
There will probably be some C++ update needed to make this compile with CODAL (untested)


pelikhan:
@mmoskal ready for review. mostly moving files around. Might need to update C++ defines to make it build on arcade.

pelikhan:
We probably need to do the radio—-platform trick to handle MCU that don’t have radio

mmoskal:
Why? You just should not be able to add the radio package on STM

mmoskal:
You could have radio—-dongle though, connected via SPI or maybe jacdac

pelikhan:
I mean we have RPI, SAMD, STM....

PR #1059, opened 2020-01-14
Particle effects and say sprites should respect relativeToCamera

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1700

This does not update particle effects after they are created because they don't keep a reference to sprites and it would have been too big a refactor. I think it's a pretty niche scenario anyways.

### DIFF
--- a/libs/game/particleeffects.ts
+++ b/libs/game/particleeffects.ts
@@ -22,11 +22,12 @@ namespace effects {
          * Attaches a new particle animation to the sprite or anchor for a short period of time
          * @param anchor
          * @param duration
-         * @param particlesPerSecond 
+         * @param particlesPerSecond
          */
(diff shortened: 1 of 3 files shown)


kevinjwalters:
What is the animated gif video demonstrating here, I don't follow what the problem is?

riknoll:
There is a sprite flag called "RelativeToCamera" that causes a sprite to be drawn relative to the camera instead of relative to the world. It's useful for making sprites that stick to the screen (e.g. health bars, informational text, etc.)

The bug was that if a sprite has that flag set, it should also set that flag on any particles or message bubbles that it creates. In the GIF above, the camera is centered on the duck in the center. The monkey is relative to the camera, so it shouldn't move with the duck. The other duck sprite is not relative to the camera, so it should move when the duck moves.

PR #1062, opened 2020-01-17
make sprite.say take any

jwunderl:
make sprite say take `any`. `sprite.say` doesn't play too nicely with `\n`'s at the moment; I just worked around it for now by replacing them with spaces, but I think supporting multi line `say`s  nicely would be a nice feature? Can see how much effort that would be if we want, but I'm not sure how that would end  up looking -- conflicts with the basic scrolling.

Noticed that we probably should add string.replace and string.replaceAll (non regex versions of course) to the string 'prototype' in pxtlib

(Also, took an extra few minutes to debug what was wrong with the first commit; pretty subtle error that feels very hard to catch)

### DIFF
--- a/libs/base/console.ts
+++ b/libs/base/console.ts
@@ -26,7 +26,8 @@ namespace console {
         function (priority: ConsolePriority, text: string) { control.__log(priority, text); }
     ];
 
-    export function add(priority: ConsolePriority, text: string) {
+    export function add(priority: ConsolePriority, message: any) {
+        let text = inspect(message);
         if (priority < minPriority) return;
         // add new line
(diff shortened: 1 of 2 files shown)


jwunderl:
just noticed a weird error popping up between javascript=>blocks when passing in non-string objs that doesn't trigger with `player.say` in minecraft, so it must be params of functions vs params of methods or something of the sort? Taking a look to see where it breaks / what's going wrong.

jwunderl:
Looks like that was actually just an issue with localhost caching / ``pxt serve --rebundle``; it wasn't updating the api info. Works:

PR #1066, opened 2020-01-27
allow resetting countdown in 'on countdown end'

jwunderl:
reset the gameEnd / timer for countdown before invoking the game over / countdown end handler, so that user can choose to reset timer in their handlers (e.g. a 'need more time' screen that pops back up with a few more ms)


stereoa:
I still see this bug in the latest release.

PR #1070, opened 2020-02-12
Add support for JACDAC display

mmoskal:
@jamesadevine would you mind taking a look at jdarcade.h and jdprotocol.h - keeping these in public pxt-common-packages for now


jamesadevine:
Hey @mmoskal,

I'll just provide an overview on what I think you've done:

* Implement JACDAC over SPI
* Implement a software display service that abstracts away hardware implementation specifics

The display service has 4 commands:

* set_window(uint16_t x, uint16_t y, uint16_t width, uint16_t  height);
* set_palette(uint32_t* palette);
* set_pixels(uint32_t* pixels, int len);
* set_brightness(uint16_t level);

The SPI JD packet looks like this:
```
typedef struct {
     uint8_t size; // of the payload
     uint8_t flags;
     uint8_t service_number;
     uint8_t service_command;
     uint8_t data[0];
 } __attribute((__packed__)) __attribute__((aligned(4))) jd_packet_t;
```
No CRC makes sense for spi, and there is also no need to send device identifiers as SPI is direct communication between two MCUs. Is the small packet size not limiting when sending display buffers?

I had a glance over the implementation and I didn't see anything that made me go "huh?" :smile:

mmoskal:
this was merged in a different PR

PR #1082, opened 2020-03-18
add location to wall hit at location, and tile at location is block

jwunderl:
did this real quick while watching the arcade stream today

add location to wall hit handler, and 'tile at location is' block; closes https://github.com/microsoft/pxt-arcade/issues/1783

will require https://github.com/microsoft/pxt-arcade/pull/2075 (which would need to be updated to apply to the correct versions)

image

(needs a bit of consensus as this adds a new block)


libs/game/obstacle.ts
```diff
@@ -35,11 +35,11 @@ namespace sprites {
         }
 
         get x(): number {
-            return this.left + this.width >> 1;
+            return this.left + (this.width >> 1);
```
jwunderl: shows how much the old on hit tile was used; this has been giving the location incorrectly since tiles were first implemented -- ``a + b >> c`` === ``(a + b) >> c``

libs/game/tilemap.ts
```diff
@@ -501,6 +501,21 @@ namespace tiles {
...
+    /**
+     * Returns true if the tile at the given location is the same as the given tile;
+     * otherwise returns false
+     * @param location
+     * @param tile
+     */
+    //% blockId=maplocationistile block="tile at $location is $tile"
+    //% location.shadow=mapgettile
+    //% tile.shadow=tileset_tile_picker tile.decompileIndirectFixedInstances=true
+    //% blockNamespace="scene" group="Tiles" blockGap=8
```
abchatra, line 513: Should this group be "collisions"  instead of "Tiles"?

abchatra:
Can we merge this change if it ready. I am ok with the changes.

riknoll:
is this not breaking?

abchatra:
Oh I didn't realize that. How is this breaking? Can we add upgrade rules?

libs/game/tilemap.ts
jwunderl: I don't think this block belongs in collisions, as this is just checking a location / does not involve a sprite 'hitting' a spot like the collisions do. Maybe the overlaps / hits wall blocks should, but I think we had decided on them being at the top of the tiles category so that they're more obvious / visible

jwunderl:
Yes, when you import an old project it doesn't add in the shadow yet

image

I don't think we currently have any of the automatic upgrade rules for handling this, but just need to add the shadow block reporter in the upgrade rule which should be easy.

On the typescript side should be fine, anyone who was using it before is now just ignoring one of the arguments, like if they were to use `[1].forEach(v => console.log(v))` they'd be ignoring the second index argument

libs/game/tilemap.ts
abchatra: image

We have above in the collisions?

jwunderl: I suppose, I think of that one as `tile sprite is hitting on the [left]` which makes sense under the collision category.

Maybe we should rename Collisions to something else, the split between tiles and it feels fairly arbitrary to me

jwunderl:
Okay, i think this is the upgrade rule that would be needed: https://github.com/microsoft/pxt-arcade/pull/2075

libs/game/tilemap.ts
```diff
@@ -509,6 +509,21 @@ namespace tiles {
...
+     * @param location
+     * @param tile
+     */
+    //% blockId=maplocationistile block="tile at $location is $tile"
+    //% location.shadow=mapgettile
+    //% tile.shadow=tileset_tile_picker tile.decompileIndirectFixedInstances=true
+    //% blockNamespace="scene" group="Collisions" blockGap=8
+    export function tileAtLocationEquals(location: Location, tile: Image) {
+        const scene = game.currentScene();
+        return location && tile && location.tileSet === scene.tileMap.getImageType(tile);
```
livcheerful, line 524: would it be a situation where this block can get called and there's no scene.tileMap?

jwunderl: Good catch, should match the behavior of others and ignore invalid calls; fixed ~

PR #1086, opened 2020-04-02
Fix crash in hw profiling

mmoskal:
Also disable auto-dumping of counters on hardware (clogs the buffer)
and minor perf fix in dialogs

### DIFF
--- a/libs/base/eventcontext.ts
+++ b/libs/base/eventcontext.ts
@@ -86,7 +86,7 @@ namespace control {
                 EventContext.lastStats = `fps:${Math.round(fps)}`;
                 if (fps < 99)
                     EventContext.lastStats += "." + (Math.round(fps * 10) % 10)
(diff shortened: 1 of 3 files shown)


libs/base/eventcontext.ts
```diff
@@ -86,7 +86,7 @@ namespace control {
                 EventContext.lastStats = `fps:${Math.round(fps)}`;
                 if (fps < 99)
                     EventContext.lastStats += "." + (Math.round(fps * 10) % 10)
-                if (control.profilingEnabled()) {
+                if (control.ramSize() > 2000000 && control.profilingEnabled()) {
```
pelikhan, line 89: should 2000000 be pulled from a bootloader constant?

mmoskal: nah, if you have over 2M of memory, the dmesg is probably going to be big enough

PR #1089, opened 2020-04-08
align c++ parsefloat with js

jwunderl:
along with https://github.com/microsoft/pxt/pull/6825 fix https://github.com/microsoft/pxt-microbit/issues/2664

fix some parsefloat mismatches with js; allow any whitespace chars (not just spaces), return value up to given point if values after first digit are non-numeric (`1f` => 1, `f1` => NaN), handle trailing e that is not followed by a number (`1ef` => 1), don't ignore spaces after first digit (`123 456` => 123). Also drop auto conversion to hex for parsefloat (js doesn't) (this makes the change require https://github.com/microsoft/pxt/pull/6825 or equivalent to keep hex parsing in parseInt)

isspace was used in a decent number of other places so I left it as is and added one to check if character is whitespace; can check if the other places should actually be checking for any whitespace

### DIFF
--- a/libs/base/core.cpp
+++ b/libs/base/core.cpp
@@ -561,7 +561,7 @@ int length(String s) {
 }
 
 #define isspace(c) ((c) == ' ')
-#define iswhitespace(c) ((c) == 0x0009 || (c) == 0x000B || (c) == 0x000C || (c) == 0x0020 || (c) == 0x00A0 || (c) == 0xFEFF || (c) == 0x000A || (c) == 0x000D || (c) == 0x2028 || (c) == 0x2029)
+#define iswhitespace(c) ((c) == 0x09 || (c) == 0x0B || (c) == 0x0C || (c) == 0x20 || (c) == 0xA0 || (c) == 0x0A || (c) == 0x0D)
 
 NUMBER mystrtod(const char *p, char **endp) {
     while (iswhitespace(*p))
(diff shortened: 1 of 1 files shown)


libs/base/core.cpp
```diff
@@ -561,49 +561,50 @@ int length(String s) {
 }
 
 #define isspace(c) ((c) == ' ')
+#define iswhitespace(c) ((c) == 0x0009 || (c) == 0x000B || (c) == 0x000C || (c) == 0x0020 || (c) == 0x00A0 || (c) == 0xFEFF || (c) == 0x000A || (c) == 0x000D || (c) == 0x2028 || (c) == 0x2029)
```
mmoskal, line 564: c is 8 bit, so characters above 0xff won't ever match; if you want to support unicode it's a bit more tricky

libs/base/core.cpp
```diff
@@ -561,49 +561,50 @@ int length(String s) {
...
     }
 
     v *= m;
 
-    if (*p) {
+    if (*p == 'e' || *p == 'E') {
         p++;
         int pw = (int)strtol(p, endp, 10);
-        v *= p10(pw);
+        if (!isnan(pw)) {
```
mmoskal, line 603: int can never be nan, it will return 0 if no number is found, and keep `*endp == p`; in other words, I think you can skip the entire if statement

libs/base/core.cpp
jwunderl: ahh forgot about that, so that's line separator, paragraph separator and zero width non breaking space; I doubt anyone will end up using those ever with parsefloat so will just drop for now

PR #1094, opened 2020-04-13
Jacdac service updates

mmoskal:
JD service updates:
* add range limit to crank
* servo now only takes pulse length
* add arcade controls client
* add `jacdac.autoBind()` to bind services (randomly if need be)
* add JD packet forwarding to HF2
* add `jacdac.onRawPacket()`
* add register query infrastructure for Ctrl service (may need to move to general Client class)

Non-JD changes:
* add Buffer.toArray()


libs/jacdac-services/servoclient.ts
```diff
@@ -17,11 +36,12 @@ namespace jacdac {
         //% servo.fieldOptions.columns=2
         //% blockGap=8        
         setAngle(degrees: number) {
-            if (!this.state[0] || this.state.getNumber(NumberFormat.Int16LE, 1) != degrees) {
-                this.state[0] = 1;
-                this.state.setNumber(NumberFormat.Int16LE, 1, degrees);
-                this.notifyChange();
-            }
+            // this isn't exactly what the internets say, but it's what codal does
+            const center = 1500
```
pelikhan, line 40: it's not uncommon that these values need to be tuned based on the drift of the servo.

mmoskal: I think I'll change the service API to have these values as registers, and the value be the angle (probably 0-0xffff); then it can come pre-configured by the jd-servo manufacturer but be changed as needed

libs/base/buffer.ts
```diff
@@ -183,6 +183,15 @@ namespace helpers {
             return r
         }
     }
+
+    export function bufferToArray(buf: Buffer, format: NumberFormat) {
```
pelikhan, line 187: would be good to have in microbit

libs/jacdac-services/servoclient.ts
mmoskal: If there is drift over usage, then I think compensation will have to be stored on the brain, not in the peripheral. Peripheral will be stateless.

PR #1097, opened 2020-04-15
Onboard pins cfg keys + keypad

pelikhan:
Add new keys to specifically configure the on-board APA or Neopixels. This is useful when a board (like neotrellis) has **both** a dotstar and a neopixel strip.
Add matrix-keypad support
Also refreshing dal.d.ts

The keypad is currently polling - we could probably use some pin events but that can be done in the future. Working on neotrellis.


libs/base/configkeys.h
```diff
@@ -212,6 +212,31 @@
 #define CFG_ANALOG_JOYSTICK_MIN 216
 #define CFG_ANALOG_JOYSTICK_MAX 217
 #define CFG_TIMERS_TO_USE 218
+#define CFG_PIN_ONBOARD_DOTSTAR_CLOCK 219
+#define CFG_PIN_ONBOARD_DOTSTAR_DATA 220
```
mmoskal, line 216: why do we have separate neopixel pin and onboard neopixel?

pelikhan: Added notes. Note this change is backward compat with pybadge.

PR #1099, opened 2020-04-21
ability to unregister a frame handler

pelikhan:
Would be nice to be able to unplug some 1-time use frame handlers.


pelikhan:
Right, i would like to implement cleanly a "setTimeout" style api

PR #1103, opened 2020-05-03
cleanup keyboard implementation

pelikhan:
- [x] Don't track keys in ts, it is done in C++.
- [x] call C++ flush to clear all keys

This also requires a fix in codal https://github.com/lancaster-university/codal-core/commit/281838f68737771482638a480f02a8b11dc5167f

### DIFF
--- a/libs/keyboard/keyboard.ts
+++ b/libs/keyboard/keyboard.ts
@@ -156,117 +156,6 @@ const enum KeyboardModifierKey {
  */
 //% icon="\uf11c" color="#303030"
 namespace keyboard {
-    // prevent catastrophic pounding of keyboard
-    class State {
-        private _keysDown: number[];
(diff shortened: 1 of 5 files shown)


libs/keyboard/keyboard.ts
```diff
@@ -156,117 +156,6 @@ const enum KeyboardModifierKey {
  */
 //% icon="\uf11c" color="#303030"
 namespace keyboard {
-    // prevent catastrophic pounding of keyboard
```
mmoskal, line 159: this is no longer needed?

pelikhan: The bug was c++ was emitting down instead of up

pelikhan: Otherwise it’s already updated

PR #1104, opened 2020-05-05
apply first frame of animation immediately

jwunderl:
fix issue we ran into on stream today; the animation does not apply until the first interval has passed, so it's always delayed when using with walking animations / etc. This makes it so the leading is frame is shown immediately

e.g. fixes https://makecode.com/_6aPCLTH8DJCz so that the pink circle immediately shows up when you press A, rather than after 500ms

### DIFF
--- a/libs/animation/animation.ts
+++ b/libs/animation/animation.ts
@@ -424,6 +424,8 @@ namespace animation {
 
                 game.eventContext().registerFrameHandler(scene.ANIMATION_UPDATE_PRIORITY, () => {
                     state.animations = state.animations.filter((anim: SpriteAnimation) => {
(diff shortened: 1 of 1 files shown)


libs/animation/animation.ts
```diff
@@ -424,6 +424,8 @@ namespace animation {
 
                 game.eventContext().registerFrameHandler(scene.ANIMATION_UPDATE_PRIORITY, () => {
                     state.animations = state.animations.filter((anim: SpriteAnimation) => {
+                        if (this.sprite.flags & sprites.Flag.Destroyed)
+                            return true;
```
riknoll, line 428: Shouldn't this be false?

PR #1109, opened 2020-05-15
rssi support in sim

pelikhan:
Support an RSSI value from a simulator.
- [x] test hardware
- [x] test sim

### DIFF
--- a/libs/radio/radio.ts
+++ b/libs/radio/radio.ts
@@ -62,7 +62,7 @@ namespace radio {
 
     function handleDataReceived() {
         let buffer: Buffer = readRawPacket();
-        while (buffer && buffer.length) {
+        while (buffer) {
             lastPacket = RadioPacket.getPacket(buffer);
             switch (lastPacket.packetType) {
                 case PACKET_TYPE_NUMBER:
@@ -84,9 +84,8 @@ namespace radio {
                         onReceivedStringHandler(lastPacket.stringPayload);
                     break;
             }
-
-                // read next packet if any
-                buffer = readRawPacket();
+            // read next packet if any
+            buffer = readRawPacket();
         }
     }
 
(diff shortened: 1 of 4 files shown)


libs/radio/sim/radio.ts
```diff
@@ -33,7 +33,17 @@ namespace pxsim.radio {
     export function readRawPacket() {
         const state = pxsim.getRadioState();
         const packet = state.datagram.recv();
-        return new RefBuffer(packet.payload.bufferData)
+        const buf = packet.payload.bufferData;
+        const n = buf.length;
+        if (!n)
+            return undefined;
+
+        const rbuf = BufferMethods.createBuffer(n + 4);
```
mmoskal, line 41: are these 4 bytes at the end stripped somewhere?

libs/radio/radio.ts
```diff
@@ -62,7 +62,7 @@ namespace radio {
 
     function handleDataReceived() {
         let buffer: Buffer = readRawPacket();
-        while (buffer && buffer.length) {
+        while (buffer) {
```
pelikhan, line 65: readRawPacket returns null/undefined when no packet is available

libs/radio/sim/radio.ts
pelikhan: they stay around, the typescript parser knows about those

pelikhan: https://github.com/microsoft/pxt-common-packages/blob/master/libs/radio/radio.ts#L182

pelikhan: We need to store the RSSI per packet; not globally otherwise it is incorrect.

pelikhan:
I did add a few checks for handling nulls.

libs/radio/radio.ts
```diff
@@ -164,6 +163,7 @@ namespace radio {
 
     export class RadioPacket {
         public static getPacket(data: Buffer) {
+            if (!data) return undefined;
```
pelikhan, line 166: null packet may be fed when calling readRawPaket

PR #1110, opened 2020-05-19
Friction

pelikhan:
Adds friction to sprite (fx and fy)
Is it worth adding 2 Fx8 per sprite?

download

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -228,10 +228,10 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         } else if (sprite._fx) {
             const fx = Fx.mul(sprite._fx, dtSec);
             const c = Fx.compare(sprite._vx, fx);
-            if (c < 0) // v < f, v = v += f
-                sprite._vx = Fx.add(sprite._vx, fx);
+            if (c < 0) // v < f, v += f
+                sprite._vx = Fx.min(Fx.zeroFx8, Fx.add(sprite._vx, fx));
             else if (c > 0) // v > f, v -= f
-                sprite._vx = Fx.sub(sprite._vx, fx);
+                sprite._vx = Fx.max(Fx.zeroFx8, Fx.sub(sprite._vx, fx));
             else
                 sprite._vx = Fx.zeroFx8
         }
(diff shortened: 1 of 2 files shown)


libs/game/physics.ts
```diff
@@ -215,24 +215,47 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
+                        sprite._ay,
+                        dtSec
+                    )
                 )
-            )
-        );
+            );
+        } else if( sprite._fy) {
+            const fy = Fx.mul(sprite._fy, dtSec);
+            const c = Fx.compare(sprite._vy, fy);
```
jwunderl, line 251: Shouldn't this be comparing vy with 0 instead of fy? 

Also, if abs(vy) < abs(fy), this will circle around zero indefinitely, right? Maybe worth a max / min on the results when assigning below to make sure it never passes 0

libs/game/sprite.ts
```diff
@@ -130,6 +132,27 @@ class Sprite extends sprites.BaseSprite {
...
 
+    //% group="Physics" blockSetVariable="mySprite"
+    //% blockCombine block="fx (friction x)" callInDebugger
+    get fx(): number {
+        return Fx.toFloat(this._fx)
+    }
+    //% group="Physics" blockSetVariable="mySprite"
+    //% blockCombine block="fx (friction x)"
+    set fx(v: number) {
+        this._fx = Fx8(Math.max(0, v))
```
jwunderl, line 143: Hm, negative friction for "slipperyness" is almost interesting, much more reasonable to just not allow it though

pelikhan:
So we could remove unused properties to data so that nothing breaks. Would be worth doing a cleanup in the next iteration.

libs/game/physics.ts
pelikhan: It pulls vy towards 0, assigns it to 0 if abs(vy) < abs(fy)

pelikhan: The else handles the close to zero case. Also this bypasses computing vx/vy is ax/ay is zero

libs/game/sprite.ts
pelikhan: It will lead to confusing, and fun, behavior but mostly confusing bugs.

libs/game/physics.ts
jwunderl: https://makecode.com/_fosiE8au7YD1 setting fx to 500 causes it to waver between `vx == 500 * dtSec` and `vx == 0` each frame and move to the right

the else here is handling if _vy === fy, so it's the same as if it were omitted and the else if became an else I think

This should work (w/ same for vx)

```typescript
const c = Fx.compare(sprite._vy, Fx.zeroFx8);
if (c < 0)
    sprite._vy = Fx.min(Fx.add(sprite._vy, fy), Fx.zeroFx8);
else if (c > 0)
    sprite._vy = Fx.max(Fx.sub(sprite._vy, fy), Fx.zeroFx8);
```

pelikhan:
with fx 500

PR #1114, opened 2020-05-25
?? does not seem to be supported with the current ts in makecode

pelikhan:
Our TS version does not like this one.


pelikhan:
Might be my local setup who got messed up again with TS version... but it was complaining.

jwunderl:
hm, that would make sense if you haven't npm installed common packages lately; not too many changes using new features in common packages that would require new typescript over the old one

PR #1115, opened 2020-06-04
Allow melody playing through jacdac

mmoskal:
@pelikhan do you want to add block for `playMelody`? I'm not sure I'm too good at that


pelikhan:
You can merge this one and i'll add the play melody block later. It's a matter of copying the comments from
https://github.com/microsoft/pxt-common-packages/blob/5ce17643c53162b82914394183db03a97ce8cb20/libs/mixer/melody.ts#L126

PR #1116, opened 2020-06-09
tile overlaps should handle overlaps not in the direction of movement

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1772

Here is a test program that demonstrates the bug:

https://makecode.com/_D497oY0yViu2

The fish should hit the red orb, teleport to the green orb, and then instantly teleport to the top-right corner. I'll check this program into arcade once this is merged and bumped.

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -372,6 +372,8 @@ class ArcadePhysicsEngine extends PhysicsEngine {
             s._lastY
         );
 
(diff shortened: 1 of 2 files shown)


eanders-MS:
@riknoll Is this ready to merge?

abchatra:
@riknoll ping.

PR #1120, opened 2020-06-12
Add dropdown block for camera properties

eanders-MS:
Original description:
> camera-right: Returns the x coordinate of the camera's right edge (right edge of the screen).
> camera-bottom: Returns the y coordinate of the camera's bottom edge (bottom edge of the screen).

New description:

Added a new block to get camera properties: `left`, `top`, `right`, `bottom`. Deprecated corresponding `camera-left` and `camera-top` blocks.

Resolves https://github.com/microsoft/pxt-arcade/issues/1836

### DIFF
--- a/libs/game/camera.ts
+++ b/libs/game/camera.ts
@@ -44,6 +44,12 @@ namespace scene {
             }
         }
 
+        get x() {
+            return this.drawOffsetX + (screen.width >> 1);
+        }
+        get y() {
+            return this.drawOffsetY + (screen.height >> 1);
+        }
         get left() {
             return this.drawOffsetX;
         }
--- a/libs/game/camera.ts
+++ b/libs/game/camera.ts
@@ -44,6 +44,19 @@ namespace scene {
             }
         }
 
+        get left() {
+            return this.drawOffsetX;
+        }
+        get right() {
+            return this.drawOffsetX + screen.width;
+        }
+        get top() {
+            return this.drawOffsetY;
+        }
+        get bottom() {
+            return this.drawOffsetY + screen.height;
+        }
+
         shake(amplitude: number = 4, duration: number = 1000) {
             if (amplitude <= 0 || duration <= 0) {
                 this.shakeStartTime = undefined;
--- a/libs/game/scenes.ts
+++ b/libs/game/scenes.ts
@@ -1,6 +1,14 @@
 /**
  * Control the background, tiles and camera
  */
+
+enum CameraProperty {
+    left,
+    right,
+    top,
+    bottom
+}
+
 //% weight=88 color="#4b6584" icon="\uf1bb"
 //% groups='["Screen", "Effects", "Tiles", "Collisions", "Camera"]'
 //% blockGap=8
@@ -149,6 +157,7 @@ namespace scene {
     //% blockId=cameraleft block="camera left"
     //% group="Camera"
     //% help=scene/camera-left
+    //% deprecated=true
     export function cameraLeft() {
         const scene = game.currentScene();
         return scene.camera.drawOffsetX;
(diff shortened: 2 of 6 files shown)


eanders-MS:
New language blocks seemed like the more friendly/discoverable solution, but an alternative would be to update the camera-left docs to include a sample for calculating camera-right, e.g.:

image

darzu:
I definitely think it'd be great if we provided right/bottom but I'm concerned about adding yet more blocks to the already very full "scene" category.

I'd  propose we create a camera property block getter and setter that has top/right/left/bottom/x/y like we have for sprites:
image

In fact, we might want the camera to have vx/vy/ax/ay as well since we've had requests for smooth camera transitions.

One setter and one getter block would use the same footprint we're using now. I'd propose we run a blocks upgrade rule to migrate existing "camera left" usage to the "camera [left]" getter if we go this direction.

darzu:
Thinking about this a little more, camera width/height might also make sense. We could have a viewport that is less than the screen height, and possibly multiple cameras at once with different scaling for e.g. a realtime minimap. This is a common technique in full-blown game engines.

pelikhan:
+1 on implementing using a drop-down . And only 1 height :)

eanders-MS:
+1 for not just adding more blocks. Are you thinking something along these lines?

image

Related thought: left/right/top/bottom getters might be useful for sprites too.

pelikhan:
We have those for sprites :) very useful

eanders-MS:
@darzu I love the idea of multiple viewports of varying size. Perhaps a future feature.

eanders-MS:
@riknoll, @shakao curious what you think of this design, if these changes look good to you.

libs/game/scenes.ts
```diff
@@ -160,8 +169,27 @@ namespace scene {
...
     export function cameraTop() {
         const scene = game.currentScene();
         return scene.camera.drawOffsetY;
     }
+
+    /**
+     * Returns the specified camera property
+     * @param property The property to get
+     */
+    //% blockId=getcameraproperty block="camera %property"
```
shakao, line 182: slight preference for "$" over "%" in blockdefs (% matches the index of the param and $ matches the name)

libs/game/scenes.ts
```diff
@@ -1,6 +1,14 @@
 /**
  * Control the background, tiles and camera
  */
+
+enum CameraProperty {
+    left,
```
pelikhan, line 6: Follow typescript naming inventions (capitalized) then use % block=« ... » to make loalizable

libs/game/camera.ts
```diff
@@ -44,6 +44,19 @@ namespace scene {
...
+        get left() {
+            return this.drawOffsetX;
+        }
+        get right() {
+            return this.drawOffsetX + screen.width;
+        }
+        get top() {
+            return this.drawOffsetY;
+        }
+        get bottom() {
```
pelikhan, line 56: x, y like sprites too?

libs/game/scenes.ts
```diff
@@ -160,8 +169,27 @@ namespace scene {
...
+
+    /**
+     * Returns the specified camera property
+     * @param property The property to get
+     */
+    //% blockId=getcameraproperty block="camera %property"
+    //% group="Camera"
+    //% field.defl=0
+    //% help=scene/get-camera-property
+    export function getCameraProperty(property: CameraProperty) {
```
pelikhan, line 186: Don’t use « get » prefix in APIs

pelikhan: Add return type do we guarantee it does not change he accidentally.

libs/game/scenes.ts
```diff
@@ -160,8 +169,27 @@ namespace scene {
...
         return scene.camera.drawOffsetY;
     }
+
+    /**
+     * Returns the specified camera property
+     * @param property The property to get
+     */
+    //% blockId=getcameraproperty block="camera %property"
+    //% group="Camera"
+    //% field.defl=0
```
shakao, line 184: don't need this, it should default to the first value automatically

eanders-MS: Removed.

libs/game/scenes.ts
eanders-MS: Fixed.

shakao:
i like the dropdown! i think it makes a lot of sense to match the sprite property access block

libs/game/camera.ts
eanders-MS: x, y added.

riknoll:
I also like it; consistent with the rest of our arcade blocks

libs/game/scenes.ts
darzu: WOW. TIL. this explains some confusion I had.

riknoll: Always use `$`!!!

`%` makes it much easier to accidentally break localization or APIs. It has caused many regressions in the past, we just can't upgrade the existing APIs because it would change the crowdin strings

darzu:
@eanders-MS , want to create an uploadtrg for this?

1. visit https://makecode.com/oauth/gettoken
2. set the token as env var for the next shell script e.g. `export PXT_ACCESS_TOKEN=`
3. run pxt uploadtrg inside pxt-arcade

this will build arcade and also produce a unique URL under https://arcade.makecode.com/app/XXXXX
that will have your changes so we can test them out.

@livcheerful and @jwunderl for awareness of how to do this

PR #1122, opened 2020-06-16
Change projectile blocks default behavior

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/1899
fixes https://github.com/microsoft/pxt-arcade/issues/1897

two main changes here:
1. defaulting velocity for new projectiles to be vx = 50, vy = 50 to match default sprite velocity
2. Not setting a random velocity on projectiles who have 0 velocity set


riknoll:
Where are you changing the default?

livcheerful:
> Where are you changing the default?

The default for velocities are 50 to begin with - the projectiles setting vy to 100 was overriding that

PR #1124, opened 2020-06-19
Add block to get tile from screen coords

eanders-MS:
Resolves https://github.com/microsoft/pxt-arcade/issues/1983 (if consensus is this is a good solution).

Adds a block to get a tile from screen coords:

image

Erring on the side of too many reviewers to get feedback on whether a new block is the best solution here.

### DIFF
--- a/libs/game/docs/reference/tiles/get-tile-location-from-screen.md
+++ b/libs/game/docs/reference/tiles/get-tile-location-from-screen.md
@@ -1,0 +1,96 @@
+# get Tile Location From Screen
+
+Get a tile location from a pair of screen coordinates.
+
+```sig
+tiles.getTileLocationFromScreen(0, 0)
+```
+
+## Parameters
+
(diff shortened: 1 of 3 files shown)


eanders-MS:
I wonder if the block signature is too similar to the col/row variant and could be a source of confusion. Maybe instead of "x" and "y", go with "screen x", "screen y".

pelikhan:
Honestly both names are confusing. Instead of peice meal fixes to the tilemap, we should do an overall review of the apis.

shakao:
oh, you should check out the "tilemaps" extension (github.com/microsoft/pxt-tilemaps)! richard and daryl came up with a bunch of supplemental tilemap APIs & we're hoping to move some of them into arcade proper for this release. there's a "location of [mySprite]" block there that does this coordinate system conversion, i believe

eanders-MS:
@shakao oh that extension has lots of nice improvements. Closing this PR in favor of that work.

PR #1131, opened 2020-07-06
Check for walls during movement animations

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/1677

Here's my change so that sprites animated with movement animations will respect walls. The issue was that any movements that are more than a pixel at time don't get checked for wall collisions, so I'm doing it myself before moving the sprite. It'll calculate a line (using the line logic from the Sprite Editor) and check along every point if we'll be colliding with a wall.

I'm not sure how much movement animations are used and if we're concerned about performance. I can try to optimize the check by simplifying the line calculation or by only checking every 16 positions - let me know what you think!

Here are two gifs to show the performance difference with the change. I made a camera sprite to move around and had a bunch of other sprites just oscillating back and forth. It's kinda hard to tell from the gifs, but the change makes it clunkier to move around.

_With change, 30 moving sprites - it's harder to move around without dragging_
a6670122-953b-43bd-9a53-83b960adf248

_Without change, 30 moving sprites - pretty smooth_
441eab99-a796-45cc-946f-0890cc89318a


riknoll:
I think this might be too big of a perf degradation.

Not sure the bresenham ray-tracing is necessary. Maybe instead we should just add another internal moveSprite method that always runs collisions in the physics engine? Otherwise we'll need to replicate the tilemap collision logic here.

jwunderl:
Hmm, could we handle with https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/physics.ts#L576 by adding a flag to say 'force check collisions for non-ghost sprites`

livcheerful:
> Hmm, could we handle with https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/physics.ts#L576 by adding a flag to say `force check collisions for non-ghost sprites` that overrides the check against movement being within one step?

hmmmmm I just tried this. The issue with only checking in `moveSprite` is that it'll only account for the current position, and not everything in between. So if we have a line movement that would take the sprite beyond the wall, it would allow that. I could have a return value, and stop animations if we can't move? I'm not sure if I can stop animations in a certain direction

(gif loops at a bad spot, sorry about that D:)
e015bca0-b766-414b-a487-6b5b53a6cb8b

@riknoll oo I'll look into how the tilemap does collisions

livcheerful:
@riknoll I made the change to use the tilemap logic :D it's super fast now! and the walls block properly

libs/animation/animation.ts
```diff
@@ -308,7 +322,9 @@ namespace animation {
             this.setStart = false;
         }
 
-        apply(target: Sprite, nodeTime: number, interval: number) {};
+        apply(target: Sprite, nodeTime: number, interval: number): boolean {
+            return true;
+        };
```
riknoll, line 327: style: extra semicolon

libs/animation/animation.ts
```diff
@@ -407,6 +436,120 @@ namespace animation {
...
+                Fx8(dy)
+            );
+            return !hasCollision(ms, tm, dx, dy);
+
+        }
+        return true;
+    }
+
+    // This is using a modified version of tilemapCollisions()
+    function hasCollision(movingSprite: MovingSprite, tm: tiles.TileMap, dx: number, dy: number): boolean {
```
riknoll, line 460: Should this be in the physics engine? I don't like duplicating this much code here. What changes are there?

libs/animation/animation.ts
```diff
@@ -279,25 +281,37 @@ namespace animation {
...
-                this.nodes[this.lastNode].apply(target, interval, interval); // Applies the last state of the previous node in case it was missed (this makes sure all moveTos fire)
-
+                if (this.nodesEnabled[this.lastNode]) this.nodes[this.lastNode].apply(target, interval, interval); // Applies the last state of the previous node in case it was missed (this makes sure all moveTos fire)
                 if (nodeIndex >= this.nodes.length) return true; // Once the nodeIndex is past the last item of the array, only then end the animation
             }
-            this.lastNode = nodeIndex;
 
-            this.nodes[nodeIndex].apply(target, nodeTime, interval);
+            this.lastNode = nodeIndex;
+            if (this.nodesEnabled[nodeIndex]) {
```
riknoll, line 309: Not sure that nodesEnabled needs to be an array. You only check it for the current node, right?

libs/animation/animation.ts
```diff
@@ -340,11 +357,15 @@ namespace animation {
             super();
         }
 
-        apply(target: Sprite, nodeTime: number, interval: number) {
+        apply(target: Sprite, nodeTime: number, interval: number): boolean {
             const x = Math.round(((this.p1.x - this.p0.x) / interval) * nodeTime) + this.p0.x;
             const y = Math.round(((this.p1.y - this.p0.y) / interval) * nodeTime) + this.p0.y;
-
+            if (!isClearPath(interval, target, target.x, target.y, x, y))
+            {
```
riknoll, line 364: style: we put open curly braces on the same lines as the if

libs/animation/animation.ts
```diff
@@ -367,7 +388,12 @@ namespace animation {
             const x = Math.round(a * this.p0.x + b * this.p1.x + c * this.p2.x);
             const y = Math.round(a * this.p0.y + b * this.p1.y + c * this.p2.y);
 
+            if (!isClearPath(interval, target, target.x, target.y, x, y))
```
riknoll, line 391: style: same as above

libs/animation/animation.ts
livcheerful: - I cut out the collision & overlapped tiles arrays
- Instead of checking if xDiff or yDiff is bigger than zero to check, I'm checking velocity instead
- There was also the bounce logic, clipping logic, and other movement things that I took out

libs/animation/animation.ts
livcheerful: ooooo yes ty for the catch

libs/animation/animation.ts
livcheerful: only for current node and the node before. I can switch it to be two bools instead!

libs/animation/animation.ts
livcheerful: Fixed!

libs/game/physics.ts
```diff
@@ -107,6 +122,130 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         this.map.draw();
     }
 
+    checkHorizontalCollisions(right:boolean, xDiff: Fx8, yDiff:Fx8, movingSprite: MovingSprite, tm: tiles.TileMap): TilemapCollisionHandlers {
```
riknoll, line 125: whitespace

package.json
```diff
@@ -24,7 +24,7 @@
     "libs/**/*"
   ],
   "dependencies": {
-    "pxt-core": "^6.1.25",
+    "pxt-core": "^5.37.15",
```
riknoll, line 27: Another accidental revert

libs/animation/animation.ts
```diff
@@ -407,6 +436,46 @@ namespace animation {
         }
     }
 
+    function isClearPath(interval: number, target: Sprite, x0: number, y0: number, x1: number, y1: number): boolean {
+        const tm = game.currentScene().tileMap;
+        if (tm && tm.enabled ) {
```
jwunderl, line 441:
```suggestion
        if (tm && tm.enabled) {
```

libs/game/docs/reference/game/on-update-interval.md
```diff
@@ -8,7 +8,7 @@ game.onUpdateInterval(500, function () {
 })
 ```
 
-There are events available to run code when sprites overlap, collide, are created, or destroyed. Also, you can use events to take action when buttons are pressed or when game counts reach zero. When you want to have code to control what happens in a game on a regular basis though, you need to run that code in an update function.
+There are events available to run code when sprites overlap, collide, are created, or destroyed. Also, you can use events to take action when buttons are pressed or whe game counts reach zero. When you want to have code to control what happens in a game on a regular basis though, you need to run that code in an update function.
```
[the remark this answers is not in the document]
livcheerful: ahhhhh yes thanks

libs/game/physics.ts
```diff
@@ -379,62 +518,12 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
-                    right ?
-                        Fx.add(hbox.right, Fx.oneFx8)
-                        :
-                        Fx.sub(hbox.left, Fx.oneFx8),
-                    Fx.oneHalfFx8
-                ),
-                tileScale
-            );
-            const collidedTiles: sprites.StaticObstacle[] = [];
+            const collisionHandlers = this.checkHorizontalCollisions(right, xDiff, yDiff, movingSprite, tm);
```
jwunderl, line 522: I actually had these pulled out into helpers at one point, but inlining them had about a 2-3 fps difference when playing jumpy platformer on meowbit (which was about 20 fps or so if I'm remembering right)

Do you have a minute to compare fps with this change vs current master? I can take a look in a bit, went out to grab something to eat. For what it's worth, there have been a few improvements in this area since I wrote this part so it might not make a difference anymore

livcheerful: Yes I can do it! is there a good way to do it though 😅

(i'm glad you're eating)

livcheerful:
> Overall looks like a good refactor. @jwunderl can you double check this? Shouldn't affect perf too much?

yay thanks for taking a look! I'll do a perf test too

libs/game/physics.ts
jwunderl: In the standard menu there's a "show fps" button that turns on some stats at the bottom of the screen - not perfect but good for a rough estimate like what's needed here ~

jwunderl: In the standard menu there's a "show fps" button that turns on some stats at the bottom of the screen - not perfect but good for a rough estimate like what's needed here ~

riknoll: If you put this line in a game:

```ts
game.stats = true;
```

It will print out the FPS to the screen on hardware.

I know we have some physics tests checked in, but I don't know if they are perf related: https://github.com/microsoft/pxt-arcade/tree/master/docs/physics-tests

jwunderl: Yeah those ones are mainly behavioral, the shapes one is a light perf test but not consistant at the moment

livcheerful: oo cool so with my test game (https://makecode.com/_WzME6j8uUewe)

Running 50 sprites on **beta**:
(A) Going left: 10 - 15 fps
(B) Bounce: 4.5 - 5.7 fps

Running 50 sprites with **these changes**:
(A) Going left: 10.4 - 12.9 fps
(B) Bounce: 3.3 - 4.9 fps

It looks like it makes it slower :\  or at least it's harder to see the faster numbers with my changes

PR #1134, opened 2020-07-08
Convert to floats for sprite properties

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1702

Return floats, not ints!

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -69,7 +69,7 @@ class Sprite extends sprites.BaseSprite {
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x" callInDebugger
     get x(): number {
-        return Fx.toInt(this._x) + (this._image.width >> 1)
+        return Fx.toFloat(this._x) + (this._image.width >> 1)
     }
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x"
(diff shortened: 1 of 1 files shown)


libs/game/sprite.ts
```diff
@@ -69,7 +69,7 @@ class Sprite extends sprites.BaseSprite {
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x" callInDebugger
     get x(): number {
-        return Fx.toInt(this._x) + (this._image.width >> 1)
+        return Fx.toFloat(this._x) + (this._image.width >> 1)
```
mmoskal, line 72:
```suggestion
        return Fx.toFloat(this._x) + (this._image.width / 2)
```
So that we're floats all the way?

libs/game/sprite.ts
```diff
@@ -80,7 +80,7 @@ class Sprite extends sprites.BaseSprite {
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="y" callInDebugger
     get y(): number {
-        return Fx.toInt(this._y) + (this._image.height >> 1)
+        return Fx.toFloat(this._y) + (this._image.height >> 1)
```
mmoskal, line 83:
```suggestion
        return Fx.toFloat(this._y) + (this._image.height / 2)
```

PR #1139, opened 2020-07-13
Text must be string in sprite.say

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2078

Is there a reason this was any before? I'm worried about breaking things -.-


livcheerful:
New PR at https://github.com/microsoft/pxt/pull/7322

PR #1140, opened 2020-07-13
add image equals block

jwunderl:
Not fully certain best wording, but to start:

image

Maybe can drop 'image' from the text, but could see someone being confused if they variables in and things 'randomly' break with an equals block.

Doesn't really fit in Drawing / Create / Transformations so I put it in it's own "Compare" group

### DIFF
--- a/libs/screen/image.d.ts
+++ b/libs/screen/image.d.ts
@@ -98,6 +98,10 @@ interface Image {
      * otherwise returns false.
      */
     //% shim=ImageMethods::equals
+    //% blockNamespace="images"
+    //% block="$this is equal to image $other"
+    //% this.shadow=variables_get
+    //% this.defl="picture"
     equals(other: Image): boolean;
 
     //% shim=ImageMethods::isStatic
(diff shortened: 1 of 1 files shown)


darzu:
Is it possible to use the "=" logic block instead? In .ts it could compile to ".compare()" or something

riknoll:
That would be a new feature, which I'm not sure we can sneak into this release. Touches many parts of the codebase.

riknoll:
But I like the idea, long term!

Jaqster:
Agree logic blocks would be great to use here.  I notice they work for image comparisons...
image

jwunderl:
They work in that case because they're based off reference equality / they're the same location in memory. Won't consistently work when someone uses images that aren't builtin

Jaqster:
Ah, okay.  
Also, along these same lines, if we can convert the comparison blocks to the appropriate function in TS, they we could allow string comparisons like this too...
image

jwunderl:
Yeah, same technical problem, though I think that'd be a more controversial change as that is a fairly useful error message (could maybe use a change in the wording)

riknoll:
Agree, it needs a rewording but the error itself is valid. "Hi" will never equal "Hello" so it's probably a bug

PR #1142, opened 2020-07-14
constrain speed changes applied outside physics calc

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2153

When friction was added the acceleration was moved into an if, but the constrain always needs to occur. This was allowing unbounded speed when the speed change is applied outside physics (in this case, the repeat was adding 30vy every frame, and since the physics engine thought the max was 500 it didn't handle it well when speeds got to around a few thousand)

As a side note, I'm pretty happy I got to make a branch named 'fixSpaceMan' and have it be an accurate name

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -235,7 +235,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         }
 
         if (sprite._ay) {
-            sprite._vy =Fx.add(
+            sprite._vy = Fx.add(
                 sprite._vy,
                 Fx.mul(
                     sprite._ay,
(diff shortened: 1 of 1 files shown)


libs/game/physics.ts
```diff
@@ -237,13 +235,11 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
 
         if (sprite._ay) {
-            sprite._vy = this.constrain(
-                Fx.add(
-                    sprite._vy,
-                    Fx.mul(
-                        sprite._ay,
-                        dtSec
-                    )
+            sprite._vy =Fx.add(
```
riknoll, line 238: whitespace

PR #1143, opened 2020-07-15
createTilemap: Default scale to TileScale.Sixteen

eanders-MS:
**The problem**
In the JS editor, typing `tiles.createTilemap(` and tab to autocomplete fills in a scale of `TileScale.Eight`, but it should be `TileScale.Sixteen`.

**My solution**
I wasn't able to find a way to have it explicitly become `TileScale.Sixteen`, but making it the default value is an improvement, I think. Curious what others think, and if there's a better solution.

Fixes https://github.com/microsoft/pxt-arcade/issues/2112

### DIFF
--- a/libs/game/tilemap.ts
+++ b/libs/game/tilemap.ts
@@ -427,7 +427,7 @@ namespace tiles {
         return i;
     }
 
-    export function createTilemap(data: Buffer, layer: Image, tiles: Image[], scale: TileScale): TileMapData {
+    export function createTilemap(data: Buffer, layer: Image, tiles: Image[], scale = TileScale.Sixteen): TileMapData {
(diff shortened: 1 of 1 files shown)


darzu:
So this fixes auto-complete and snippets from the toolbox? If so, this seems like a great solution

riknoll:
does this not work?
```
//% scale.defl="TileScale.Sixteen"
```

eanders-MS:
> does this not work?
> 
> ```
> //% scale.defl="TileScale.Sixteen"
> ```

@riknoll That worked beautifully! Autocomplete fills it in now:

```ts
tiles.createTilemap(null, img` `, [], TileScale.Sixteen)
```

@darzu Does this work for you?

PR #1146, opened 2020-07-15
save screen brightness modified in settings

pelikhan:
Just like volume, save brightness modified through the settings menu. It's already capped to 10-100

### DIFF
--- a/libs/game/systemmenu.ts
+++ b/libs/game/systemmenu.ts
@@ -284,11 +284,18 @@ namespace scene.systemMenu {
     }
 
     function brightnessUp() {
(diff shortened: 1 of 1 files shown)


libs/game/systemmenu.ts
```diff
@@ -389,10 +396,17 @@ namespace scene.systemMenu {
 
     function initVolume() {
         const vol = settings.readNumber("#volume")
-        if (vol != null)
+        if (vol !== null)
```
mmoskal, line 399: Are you sure read number doesn't return undefined?

pelikhan: updated pr

PR #1147, opened 2020-07-16
Better algorithm for breaking up text in dialogs

riknoll:
This is bringing over the greedy line break algorithm I implemented in my story extension (with a few fixes). It produces much, much, better results than the current behavior. It also respects line breaks including those from blocks (i.e. both "\n" and "\\\\n").

Here is some lorem ipsum text on the current editor:

and here it is with my changes:

Note the lack of annoying hyphens!

### DIFF
--- a/libs/game/textDialogs.ts
+++ b/libs/game/textDialogs.ts
@@ -180,7 +180,7 @@ namespace game {
     }
 
     export class Dialog extends BaseDialog {
-        chunks: string[];
+        chunks: string[][];
         chunkIndex: number;
 
         constructor(width: number, height: number, frame?: Image, font?: image.Font, cursor?: Image) {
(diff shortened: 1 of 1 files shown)


libs/game/textDialogs.ts
```diff
@@ -702,5 +651,81 @@ namespace game {
         pauseUntil(() => done);
         controller._setUserEventsEnabled(true);
     }
+
+    function isBreakCharacter(charCode: number) {
+        return charCode <= 32 ||
+            (charCode >= 58 && charCode <= 64) ||
+            (charCode >= 91 && charCode <= 96) ||
+            (charCode >= 123 && charCode <= 126);
```
darzu, line 659: Maybe comment what these are?

PR #1149, opened 2020-07-16
Align meowbit acceleration axis with other boards

mmoskal:
Reverts https://github.com/microsoft/pxt-common-packages/pull/1028
Fixes https://github.com/microsoft/pxt-arcade/issues/1745


mmoskal:
@abchatra you may want that in release

PR #1152, opened 2020-07-17
dont crash when using place on random tile when tile count >>> big number

jwunderl:
When we do place on random tile, it creates an array of all of the tiles of that type; that's fine a lot of the time, but VERY BAD when that tile is the base of a big level (e.g. the standard floor in a ~80x80 tilemap)

This adds a helper to get an n-length random sample from the tiles of the given type w/ a [reservoir sample](https://en.wikipedia.org/wiki/Reservoir_sampling) (Yay for all the times I had to help students with interview questions prep!).

test that it's actually random (I tried a handful of different lengths / tile arrangements, just need to draw on tilemap): https://makecode.com/_5tx7bz2ua59o

w/ two quick runs:

image

fixes issue with Sonia's game (see teams)

### DIFF
--- a/libs/game/tilemap.ts
+++ b/libs/game/tilemap.ts
@@ -602,8 +602,8 @@ namespace tiles {
     }
 
     /**
-     *
-     * @param tile a random tile of the get
+     * Get a random tile of the given type
+     * @param tile the type of tile to get a random selection of
      */
     export function getRandomTileByType(tile: Image): Location {
         const scene = game.currentScene();
--- a/libs/game/tilemap.ts
+++ b/libs/game/tilemap.ts
@@ -315,6 +315,32 @@ namespace tiles {
             return output;
         }
 
+        public sampleTilesByType(index: number, maxCount: number): Location[] {
+            if (this.isInvalidIndex(index) || !this.enabled || maxCount <= 0) return [];
+
+            let count = 0;
+            const reservoir: Location[] = [];
+            for (let col = 0; col < this._map.width; ++col) {
+                for (let row = 0; row < this._map.height; ++row) {
+                    let currTile = this._map.getTile(col, row);
+                    if (currTile === index) {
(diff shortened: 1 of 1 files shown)


jwunderl:
Yup, it's mostly the same as sampleTilesByType, just unconditionally puts the matching tiles into the returned array. It's right above `sampleTilesByType` so if you click the little 'more context' arrow above that fn it should show up

livcheerful:
> Yup, it's mostly the same as sampleTilesByType, just unconditionally puts the matching tiles into the returned array. It's right above `sampleTilesByType` so if you click the little 'more context' arrow above that fn it should show up

oooooo nice thanks

jwunderl:
@abchatra this might be a good one to port to release? It's a very easy bug to run into, and anecdotally I feel like I've seen a lot of forum users making largish mazes on the forums

libs/game/tilemap.ts
```diff
@@ -315,6 +315,32 @@ namespace tiles {
...
+
+            let count = 0;
+            const reservoir: Location[] = [];
+            for (let col = 0; col < this._map.width; ++col) {
+                for (let row = 0; row < this._map.height; ++row) {
+                    let currTile = this._map.getTile(col, row);
+                    if (currTile === index) {
+                        // first **maxCount** elements just enqueue
+                        if (count < maxCount) {
+                            reservoir.push(new Location(col, row, this));
```
darzu, line 329: Hmm, @jwunderl  please correct me if I'm reading this wrong but it looks like if (0,0) matches the tile type you need, it'll always be in the reservoir and if the reservoir is size 1 it'll always be returned. Is that right?

darzu: ooh I see how this works now. Nvm

abchatra:
I will port to stable channel. Nice algo.

PR #1154, opened 2020-07-22
Run "small hitbox change" logic on image manipulation

livcheerful:
Move the logic of "not changing the hitbox if it's a small change" out of setImage() and into setHitbox() so we can use it if the sprite has its image manipulated.

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -253,7 +253,11 @@ class Sprite extends sprites.BaseSprite {
     setImage(img: Image) {
         if (!img) return; // don't break the sprite
         this._image = img;
-        const newHitBox = game.calculateHitBox(this);
+        this.setHitbox();
+    }
+
+    setHitbox() {
+        let newHitBox = game.calculateHitBox(this);
 
         if (!this._hitbox) {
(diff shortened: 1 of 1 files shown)


abchatra:
@livcheerful do I need to take this to release?

livcheerful:
@abchatra this is something @jwunderl suggested as a result of this change in the release https://github.com/microsoft/pxt-common-packages/pull/1133/files 

I think this is a good change to take so future games won't have weird bugs

jwunderl:
Yeah, Abhijith and I were talking while figuring out merging some other stuff.

For future reference / reasoning for this change:

A while ago in `.setImage` we added logic to prevent tiny changes to the hitbox - this fixed issues where there was a sprite with an animation that changed it's dimensions by ~1 or 2 pixels and would cause it to clip into the wall.

The change to add `.setHitbox` / detect mutations in the image fixed cases where a sprite's image was changed after being set, but added a new path for setting this hitbox. This change just merges the paths for creating / setting hitboxes, so that mutating a sprite's image and setting to a new image will result in the same hitbox / behave consistently

PR #1155, opened 2020-07-28
Pass received buttons as player 2 controls

livcheerful:


libs/online-multiplayer/sim/multiplayer.ts
```diff
@@ -18,10 +18,9 @@ namespace pxsim {
 
     export interface SimulatorMultiplayerMessage extends SimulatorMessage {
         type: "multiplayer";
-        origin: "server" | "client";
+        origin?: "server" | "client";
         content: string;
-        id: number;
-        data: any;
+        id?: number;
```
livcheerful, line 23: i dunno why this is showing as a change since it seems like you already made it @jwunderl ?

jwunderl: weird

PR #1156, opened 2020-07-28
Store most recent image message

livcheerful:


libs/online-multiplayer/multiplayer.d.ts
```diff
@@ -2,6 +2,8 @@ declare namespace multiplayer {
     //% shim=multiplayer::postImage
     function postImage(im: Image, goal: string): void;
 
+    function setImage(im: Image): void;
```
jwunderl, line 5: need decl for getOrigin

PR #1162, opened 2020-08-07
basic support for playdate

mmoskal:


pelikhan:
What is needed to drop the uf2 on the playdate? new firmware?

PR #1169, opened 2020-08-13
fix sprite.say(0)

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2331; we still want it to bail out on null, undefined, and "" (as that's commonly used as 'clear say' in games)

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -467,7 +467,7 @@ class Sprite extends sprites.BaseSprite {
     //% help=sprites/sprite/say
     say(text: any, timeOnScreen?: number, textColor = 15, textBoxColor = 1) {
         // clear say
-        if (text == null || text === "") {
+        if (text === null || text === undefined || text === "") {
             this.updateSay = undefined;
             if (this.sayBubbleSprite) {
                 this.sayBubbleSprite.destroy();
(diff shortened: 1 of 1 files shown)


libs/game/sprite.ts
```diff
@@ -467,7 +467,7 @@ class Sprite extends sprites.BaseSprite {
     //% help=sprites/sprite/say
     say(text: any, timeOnScreen?: number, textColor = 15, textBoxColor = 1) {
         // clear say
-        if (!text) {
+        if (text == null || text === "") {
```
darzu, line 470: Maybe instead of "==", use "===" and explicit null and undefined checks?

jwunderl: Sure, behaves same way and probably makes it a bit more clear why it's not ! that way ~

eanders-MS: I'd suggest adding a comment about wanting to allow `0` though. Otherwise someone like me might happen across this and try to "simplify" it.

darzu: I had to check in the console to make sure "0 == null" and "null == 0" are both false. I'm actually not sure why they are false since I would have expected a coercion.

jwunderl: Yeah, null only equals itself and undefined, unless you're doing a comparison, as `null >= 0` is true and `null > 0` is false as the comparison coerces null to 0

Javascript can be weird ¯\\_(ツ)_/¯

PR #1177, opened 2020-09-24
Add tuned noise and cycle noise implementations

klausw:
This adds a few new waveforms for special effects:

SW_TUNEDNOISE (waveform 4) - a variable-pitch noise based on pseudorandom impulses.
It sounds similar to SW_NOISE at maximum frequency, but reacts to
the frequency parameter to modify the perceived pitch. (It reuses the
waveform number for the retired SW_METALLIC waveform.)

SW_CYCLENOISE_4, SW_CYCLENOISE_5, SW_CYCLENOISE_6 (waveforms 16-18) -
variable-pitch short-sequence repeating pseudorandom impulses with
4/5/6 bit sequences (16/32/64 cycles).

The sounds are vaguely inspired by the POKEY chip used in 8-bit Atari
computers, cf. https://www.atariarchives.org/dere/chapt07.php, with its
4/5/17-bit polynomial counters that modulate square waves, but uses a
simpler generator and doesn't try to emulate any particular hardware.

See
https://forum.makecode.com/t/testing-modifications-to-the-sound-engine/3351
for context and links to example sound output.


klausw:
Warning, this is **untested**, I'm blocked on https://github.com/microsoft/pxt-arcade/issues/2423 which crashes my PyGamer when using `music.playInstructions` even without this patch included.

libs/mixer/melody.cpp
```diff
@@ -56,23 +60,85 @@ static int sineTone(uintptr_t userData, uint32_t position) {
...
     return position < (102 * (wave - SW_SQUARE_10 + 1)) ? -0x7fff : 0x7fff;
 }
 
-static int silenceTone(uintptr_t userData, uint32_t position) {
-    (void)userData;
+static int tunedNoiseTone(uint8_t wave, uint32_t position, int channel, uint8_t cycle) {
+    // Generate a square wave filtered by a random bit sequence. Since the generator
+    // is called multiple times per wave, use static data to ensure we only generate
+    // a random bit once per wave, and then reuse it for future calls for that wave.
+    // Do this per channel to ensure results remain consistent when mixing multiple
```
klausw, line 87: FYI, if you'd prefer to avoid passing the extra "channel" argument to all the generators, it should be possible to skip the per-channel tracking and use simple static state instead. This would mean that mixing multiple channels each playing tuned noise would have extra changes when switching channels. In the current setup, this would be a flipped bit every numsamples, so something like a 44Hz added noise assuming 44 KHz sample rate and 1024 sample buffers. This may well be inaudible, but I haven't been able to test that.

libs/mixer/melody.h
```diff
@@ -7,9 +7,13 @@
 #define SW_TRIANGLE 1
 #define SW_SAWTOOTH 2
 #define SW_SINE 3 // TODO remove it? it takes space
```
klausw, line 9: Drive-by comment about the sine tone - if the space used for the lookup table is a concern, it should be fairly straightforward to replace the table lookup with an approximating function, i.e. https://en.wikipedia.org/wiki/Bhaskara_I%27s_sine_approximation_formula or similar. For sound generation purposes it should not be necessary to be very accurate.

mmoskal: The lookup table is 512 bytes, the interpolation would be probably under 100, so if you feel like doing a PR, that would be great!

mmoskal:
I'll look into microsoft/pxt-arcade#2423 tomorrow.

libs/mixer/melody.cpp
mmoskal: Maybe we can pass PlayingSound to the generator function, and also add a word or two there for wave-form-data? Then maybe we can get rid of all that static state in these functions.

libs/mixer/melody.h
klausw: Done, see https://github.com/microsoft/pxt-common-packages/pull/1178

mmoskal:
You can now merge my fix and test your changes in hardware. Once tested, I'll be happy to merge as is or you can try to move the state to the PlayingSound object.

klausw:
Thanks for the fix, it's working on hardware now. (I did need to fix a bug, I got a variable name mixed up.)

I've also added a playingSound pointer to the generator function and use that instead of the wave and channel args, using a generatorState word for the pseudorandom waves. I've kept the random seeds globally static.

klausw:
Let me know if you have any further comments, I don't currently have any outstanding changes planned for this PR. I had renamed some of the new #define names, I'll update the simulator PR to match that.

I did run across an unexpected issue when playing sweeps in sequence which seems to be pre-existing, see https://github.com/microsoft/pxt-arcade/issues/2441 for that. I worked around it by inserting some short silences between frequency sweeps.

https://makecode.com/_1wwRUaMpCDho is a quick demo using some of the sound sequences from the forum thread, use the A and B buttons to play sounds.

mmoskal (approved):
Awesome!

Could you please fix the merge conflicts?

klausw:
> Awesome!
> 
> Could you please fix the merge conflicts?

Done.

mmoskal:
Thank you!

klausw:
I'm happy to contribute. Out of curiosity, what's the expected timeline for the changes to be available on the public arcade.makecode.com site? Looking forward to trying out the new sounds in projects :-)

mmoskal:
I'll try to push it to beta today, but the main site is only updated once every few months.

PR #1178, opened 2020-09-24
Replace sineTone lookup table with approximation

klausw:
Instead of a 512-byte lookup table, use Bhaskara's approximation
formula, with some rescaling to match the input and output ranges.

The maximum error is 54/32767 = 0.00165 which should be good enough
for sound generation purposes, especially on target devices with
small non-HiFi speakers.

### DIFF
--- a/libs/mixer/melody.cpp
+++ b/libs/mixer/melody.cpp
@@ -9,28 +9,6 @@ namespace music {
 
 SINGLETON(WSynthesizer);
 
-static const int16_t sinQ[256] = {
-    0,     201,   403,   605,   807,   1009,  1210,  1412,  1614,  1815,  2017,  2218,  2419,
-    2621,  2822,  3023,  3224,  3425,  3625,  3826,  4026,  4226,  4426,  4626,  4826,  5026,
-    5225,  5424,  5623,  5822,  6020,  6219,  6417,  6615,  6812,  7009,  7206,  7403,  7600,
-    7796,  7992,  8187,  8383,  8578,  8772,  8967,  9161,  9354,  9547,  9740,  9933,  10125,
-    10317, 10508, 10699, 10890, 11080, 11270, 11459, 11648, 11836, 12024, 12212, 12399, 12585,
-    12772, 12957, 13142, 13327, 13511, 13695, 13878, 14060, 14243, 14424, 14605, 14785, 14965,
-    15145, 15323, 15501, 15679, 15856, 16032, 16208, 16383, 16557, 16731, 16905, 17077, 17249,
-    17420, 17591, 17761, 17930, 18099, 18267, 18434, 18600, 18766, 18931, 19096, 19259, 19422,
-    19585, 19746, 19907, 20067, 20226, 20384, 20542, 20699, 20855, 21010, 21165, 21318, 21471,
-    21623, 21774, 21925, 22074, 22223, 22371, 22518, 22664, 22810, 22954, 23098, 23241, 23382,
-    23523, 23663, 23803, 23941, 24078, 24215, 24350, 24485, 24618, 24751, 24883, 25014, 25144,
-    25273, 25401, 25528, 25654, 25779, 25903, 26026, 26148, 26269, 26389, 26509, 26627, 26744,
-    26860, 26975, 27089, 27202, 27314, 27425, 27535, 27644, 27752, 27859, 27964, 28069, 28173,
-    28275, 28377, 28477, 28576, 28674, 28772, 28868, 28963, 29056, 29149, 29241, 29331, 29421,
-    29509, 29596, 29682, 29767, 29851, 29934, 30015, 30096, 30175, 30253, 30330, 30406, 30480,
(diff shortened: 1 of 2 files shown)


klausw:
To check accuracy, I tried the sineTone function with this wrapper:

```cpp
#include <stdint.h>
#include <math.h>
#include <iostream>

int main() {
    for (int x = 0; x < 1024; ++x) {
        int approx = sineTone(0, x, 0, 0);
        int sine = floor(32767*sin(x*M_PI/512));
        std::cout << x << " " << approx << " " << sine << " " << (approx - sine) << std::endl;
    }
    return 0;
}
```

Here's the output graphed using Gnuplot. Since the error isn't easily visible, I included an extra error line scaled up by a factor of 100.

```gnuplot
plot "s.dat" using 1:2 with lines title "approx", \
     "s.dat" using 1:3 title "sin", \
     "s.dat" using 1:4 with lines title "error", \
     "s.dat" using 1:($4*100) with lines title "error*100"
```

sine-approx
[sine-approx.pdf](https://github.com/microsoft/pxt-common-packages/files/5279160/sine-approx.pdf)

klausw:
@mmoskal please take a look, and ideally test it on real hardware, I just used it standalone using the wrapper snippet.

klausw:
FYI, this approach assumes that integer division is OK performance-wise. Do the target platforms all have hardware division instructions that are used by the build target? There seem to be a couple of divisions by samplesLeft in the inner loop already.

mmoskal:
This looks great, but the division may be a problem.

Division is 2-12 cycles on M4 (Arcade), but is software-implemented on M0 (eg. Circuit Playground Express). The divisions already there are only performed once per note (or per sample buffer), and this one would be once per sample.

Multiplications are 1 cycle though on M0 and up (there is apparently a smaller, slower 32 cycle multiplier but no one is using that).

Maybe we'll need to ifdef the lookup table if there is no way to approximate without division :/

Sources:
* https://developer.arm.com/documentation/ddi0439/b/Programmers-Model/Instruction-set-summary/Cortex-M4-instructions
* https://developer.arm.com/documentation/ddi0432/c/programmers-model/instruction-set-summary

klausw:
I've switched to a new approach using a fitted 5th order polynomial, with no division. It's also more accurate, with max error of 7/32767 = 0.02%. (Note to self: while using 7th century math is cool, there has apparently been some progress since then.)

sine-approx-5th
[sine-approx-5th.pdf](https://github.com/microsoft/pxt-common-packages/files/5286264/sine-approx-5th.pdf)

I've tested it on my PyGamer, it sounds OK to me: https://makecode.com/_cE4X4m1KbA1Y 

```js
controller.B.onEvent(ControllerButtonEvent.Pressed, function () {
    // Play sine waves at 880 Hz, 440 Hz, 220 Hz, 110 Hz

// ww = waveform
// ffFF = frequency, low byte first
// ddDD = duration in milliseconds
// vvVV = volume
// wwWW = end volume
// ggGG = end frequency
//  ww -- ffFF ddDD vvVV wwWW ggGG
    const sound2: Buffer = hex`
    03 00 7003 0004 0004 0000 7003

    03 00 b801 0004 0004 0000 b801

    03 00 dc00 0004 0004 0000 dc00

    03 00 6e00 0004 0004 0000 6e00
    `;
music.playInstructions(0, sound2)
})
controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
    // Play a sweep from 1024 Hz down to 1 Hz.

    const sound: Buffer = hex`
    03 00 0004 0008 0004 0004 0100
    03 00 0004 2000 0000 0000 0004
    `;
music.playInstructions(0, sound)
})
namespace music {
    //% shim=music::queuePlayInstructions
    export function playInstructions(timeDelta: number, buf: Buffer) { }
}
scene.setBackgroundColor(6)
game.splash("Press A or B to test sound")
```

klausw:
For completeness's sake, I got the new graph using the same main function as https://github.com/microsoft/pxt-common-packages/pull/1178#issuecomment-698587352 , and the following Gnuplot command to graph the resulting output:

```gnuplot
plot [0:1024] 32767*sin(x*pi/512) with points, \                                                                                   
  "s.dat" using 1:2 with lines title "approx", \                                                                                            
  "s.dat" using 1:4 with lines title "error", \                                                                                             
  "s.dat" using 1:($4*100) with lines title "error*100", \                                                                                  
  "s.dat" using 1:($4*1000) with lines title "error*1000"  
```

I also squashed the commit to make the history less confusing. The commit message includes the specific sequence of Emacs Calc commands I used to get the fitted parameters for the polynomial. (I calculated them from scratch to understand it better myself and to avoid potential IP issues. https://www.coranac.com/2009/07/sines/ had useful tips (including rescaling to 0..1 to simplify the math which indeed helped), but I didn't follow the method there specifically.)

mmoskal (approved):

mmoskal:
Thank you so much!

Next, don't worry about squashing on your side too much - we always squash&merge in pxt projects.

PR #1185, opened 2020-10-08
New emit for images for asset editor

riknoll:
To support the new emit style for images/tiles/tilemaps and animations


libs/game/assetTemplates.ts
```diff
@@ -0,0 +1,90 @@
...
+
+//% helper=getTilemapByName
+//% pyConvertToTaggedTemplate
+function tilemap16(lits: any, ...args: any[]): tiles.TileMapData { return null }
+
+//% helper=getTilemapByName
+//% pyConvertToTaggedTemplate
+function tilemap32(lits: any, ...args: any[]): tiles.TileMapData { return null }
+
+namespace assets {
```
jwunderl, line 17: worth noting this is a new top level symbol, but `assets` feels much safer / less likely to conflict than the namespaces we've had trouble with in the past & assets is generic enough we can add in more in here later.

jwunderl (approved):
LGTM

PR #1194, opened 2020-10-28
add flag to disable msg broadcast in simmessages

jwunderl:
So it can just go to the parent, instead of creating a second simulator each time.

I don't see how the cpp side of things works here, so not sure if anything would be needed there? The only use of it I've seen right away is [here](https://github.com/microsoft/pxt-jacdac-services/commit/7cdcf8e2e78bed0f3ca14a7e40b7d190e2f8aba1#diff-52838880e988354f72feddfcd58faddb78dc93e73f050139f97cd377405cb2d7R15), is the behavior just that this shim only applies on hardware so the simulator goes through the function body / calls this api?

For reference, I was looking at using this here: https://github.com/jwunderl/arcade-tick-event/blob/master/main.ts#L48

### DIFF
--- a/libs/base/controlmessage.ts
+++ b/libs/base/controlmessage.ts
@@ -4,7 +4,7 @@ namespace control.simmessages {
     export const CONTROL_MESSAGE_RECEIVED = 1;
 
     //% shim=pxt::sendMessage
-    export declare function send(channel: string, message: Buffer, noBroadcast?: boolean) : void;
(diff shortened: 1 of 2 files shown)


pelikhan:
This is not really a decision that the API user should be making. It is more a flag on the editor to determine the behavior of this message.

pelikhan (changes requested):
maybe "parentOnly" instead of "noBroadcast"?

pelikhan (approved):

PR #1196, opened 2020-10-29
Don't unregister melody if stopped and restarted

klausw:
If a melody is stopped and restarted while playing, it must
not be unregistered by the playCore handler to avoid keeping
unstoppable melodies running in the background.

Fixes microsoft/pxt-arcade#2541

### DIFF
--- a/libs/mixer/melody.ts
+++ b/libs/mixer/melody.ts
@@ -250,7 +250,11 @@ namespace music {
                     if (!loop)
                         break
                 }
-                this.unregisterMelody();
+                // Unregister the melody if it was stopped and not restarted.
+                // (It was restarted if this._player is a new MelodyPlayer.)
+                if (!this._player) {
+                    this.unregisterMelody();
+                }
             })
         }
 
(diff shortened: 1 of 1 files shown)


libs/mixer/melody.ts
```diff
@@ -250,7 +250,11 @@ namespace music {
                     if (!loop)
                         break
                 }
-                this.unregisterMelody();
+                // Unregister the melody if it was stopped and not restarted.
+                // (It was restarted if this._player is a new MelodyPlayer.)
+                if (!this._player) {
```
mmoskal, line 255:
```suggestion
                if (this._player == p) {
```
I think if it's null, it would already have been unregistered by stop(). Could you check that this works for you?

mmoskal (changes requested):

libs/mixer/melody.ts
klausw: That does seem to be the case, unregistering doesn't appear to be necessary for looping sounds

Instead, I moved the check to the (!loop) branch since non-looping melodies didn't get cleaned up when they ended.

Here's a test using a modified "music2.ts" including this change: https://makecode.com/_dLCTje3iXeMR

mmoskal (approved):

mmoskal:
Sorry for delay, I only found it again when cleaning up my mailbox.

Thank you for the PR!

PR #1197, opened 2020-11-10
fix typo breaking arcade build

jwunderl:
typo in https://github.com/microsoft/pxt-common-packages/pull/1193 breaking arcade serve; I don't think this would need to be ported to microbit/stable as it's just part of the linux build


mmoskal (approved):
Whoops, sorry

PR #1200, opened 2020-11-28
display a neopixel strip as a matrix

tballmsft:


libs/core/sim/neopixel.ts
```diff
@@ -158,14 +161,12 @@ namespace pxsim.visuals {
...
-        public pin: number;
-        public pixels: NeoPixel[];
+        private pixels: NeoPixel[];
         private viewBox: [number, number, number, number];
         private background: SVGRectElement;
 
-        constructor(pin: number) {
+        constructor(pin: number, public cols: number = 1) {
             this.pixels = [];
-            this.pin = pin;
```
pelikhan, line 168: why are we removin this?

pelikhan:
Do the regular neopixel still work with this change?

PR #1207, opened 2020-12-29
Update platform.h to include new pins for Adafruit Grand Central M4

wallarug:
on behalf of @lalauria2000. Part 2 of 2.

See: microsoft/pxt-maker#324

May need to find a better way to move to the 128 pin packages.  I just made it so that if the board is SAMD51, it will always default to 128 pins.  Not sure if this is 100% accurate since a lot of the variants do not have 128 pins.

### DIFF
--- a/libs/core---samd/platform.h
+++ b/libs/core---samd/platform.h
@@ -48,7 +48,11 @@ typedef int PinName;
 #define USB_HANDOVER 0
 
 // if we ever want to support 100+ pin packages, need to add PC,PD ports and increase this to 128
+#ifdef SAMD51
+#define DEV_NUM_PINS 128
+#else
 #define DEV_NUM_PINS 64
+#endif
 
 #define IS_ANALOG_PIN(id) 1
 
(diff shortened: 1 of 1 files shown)


libs/core---samd/platform.h
```diff
@@ -48,7 +48,11 @@ typedef int PinName;
 #define USB_HANDOVER 0
 
 // if we ever want to support 100+ pin packages, need to add PC,PD ports and increase this to 128
+#ifdef SAMD51
```
pelikhan, line 51: you could have a define DEV_PINS_128 to enable this mode.

libs/core---samd/platform.h
```diff
@@ -135,4 +139,72 @@ typedef int PinName;
 #define PB30 62
 #define PB31 63
 
-#endif
\ No newline at end of file
+// 128 pin packages
+#ifdef SAMD51
+#define PC00 = 64
```
mmoskal, line 144: There is no `=` in #define syntax, you'll need to remove them

mmoskal (changes requested):

libs/core---samd/platform.h
lalauria2000: https://github.com/lalauria2000/pxt-common-packages/commit/f51b311ec37b114913bec683ee76ad55de341040

pelikhan:
This PR is outdated I beleive.

PR #1208, opened 2021-01-03
Add - Triplet note

chatelao:
My daughter complained about the triplets missing in MakeCode - So here they are and hopefully okay ;-)

### DIFF
--- a/libs/music/music.ts
+++ b/libs/music/music.ts
@@ -123,7 +123,9 @@ enum BeatFraction {
     //% block="2"
     Double = 32,
     //% block="4",
-    Breve = 64
+    Breve = 64,
+    //% block="1/3",
+    Triplet = 128
 }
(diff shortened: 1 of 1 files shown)


riknoll (approved):
LGTM! @pelikhan and @Jaqster is this okay to merge? It will show up in the dropdown list on the block.

Also, should we add eighth and quarter triplets? I think they are used a little more frequently. The standard labelling for those are "1/4t" and "1/8t". Could also do sixteenth but that might be overkill...

Jaqster:
Forgive my musical ignorance, but is a triplet a 1/3 beat?
While we're doing this, can we also list them in order in the dropdown?
image

blaquezzz (approved):

pelikhan:
@abchatra next release?

PR #1209, opened 2021-01-04
sprites shouldn't float due to lag

jwunderl:
Fix https://github.com/microsoft/pxt-arcade/issues/2320
fix https://github.com/microsoft/pxt-arcade/issues/2744

When there's a lot of lag and sprites have high speeds, it's possible to overflow fx8 in this calculation (or underflow it when the calculation is flipped). Replaces this one with a normal floating point normal mul / divide so that doesn't happen

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -277,27 +277,8 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         sprite._vx = this.constrain(sprite._vx);
         sprite._vy = this.constrain(sprite._vy);
 
-        const dx = Fx.idiv(
-            Fx.mul(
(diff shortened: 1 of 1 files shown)


riknoll:
LGTM! Any noticeable change on hardware? I think we should to merge this either way but i'm curious.

riknoll (approved):

jwunderl:
Nothing noticeable from my tests, the part changed here runs once per sprite per frame, and the vast majority of the time is spent later in the physics checking collisions / overlaps. It looks like the operations take about 3x as long (rough comparison of running 10,000,000 iterations of the calculations), but still on the order of ~0.003ms or so

PR #1211, opened 2021-01-05
fix ask block inside on a / b pressed controller event

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2738
fix https://github.com/microsoft/pxt-arcade/issues/1639

Someday we should probably see about better handling for controller events creating new scenes / maybe think on the queuing behavior~


riknoll (approved):

riknoll:
Is this is a problem with our other dialogs? like show long text or ask for string

jwunderl:
The other ones all handle it properly already~

livcheerful (approved):

PR #1215, opened 2021-01-19
Add animation blocks by default

jwunderl:
Move the 'new' animation blocks in as a default blocks, top of advanced:

alwaysAnimations

The animations extension now just adds in the 'legacy' blocks. I can rename the extension if we want, but that ends up being a bit tedious as built in extensions use the name from pxt.json across the editor (e.g. for finding the path for the icon, in the compiler, etc) and would require a change in all existing docs.

close https://github.com/microsoft/pxt-arcade/issues/2903


Jaqster:
Looks good Joey.  Can we at least update the extension description?  Make sure we specify that these are older, legacy blocks?  I'm anticipating there will be a bit of confusion between the built-in and extension.  Especially for folks who were used to using the extension...

jwunderl:
Yes, I changed it a bit here https://github.com/microsoft/pxt-common-packages/pull/1215/files#diff-5dd59269af30bfdb230293ba8d0d31919cf60eb7e436cd5932404f3855425967R3

Maybe adding legacy would be enough?

```
"Legacy state based animations for sprites"
```

Jaqster:
Yes, I think that would be helpful.  By the way, what do you think about fleshing out the https://arcade.makecode.com/extensions page as a way to give people more information about all the different extensions available without having to go into the editor?  Like we do on https://makecode.microbit.org/extensions?  @ganicke?

jwunderl:
That definitely sounds useful! For what it's worth a community member on the forum started their own gist like that: [gist](https://gist.github.com/UnsignedArduino/3bc2adce4b3f687ea1967578449d014d) [src]( https://forum.makecode.com/t/extensions/5006/4?u=jwunderl)

ganicke:
Yeah @jwunderl, we can beef up the [extensions](https://arcade.makecode.com/extensions) page and add a library of selected ones.

abchatra:
I think it is time to remove "legacy" and say "State based". We added "legacy" thinking we will deprecate in six months or so. It is been over an year and half and we still plan to support it as folks find it useful.  In my opinion this is going to stay forever.  What do you all think?

Jaqster:
We used to have all animations in an extension.  Then we moved some of these blocks out of the extension and into the editor.  But the extension is still there.  This is going to be confusing for a while.  The more we can differentiate the built-in blocks from the extension, the better.  Maybe we can change the image on the extension if we can't change the name?  Just changing the description to "state-based animations" is not going to be enough IMO.

jwunderl:
Okay, all up to date with switch from legacy -> advanced ~

livcheerful (approved):

PR #1217, opened 2021-01-20
Updated pull request (1207) platform.h

lalauria2000:
on behalf of @wallarug. Part 3 of 3.
changes for the platform.h in common packages
See: https://github.com/microsoft/pxt-common-packages/pull/1207
For rest of files: https://github.com/microsoft/pxt-maker/pull/329

### DIFF
--- a/libs/core---samd/platform.h
+++ b/libs/core---samd/platform.h
@@ -135,4 +135,73 @@ typedef int PinName;
 #define PB30 62
 #define PB31 63
 
+
+// 128 pin packages
+#ifdef SAMD51
+#define PC00  64
+#define PC01  65
+#define PC02  66
+#define PC03  67
+#define PC04  68
+#define PC05  69
+#define PC06  70
+#define PC07  71
+#define PC08  72
+#define PC09  73
+#define PC10  74
+#define PC11  75
+#define PC12  76
+#define PC13  77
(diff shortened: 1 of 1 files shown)


libs/core---samd/platform.h
```diff
@@ -135,4 +135,73 @@ typedef int PinName;
 #define PB30 62
 #define PB31 63
 
+
+// 128 pin packages
+#ifdef SAMD51
```
pelikhan, line 140: @mmoskal i assume we always want those so probably remove this #ifdef

mmoskal: yes

mmoskal (approved):

mmoskal:
Thank you!

lalauria2000:
you're welcome! I was wondering, does dal.d.ts need to be updated with PC and PD ports too?

mmoskal:
well, yes, but that happens in particular editors (pxt-maker, pxt-arcade etc)

lalauria2000:
okay got it. Whenever you have time please check this with @pelikhan
https://github.com/microsoft/pxt-maker/pull/329  which is the follow up to this issue 
https://github.com/microsoft/pxt-maker/issues/328
Thanks!

PR #1219, opened 2021-01-20
Split wall & tile flag, add top level blocks for flags in tutorials

jwunderl:
A decent amount of the changes are whitespace, I'll add some annotations where I made actual changes & here's an 'ignore whitespace' view: https://github.com/microsoft/pxt-common-packages/pull/1219/files?diff=unified&w=1

Separate ghost flags even more, so walls and tile overlaps are handled by separate flags. @riknoll @shakao @livcheerful question on behavior for an edge case; if a sprite is set to 'ghost through walls' and has an 'on overlap tile' event, if it overlaps a wall should that trigger? Currently we don't trigger overlap events with walls (as you shouldn't overlap with a wall / it should push you back), but I'm leaning towards making it so when you're a wall ghost you can overlap those when you can walk through them. (this commit is the change for this behavior: https://github.com/microsoft/pxt-common-packages/pull/1219/commits/945cd2b5bd70bf453a455b045303c4ba6cc105e5)

For the blocks to separate out, I checked which flags are used in docs/tutorials and it matched what we expected (plus auto destroy, which I think is a good one as well):

image

image

and here's a build in case anyone wants to test, my linked blockly is a little out of date so the text input bug is still there:

https://arcade.makecode.com/app/ebcd7e4adbfc152f68fa666eefa50ac2c445b000-e509f9a3d2

and a quick bad sample game for that: https://makecode.com/_3TDTLiWV888A

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -570,7 +570,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                         tileScale
                     );
 
-                    // if the sprite can can move through walls, it can overlap the underlying tile.
+                    // if the sprite can move through walls, it can overlap the underlying tile.
                     if (!tm.isObstacle(x0, y0) || !!(s.flags & sprites.Flag.GhostThroughWalls)) {
                         overlappedTiles.push(tm.getTile(x0, y0));
                     }
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -570,7 +570,8 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                         tileScale
                     );
 
-                    if (!tm.isObstacle(x0, y0)) {
+                    // if the sprite can can move through walls, it can overlap the underlying tile.
+                    if (!tm.isObstacle(x0, y0) || !!(s.flags & sprites.Flag.GhostThroughWalls)) {
                         overlappedTiles.push(tm.getTile(x0, y0));
                     }
                 }
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -24,9 +24,11 @@ class PhysicsEngine {
 }
 
 const MAX_TIME_STEP = 100; // milliseconds
-const SPRITE_CANNOT_COLLIDE = SpriteFlag.NoTileCollisions | sprites.Flag.Destroyed | SpriteFlag.RelativeToCamera;
-const SPRITE_CANNOT_OVERLAP = SpriteFlag.NoSpriteOverlaps | sprites.Flag.Destroyed | SpriteFlag.RelativeToCamera;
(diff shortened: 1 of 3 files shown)


libs/game/physics.ts
```diff
@@ -373,215 +370,215 @@ class ArcadePhysicsEngine extends PhysicsEngine {
             s._lastY
         );
 
-        const overlappedTiles: tiles.Location[] = [];
```
jwunderl, line 376: We were making this early in the step for dx and populating it with tiles we 'ran into', but this is unnecessary as the portion at the bottom already checks the tiles the sprite is touching & we're only ever moving < 2px per step here anyways. (and the dy one was actually just keeping it in the scope of the dy movement, and not doing anything with the array)

libs/game/sprite.ts
```diff
@@ -726,10 +728,43 @@ class Sprite extends sprites.BaseSprite {
...
+        this.setFlag(SpriteFlag.BounceOnWall, on);
+    }
+
+    /**
+     * Set whether a sprite should be destroyed when it leaves the screen (on) or not (off)
+     */
+    //% group="Effects"
+    //% weight=20
+    //% blockId=spritesetsetautodestroy block="set %sprite(mySprite) auto destroy %on=toggleOnOff"
+    //% on.defl=true
```
jwunderl, line 758: I figured we might as well save a step in tutorials and default these to on, as they all default to off for normal sprites.

libs/game/physics.ts
```diff
@@ -352,9 +351,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         // if the sprite is already clipping into a wall,
         // allow free movement rather than randomly 'fixing' it
         if (s.flags & sprites.Flag.IsClipping) {
-            if (tm.isOnWall(s)) {
```
jwunderl, line 355: Technically this is a behavior change, but removing this portion makes it so we can still trigger overlaps with tiles if the sprite is clipping; I think this is the right behavior, no reason that a sprite being stuck in a wall should mean it doesn't take lava damage / etc

livcheerful (approved):

libs/game/physics.ts
```diff
@@ -373,215 +370,216 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
+                                hbox.bottom
+                            ),
+                            Fx.oneHalfFx8
+                        ),
+                        tileScale
+                    );
 
-                if (!tm.isObstacle(x0, y0)) {
-                    overlappedTiles.push(tm.getTile(x0, y0));
+                    // if the sprite can can move through walls, it can overlap the underlying tile.
```
riknoll, line 573: nitpick: "can can"

PR #1220, opened 2021-01-21
[stable8.4] Split wall & tile flag, add top level blocks for flags in tutorials (#1219)

jwunderl:
port #1219 and https://github.com/microsoft/pxt-common-packages/pull/1221 to stable


jwunderl:
Updated to remove autodestroy block, default dropdown to autodestroy:

image

PR #1226, opened 2021-01-29
Add effects to music "play sound" block

klausw:
These are intended to provide some additional options for games and provide inspiration for custom sounds.

Test program based on using a modified editor with this PR included: https://makecode.com/_4PfMrpP9bg0h

Adapted from this standalone demo: https://makecode.com/_2uac15Pjbig9

PR #1216, opened 2021-01-20
Update the sprite 'set-flag' ref page

ganicke:
Update the flags in the ref page for `sprite.setFlag()`. A new **Flags** section is added with descriptions and usage examples for each flag.

RE: https://github.com/microsoft/pxt-arcade/issues/2923

PR #1229, opened 2021-02-04
[stable8.4] Cherry picks for release

jwunderl:
* #1224 : github actions
* #1226: new sounds from klaus that are amazing
* #1216: ref page updates for flags; need release to pick up the two new help links, also needed for packaging into electron app anyways
* #1225: subtle issue where sprite remains in allOfKind temporarily when destroyed with effects; we ran into this one several times on stream
* #1223: fix friction calculation when vy > 0
* #1227: fix visual bug when drawing sprites partially offscreen on left / top with non-integer x/y; https://github.com/microsoft/pxt-arcade/issues/3032 hits this bug (it's what causes the shifting in the energy bars. Not strictly a regression, but the floating point fix caused this to occur slightly more often / be more common to hit
* #1228: the standalone sprite create block requested for course / ease of use


abchatra (approved):

abchatra:
Looks good to me. Feel free to merge & bump

PR #1233, opened 2021-02-25
improvements to VM for ESP32 support

mmoskal:
Add support for VM running on ESP32 from flash. Also:

* fix a bug where string concatenation would be completely broken in some cases (not sure if there's any bug on this; the "nice" thing is that only the first access to such concatenated string would point into garbage memory)
* use C++ template magic for .text-section string literals with VT pointers => also don't runtime-allocate string literals in VM
* add skiplist_packed string literals (take less space in flash)
* change `vtable` static type to be pointer (it usually is - helps with gdb)
* optimize `x/1`
* there's no longer pre-alloc phase for VM - a few arrays are allocated with regular malloc(), but no GC-able objects are created during VM image load
* the VM image can now be read-only (provided it has the right VT pointers set)


eanders-MS (approved):
:shipit:

PR #1235, opened 2021-03-01
Changes for vm operation

eanders-MS:
Changes to get pxt.dll working in Unity.

### DIFF
--- a/libs/core---vm/scheduler.cpp
+++ b/libs/core---vm/scheduler.cpp
@@ -97,8 +97,12 @@ static void panic_core(int error_code) {
 extern "C" void target_panic(int error_code) {
     panic_core(error_code);
 
+#if defined(PXT_VM)
+    systemReset();
+#else
     while (1)
         sleep_core_us(10000);
+#endif
 }
 
 DLLEXPORT int pxt_get_panic_code() {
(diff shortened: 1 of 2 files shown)


mmoskal (approved):
we'll need to change this when PXT64 is not defined (especially ESP32), but I'll take care of that together with a ton of other changes that are there already

libs/core---vm/scheduler.cpp
```diff
@@ -97,8 +97,12 @@ static void panic_core(int error_code) {
 extern "C" void target_panic(int error_code) {
     panic_core(error_code);
 
+#if defined(PXT_VM)
```
eanders-MS, line 100: The `while (1)` loop was locking pxt.dll, and would freeze Unity the next time it made a dll call. I'm not certain this is the best solution, but does improve things.

mmoskal: I think that's cool. My comment was about allocating 1M always :)

eanders-MS: For pxt.dll/.so/.dylib, could it just use the regular heap? No strict size limitation?

mmoskal: It could be it will be difficult to track which pointers are in heap (isReadOnly() intention). I would rather allocate 16M if needed. The memory situation is quite bi-modal - either you have ~100k or you have ~100M it seems.

PR #1238, opened 2021-03-16
add protected flash writes/erases for when SD is in use

jamesadevine:
Soft device requires control over everything...

### DIFF
--- a/libs/settings/NRF52Flash.cpp
+++ b/libs/settings/NRF52Flash.cpp
@@ -9,6 +9,30 @@ namespace codal {
 
 #define waitForLast() while ((NRF_NVMC->READY & NVMC_READY_READY_Msk) == 0);
 
+static volatile bool flash_op_complete = false;
+
+#ifdef SOFTDEVICE_PRESENT
+#include "nrf_sdh_soc.h"
+
+static void nvmc_event_handler(uint32_t sys_evt, void *)
+{
+    if (sys_evt == NRF_EVT_FLASH_OPERATION_SUCCESS)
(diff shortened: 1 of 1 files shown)


libs/settings/NRF52Flash.cpp
```diff
@@ -56,21 +98,40 @@ int ZFlash::writeBytes(uintptr_t dst, const void *src, uint32_t len) {
...
-            *dp++ = v;
-            waitForLast();
-        } else {
-            dp++;
+        while(1)
+        {
+            if (sd_flash_write((uint32_t *)dp, (uint32_t *)sp, len) == NRF_SUCCESS)
+                break;
+
+            system_timer_wait_ms(10);
```
mmoskal, line 111: I think this may be too slow for anything useful. Maybe you can spin on flash_op_complete?

jamesadevine: If the previous call does not go through then flash_op_complete will never trigger as the async callback will never take place. This is the same code that is used in microbit codal now!

mmoskal: so you're saying the wait_ms is usually not hit?

jamesadevine: From my observations during my primitive testing, I have not yet hit this condition. I have only ever gotten NRF_SUCCESS back from these calls.

mmoskal (approved):

jamesadevine:
@mmoskal Can we merge?

mmoskal:
I mean it was approved...

PR #1239, opened 2021-03-19
Fixes stop animation not working with legacy state animation

abchatra:
Simple fix

### DIFF
--- a/libs/game/animation.ts
+++ b/libs/game/animation.ts
@@ -567,7 +567,7 @@ namespace animation {
                 return true;
             });
         }
-        if (type == AnimationTypes.All) {
+        if (type == AnimationTypes.All || type == AnimationTypes.ImageAnimation) {
             //stop state based animation if any as well
             sprite._action = -1
         }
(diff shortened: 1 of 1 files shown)


abchatra:
Not sure fix is clean, but works.

abchatra:
Fixes https://github.com/microsoft/pxt-arcade/issues/3278

libs/game/animation.ts
```diff
@@ -567,6 +567,10 @@ namespace animation {
                 return true;
             });
         }
+        if (type == AnimationTypes.All) {
```
jwunderl, line 570:
```suggestion
        if (type == AnimationTypes.All || type == AnimationTypes.ImageAnimation) {
```

Feels like it should also probably apply when you pass `ImageAnimation`, as the options are movement vs picture changing?

jwunderl (approved):
Yeah this is the right fix ~ LGTM either way but I'd lean towards taking the suggestion for consistency~

libs/game/animation.ts
abchatra: Yes that would be better. Thanks!

PR #1241, opened 2021-03-25
Update example for 'place on random tile' page

ganicke:
Add a `jres` section and update example to use the defined tilemap.

Fixes https://github.com/microsoft/pxt-arcade/issues/3358


jwunderl (approved):
LGTM

PR #1252, opened 2021-04-28
define ble_running for non-microbit NRF-based boards

jamesadevine:
ble_running is defined by MicroBitDevice.h and everything compiles successfully. For other boards with softdevice, ble_running is now defined here.


jamesadevine:
@mmoskal What do you think? Define ble_running here? or in the CODAL target?

mmoskal (approved):
it's cool to do this here; also thanks for including my fixes! 

please merge when ready

PR #1254, opened 2021-05-09
refactor gc/perf functions

pelikhan:
Split control.cpp so that we can easily add the GC/perf counter functionalities in other targets like micro:bit.


libs/base/control.cpp
```diff
@@ -100,50 +100,4 @@ namespace control {
     void dmesgValue(TValue v) {
         anyPrint(v);
     }
-
```
pelikhan, line 103: moved to controlgc.cpp

libs/base/control.ts
```diff
@@ -64,18 +64,6 @@ namespace control {
     //% shim=pxtrt::panic
     export function panic(code: number) { }
 
-    /**
```
pelikhan, line 67: moved to perfcounters.ts

mmoskal (approved):

PR #1257, opened 2021-05-18
Add game.push/popScene (aliases to push/popEventContext)

mmoskal:
@eanders-MS feel free to add a little more here - whatever you think is generally useful and you need in kojac. With these two functions my jacdac-multitool (menu system) works fine with game---light

### DIFF
--- a/libs/game---light/compat.ts
+++ b/libs/game---light/compat.ts
@@ -8,8 +8,11 @@ namespace scene {
     export const SCREEN_CLEAR_PRIORITY = 1;
     export const UPDATE_INTERVAL_PRIORITY = 19;
     export const UPDATE_PRIORITY = 20;
-    export const ON_PAINT_PRIORITY = 100;
-    export const ON_SHADE_PRIORITY = 110;
+
+    export const PRE_RENDER_UPDATE_PRIORITY = 55;
+    export const RENDER_BACKGROUND_PRIORITY = 60;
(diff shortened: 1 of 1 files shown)


libs/game---light/compat.ts
```diff
@@ -28,6 +28,18 @@ namespace game {
         return _scene
     }
 
+    export function pushScene() {
```
eanders-MS, line 31: `pushScene` and `popScene` here might be confusing to some. Are they needed? Kojac calls `push/popEventContext` directly.

mmoskal: well, I wanted to have something cross-compatible between game and game---light, but maybe that should not be the goal?

eanders-MS: I lean toward keeping it out until there's a clear need. Are there any other missing APIs, or is this the only one?

mmoskal: For my use case it was just this one.

I think we should remove game.pushEventContext from game---light and replace it with game.pushScene. This way whatever works in game---light should work the same in standard game, allowing for UI library targeting both.

I don't think there's much risk of confusion here - the Scene class in game---light has almost no fields, and things like Sprites are missing altogether.

Though maybe the UI library could do its own drawing handlers and just use control.*EventContext...

eanders-MS: That makes sense.

eanders-MS (approved):

PR #1260, opened 2021-06-09
add image blit method

eanders-ms:
**For your consideration.** When working in JavaScript I often want to write directly to the screen rather than use the Sprite system, and have found myself implementing this `blit` method multiple times. The method implemented here is modeled after the WinGDI's venerated [`StretchBlt`](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-stretchblt).

This `blit` method copies an image from a source rectangle to a destination rectangle, stretching or compressing to fit the dimensions of the destination rectangle, if necessary.

If we think this is a useful addition to the runtime, then I'll implement the cpp side.

blit

### DIFF
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -1073,8 +1073,17 @@ void blit(Image_ dst, int xDst, int yDst, int wDst, int hDst, Image_ src, int xS
 }
 
 //%
-void _blit(Image_ dst, int xyDst, int whDst, Image_ src, int xySrc, int whSrc, bool transparent) {
-    blit(dst, XX(xyDst), YY(xyDst), XX(whDst), YY(whDst), src, XX(xySrc), YY(xySrc), XX(whSrc), YY(whSrc), transparent);
+void _blit(Image_ dst, Image_ src, RefCollection args) {
+    int xDst = toInt(args.getAt(0));
+    int yDst = toInt(args.getAt(1));
+    int wDst = toInt(args.getAt(2));
+    int hDst = toInt(args.getAt(3));
+    int xSrc = toInt(args.getAt(4));
+    int ySrc = toInt(args.getAt(5));
+    int wSrc = toInt(args.getAt(6));
+    int hSrc = toInt(args.getAt(7));
+    auto transparent = toBoolQuick(args.getAt(8));
+    blit(dst, xDst, yDst, wDst, hDst, src, xSrc, ySrc, wSrc, hSrc, transparent);
 }
 
 void fillCircle(Image_ img, int cx, int cy, int r, int c) {
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -1068,22 +1068,50 @@ void _blitRow(Image_ img, int xy, Image_ from, int xh) {
     blitRow(img, XX(xy), YY(xy), from, XX(xh), YY(xh));
 }
 
-void blit(Image_ dst, int xDst, int yDst, int wDst, int hDst, Image_ src, int xSrc, int ySrc, int wSrc, int hSrc, bool transparent) {
-    // TODO
+void blit(Image_ dst, Image_ src, pxt::RefCollection *args) {
+    int xDst = pxt::toInt(args->getAt(0));
+    int yDst = pxt::toInt(args->getAt(1));
+    int wDst = pxt::toInt(args->getAt(2));
+    int hDst = pxt::toInt(args->getAt(3));
+    int xSrc = pxt::toInt(args->getAt(4));
+    int ySrc = pxt::toInt(args->getAt(5));
+    int wSrc = pxt::toInt(args->getAt(6));
+    int hSrc = pxt::toInt(args->getAt(7));
+    bool transparent = pxt::toBoolQuick(args->getAt(8));
+
+    if (xSrc < 0 || ySrc < 0 || wSrc <= 0 || hSrc <= 0 ||
+        xSrc > src->width() || ySrc > src->height() ||
+        xSrc + wSrc > src->width() || ySrc + hSrc > src->height()) {
+        // Would read outside src image
+        return;
+    }
+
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
If nothing else, this would be amazing for users for simple 'intro cutscenes' where the main character is 'teleported to a new world' -- a series of stretches like these ones witha portal-ish background is a pretty common trope for that sort of thing and always looks fun :)

Should be able to do the stretching part with a screen effect even to make it super easy https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/effects.ts#L3

libs/screen/image.cpp
```diff
@@ -1068,6 +1068,15 @@ void _blitRow(Image_ img, int xy, Image_ from, int xh) {
     blitRow(img, XX(xy), YY(xy), from, XX(xh), YY(xh));
 }
 
+void blit(Image_ dst, int xDst, int yDst, int wDst, int hDst, Image_ src, int xSrc, int ySrc, int wSrc, int hSrc, bool transparent) {
+    // TODO
+}
+
+//%
+void _blit(Image_ dst, int xyDst, int whDst, Image_ src, int xySrc, int whSrc, bool transparent) {
```
mmoskal, line 1076: there is 4 argument limit on C++ functions (which is why there's this XX/YY macros)

I think the fastest way to pass the arguments would be an array (you can allocate just one array); probably just stick the numbers in the array and pass the images separately

mmoskal (approved):
great idea! there's already some code that does this for a single column (used in ray casting), but a more generic one is useful as well!

libs/screen/image.cpp
eanders-MS: I was wondering why these parameters were packed!

eanders-MS: @mmoskal How do I test the cpp side? Does it need to round-trip to the compiler backend? Use uploadtrg?

mmoskal: ARM passes first 4 args in registers; I never implemented passing args on stack...

riknoll: @eanders-MS you can just test it using local serve! Just don't use --rb

mmoskal: I typically build from command line with `pxt` in `pxt-arcade/projects/myproject`; you need something like the following in `pxt.json`: 
```json
    "dependencies": {
        "hw---stm32f401": "file:../../libs/hw---stm32f401",
        "device": "file:../../libs/device"
    },
```

I have the following env variables set:
```json
declare -x PXT_ASMDEBUG="1"
declare -x PXT_COMPILE_SWITCHES="size"
declare -x PXT_FORCE_LOCAL="yes"
declare -x PXT_NODOCKER="yes"
declare -x PXT_RUNTIME_DEV="yes"
```

Check out `pxt help` for descriptions.

eanders-MS: Where would I declare the static array? I don't quite see how it would be shared between TS and c++ implementations.

mmoskal: Declare in TypeScript and pass as argument of type RefCollection to c++

libs/screen/image.ts
```diff
@@ -126,14 +126,25 @@ namespace helpers {
...
 
     //% shim=ImageMethods::_blit
-    declare function _blit(img: Image, xyDst: number, whDst: number, src: Image, xySrc: number, whSrc: number, transparent: boolean): void;
+    declare function _blit(img: Image, src: Image, args: any[]): void;
 
     function pack(x: number, y: number) {
         return (Math.clamp(-30000, 30000, x | 0) & 0xffff) | (Math.clamp(-30000, 30000, y | 0) << 16)
     }
 
+    const _blitArgs: any[] = [];
```
eanders-MS, line 135: @mmoskal something like this? I'm not having luck compiling for native yet.

mmoskal: I thought sim also uses RefCollection but I may be wrong. Cpp looks good!

eanders-ms:
@jwunderl great idea about the cutscene effect. What parameters would you like to see? Can you point me to an example?

jwunderl:
The 'normal' screen effects that go in the drop down don't take parameters, but we can wrap it in a block in an extension to make it customizable -- closest would probably be the image effect 'melt' which is the one that is used in the game over screen: https://github.com/microsoft/pxt-common-packages/blob/7629e5cc88ea7a14f008cd7c92e776ff81543143/libs/game/effects.ts#L73 -- if you were to make one in the same way as that & wrap it in a block in an extension that applies it every frame to start we could make something fun (but that one is just an image effect -- screen effects are currently just the particle effects that to the whole screen https://github.com/microsoft/pxt-common-packages/blob/7629e5cc88ea7a14f008cd7c92e776ff81543143/libs/game/particleeffects.ts#L332)

With that said, that shouldn't block this pr / is only supplementary and can go in an extension~

jwunderl (approved):

libs/screen/image.cpp
```diff
@@ -1068,22 +1068,50 @@ void _blitRow(Image_ img, int xy, Image_ from, int xh) {
...
+    int hDst = pxt::toInt(args->getAt(3));
+    int xSrc = pxt::toInt(args->getAt(4));
+    int ySrc = pxt::toInt(args->getAt(5));
+    int wSrc = pxt::toInt(args->getAt(6));
+    int hSrc = pxt::toInt(args->getAt(7));
+    bool transparent = pxt::toBoolQuick(args->getAt(8));
+
+    if (xSrc < 0 || ySrc < 0 || wSrc <= 0 || hSrc <= 0 ||
+        xSrc > src->width() || ySrc > src->height() ||
+        xSrc + wSrc > src->width() || ySrc + hSrc > src->height()) {
```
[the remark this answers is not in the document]
eanders-ms: Good point! Changed it to clamp the ranges.

PR #1261, opened 2021-06-18
track revision count on images to fix perf for sprites mutated at runtime

jwunderl:
For games like https://forum.makecode.com/t/how-to-debug-performance/8681, and I remember we had a forum game a while ago where they had a screen sized image that was causing horrible perf issues that was probably hitting this.

The issue is that when a sprite is mutated, we need to update the hitbox to avoid it passing through walls / other sprites when the image is mutated underneath it. Doing this constantly is a huge perf drain though, as we have to identify the bounding box on the sprite. This is tracking a revision count to identify when the hitbox actually needs to be updated. Bumps the fps in the linked game from about 10-12 to 16-18 on meowbit.

https://arcade.makecode.com/app/096ee709f0b94bed3d6a9440ee445f1d1c2b68d7-ff832148ba#pub:_c6PRXUCV2c2Y


libs/base/pxtbase.h
```diff
@@ -815,6 +815,7 @@ struct ImageHeader {
 
 class RefImage : public RefObject {
   public:
+    uint16_t revision;
```
jwunderl, line 818: I guess could just be uint8_t now that I'm thinking of it, since it only matters that it matches / if it overflows back to 0 it'd be fine unless there are exactly 256 mutations separate mutations on the sprite within a single frame

mmoskal: it doesn't matter - allocations are word-aligned; probably best to make it uint32_t to be safe and avoid weird bugs

however do move it after the buffer field

libs/screen/sim/image.ts
```diff
@@ -5,6 +5,7 @@ namespace pxsim {
         _bpp: number;
         data: Uint8Array;
         isStatic = true;
+        revision: number;
 
         constructor(w: number, h: number, bpp: number) {
```
mmoskal, line 10: this.revision = 0 missing?

mmoskal (approved):

riknoll (approved):

livcheerful (approved):

PR #1262, opened 2021-06-24
[stable8.8] track revision count on images to fix perf for sprites mutated at run…

jwunderl:
port #1261

significant perf improvement (primarily on hardware) when game includes mutated images

(also need https://github.com/microsoft/pxt-common-packages/pull/1259 right?)

### DIFF
--- a/libs/base/pxtbase.h
+++ b/libs/base/pxtbase.h
@@ -818,6 +818,7 @@ struct ImageHeader {
 class RefImage : public RefObject {
   public:
     BoxedBuffer *buffer;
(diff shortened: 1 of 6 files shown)


abchatra:
We have a arcade release in a month. Can this bake and go into the next release?

jwunderl:
Richard mentioned in standup this might be a good one to port, since it's small but has a massive effect on a lot of games; I'm fine with waiting for the release if that's preferred though

abchatra:
Lets wait for now.

PR #1264, opened 2021-07-16
Check if a sprite was destroyed before registering another obstacle

dmonroym:
Fixes https://github.com/microsoft/pxt-arcade/issues/2053

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -428,7 +428,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                     );
 
                     for (const tile of collidedTiles) {
-                        if(!(s.flags & sprites.Flag.Destroyed)) {
+                        if(!(s.flags & SPRITE_NO_WALL_COLLISION)) {
                             s.registerObstacle(collisionDirection, tile, tm);
                         }
                     }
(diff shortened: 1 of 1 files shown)


libs/game/physics.ts
```diff
@@ -428,7 +428,9 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                     );
 
                     for (const tile of collidedTiles) {
-                        s.registerObstacle(collisionDirection, tile, tm);
+                        if(!(s.flags & sprites.Flag.Destroyed)) {
```
jwunderl, line 431:
```suggestion
                        if(!(s.flags & SPRITE_NO_WALL_COLLISION)) {
```

Ahh I missed this in the message earlier, it should also bail out here in the same way if it became a ghost

libs/game/physics.ts
```diff
@@ -507,7 +509,9 @@ class ArcadePhysicsEngine extends PhysicsEngine {
...
```
jwunderl, line 512:
```suggestion
                        if(!(s.flags & SPRITE_NO_WALL_COLLISION)) {
```

jwunderl (approved):
Nice! suggestion to switch to a different const for the check to bail out but besides that good to go

livcheerful (approved):

PR #1266, opened 2021-07-27
Add shadow to vx and vy on moveSprite block so it shows field editor

dmonroym:
Fixes https://github.com/microsoft/pxt-arcade/issues/1616

Adds the field editor to the "Move mySprite with buttons" block

image

### DIFF
--- a/libs/game/controller.ts
+++ b/libs/game/controller.ts
@@ -42,7 +42,7 @@ namespace controller {
     }
 
     export function _moveSprites() {
-        // todo: move to currecnt sceane
+        // todo: move to current scene
         control.enablePerfCounter("controller")
         players().forEach(ctrl => ctrl.__preUpdate());
     }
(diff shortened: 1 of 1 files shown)


libs/game/controller.ts
```diff
@@ -162,6 +162,8 @@ namespace controller {
         //% vx.defl=100 vy.defl=100
         //% help=controller/move-sprite
         //% group="Multiplayer"
+        //% vx.shadow="speedPicker"
+        //% vy.shadow="speedPicker"
```
jwunderl, line 166:
```suggestion
        //% vx.shadow="spriteSpeedPicker"
        //% vy.shadow="spriteSpeedPicker"
```

I believe to show px/s?

jwunderl (approved):
Nice! One suggestion / clarification~

PR #1276, opened 2021-09-02
more azure iot twin support

mmoskal:
plus fixes for net.getJSON/Text and color-coding for console.log

### DIFF
--- a/libs/base/sim/control.ts
+++ b/libs/base/sim/control.ts
@@ -160,14 +160,12 @@ namespace pxsim.control {
     }
 
     export function __log(priority: number, str: string) {
-        let prefix = "";
         switch (priority) {
-            case 0: prefix = "d>"; break;
-            case 1: prefix = "l>"; break;
-            case 2: prefix = "w>"; break;
-            case 3: prefix = "e>"; break;
+            case 0: console.debug("d>" + str); break;
+            case 1: console.log("l>" + str); break;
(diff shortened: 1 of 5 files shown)


libs/base/sim/control.ts
```diff
@@ -160,14 +160,12 @@ namespace pxsim.control {
...
-        let prefix = "";
         switch (priority) {
-            case 0: prefix = "d>"; break;
-            case 1: prefix = "l>"; break;
-            case 2: prefix = "w>"; break;
-            case 3: prefix = "e>"; break;
+            case 0: console.debug("d>" + str); break;
+            case 1: console.log("l>" + str); break;
+            case 2: console.warn("w>" + str); break;
+            case 3: console.error("e>" + str); break;
```
mmoskal, line 167: @pelikhan any reason this would be problematic?

pelikhan:
no idea

libs/base/sim/control.ts
mmoskal: @riknoll @jwunderl @shakao  @abchatra - any reason sending messages to console not always with console.log but with console.warn etc would be problematic? it will give better experience when looking at JS console when running sim

jwunderl: Seems fine to me~

abchatra: looks good to me.

PR #1279, opened 2021-09-14
Add new tilemap blocks and reorganize the scene category

riknoll:
This PR does a few things:

1. Ports several blocks from the tilemap extension into the scene category
2. Reorders the scene groups. Mostly just swapped the "Camera" and "Effects" groups
3. Reordered the blocks in the "Camera" group. We never organized them in the first place; they were in random order. Now "camera follow sprite" is on top
4. Created some new groups for the tilemap-related blocks. They are "Tilemaps", "Operations", and "Locations"
5. Adds "destroy all sprites with effect" to the sprites category
6. Deprecates the old "set tilemap to" block and replaces it with a new one with a shadow block

None of these changes require upgrade rules.

There is one "gotcha" if you use this with the existing tilemaps extension where the tilemap shadow blocks from that extension are not compatible with the new blocks in the scene category. I can push a change to the tilemaps extension to fix this (that would require an upgrade rule, but just one that bumps the extension version).

Here is the "destroy all sprites" block:

Here is how the toolbox looks now:

Screen Shot 2021-09-14 at 10 36 37 AM
Screen Shot 2021-09-14 at 10 16 52 AM
Screen Shot 2021-09-14 at 10 16 59 AM
Screen Shot 2021-09-14 at 10 17 13 AM

And here's an upload target: https://arcade.makecode.com/app/4ef11d7a279b78f072c186abf39a614d0b7df901-64b8e2dee4#editor

### DIFF
--- a/libs/game/assetTemplates.ts
+++ b/libs/game/assetTemplates.ts
@@ -1,5 +1,6 @@
 //% helper=getTilemapByName
 //% pyConvertToTaggedTemplate
+//% blockIdentity="tiles._tilemapEditor"
 function tilemap(lits: any, ...args: any[]): tiles.TileMapData { return null }
 
 //% helper=getTilemapByName
@@ -17,6 +18,7 @@ function tilemap32(lits: any, ...args: any[]): tiles.TileMapData { return null }
 namespace assets {
     //% helper=getTilemapByName
     //% pyConvertToTaggedTemplate
+    //% blockIdentity="tiles._tilemapEditor"
     export function tilemap(lits: any, ...args: any[]): tiles.TileMapData { return null }
 
     //% helper=getImageByName
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -597,7 +597,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
         const alreadyHandled: tiles.Location[] = [];
 
         for (const tile of overlappedTiles) {
-            if (alreadyHandled.some(l => l.col === tile.col && l.row === tile.row)) {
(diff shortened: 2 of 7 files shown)


jwunderl (approved):

abchatra (approved):

eanders-MS (approved):

Jaqster:
Can we bump the Effects category above Camera?  They were the 2nd section before, and quite widely used.

Jaqster:
I would also rename "Operations" to "Tilemap Operations"

Jaqster:
Also, why is the Array of Tile Locations block not in the Locations category?  I also feel like the Is mySprite Hitting Wall Left boolean block seems like it might fit under Locations better too... seems similar to Tile to Left of mySprite block.

Jaqster:
I'm also thinking about how do we make it clear through these blocks that location = tilemap col, row....
This is a breaking change, so we probably wouldn't do this now, but what if we used "location col, row" instead of "tilemap col, row".  That might make it more clear?

livcheerful (approved):

dmonroym (approved):

riknoll:
@jaqster I moved up the effects category, but I want to put it after camera. I would argue that "camera follow sprite" is more widely used than the screen effects. I made all of the other changes you mentioned.

Here's the toolbox now:

Jaqster:
Looks really great!  Let's pull the numbers for Camera vs. Effects as a matter of due diligence.

abchatra:
@riknoll what is the status of this PR?

abchatra:
@riknoll ping!

riknoll:
Sorry, yes, this is good to go!

PR #1280, opened 2021-09-17
Use min and max implementation from the pxt namespace

nedseb:
Inside the file `pxtbase.h`, the function `min`, `max` and `swap` are redefined inside the namespace `pxt`. 
If you make use of a recent version of GCC or clang, there is an ambiguity with the standard definition.

To remove this ambiguity, I have just prefixed the use of these three functions to be sure to use only the one from the namespace `pxt`


mmoskal (approved):
Thank you, looks good!

pelikhan:
@mmoskal ready to merge?

PR #1301, opened 2021-11-22
Update 'placeOnRandomTile' ref page

ganicke:
This is a typo fix for https://github.com/microsoft/pxt-arcade/issues/3881 along with a some updates for the new api. This really should be part of the fixes need later for https://github.com/microsoft/pxt-arcade/issues/2728.

**Note:** The "See also" link is stale but I'll leave it in until the new 'set' and 'place' ref docs are in.

Fixes https://github.com/microsoft/pxt-arcade/issues/3881

### DIFF
--- a/libs/game/docs/reference/scene/place-on-random-tile.md
+++ b/libs/game/docs/reference/scene/place-on-random-tile.md
@@ -3,7 +3,7 @@
 Move a sprite's position to the center of a random tile in the scene.
 
 ```sig
-scene.placeOnRandomTile(null, 0)
+tiles.placeOnRandomTile(null, null)
 ```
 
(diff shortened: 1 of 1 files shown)


libs/game/docs/reference/scene/place-on-random-tile.md
```diff
@@ -6,60 +6,114 @@ Move a sprite's position to the center of a random tile in the scene.
 scene.placeOnRandomTile(null, 0)
 ```
 
-You a can make a sprite to locate itself right on top of a random tile in the tile map. Use the color number of type of tile to select for the sprite to locate on one of them.
-
+You can make a sprite locate itself right on top of a random tile in the tile map. Use the color number of type of tile to select for the sprite to locate on one of them.
```
shakao, line 9: (for the next change) i guess we should eventually take out all the color number stuff since that doesn't apply to the current tilemap blocks?

shakao (approved):

libs/game/docs/reference/scene/place-on-random-tile.md
ganicke: Actually, I wanted to change that paragraph too. I forgot to do that yesterday. Thanks.

PR #1308, opened 2021-12-13
Add browser text to speech library

riknoll:
Adds text to speech support in the browser simulator.

### DIFF
--- a/libs/text-to-speech/sim/tts.ts
+++ b/libs/text-to-speech/sim/tts.ts
@@ -1,10 +1,8 @@
 namespace pxsim.tts {
-    //%
     export function _getLanguageCode() {
         return window.navigator.language;
     }
 
-    //% promise
     export function _speakAsync(text: string, pitch?: number, rate?: number, volume?: number, language?: string, onStart?: RefAction, onBoundary?: RefAction): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             const utterance = new SpeechSynthesisUtterance(text);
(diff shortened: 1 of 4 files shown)


libs/text-to-speech/sim/tts.ts
```diff
@@ -0,0 +1,78 @@
...
+    //%
+    export function _getLanguageCode() {
+        return window.navigator.language;
+    }
+
+    //% promise
+    export function _speakAsync(text: string, pitch?: number, rate?: number, volume?: number, language?: string, onStart?: RefAction, onBoundary?: RefAction): Promise<void> {
+        return new Promise((resolve, reject) => {
+            const utterance = new SpeechSynthesisUtterance(text);
+            utterance.voice = getVoiceForLanguage(language || _getLanguageCode());
```
jwunderl, line 11: Is it good to default to `window.navigator.language`? It definitely seems good for ease of use but it feels like it could make shared games messed up / I guess I'm just not sure what behavior would be expected, feel free to ignore

jwunderl (commented):
Can we add a cpp that does no-ops for all browser calls? I think it'll just immediately crash on hw as is right?

Also not sure right away why build is failing >>

libs/text-to-speech/sim/tts.ts
riknoll: I'm thinking of this as more of an accessibility feature than a game feature, so I think system language is the right thing to do here because i believe that's how screen readers work. I did add a string argument to hardcode it to a specific iso language code (not super discoverable though)

riknoll:
The build was failing because of the `//%` annotations in the simulator

jwunderl (approved):

dmonroym (approved):

riknoll:
@jwunderl I might add c++ shims later, but I'm not going to do it now because I don't want this to trigger a cloud compile when added to a project.

pelikhan:
Eventually c++ could be provided by jacdac. We already support tts service.

PR #1311, opened 2022-01-23
Add sprite scaling support

eanders-ms:
Adds scale properties to Sprite class:
- `sx (scale x)`
- `sy (scale y)`
- `scale` (returns greater of sx and sy)

Adds a few built-in basic scaling blocks:
image

And adds additional blocks for advanced scaling in an extension:
image

image

### DIFF
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -73,8 +73,8 @@ class Sprite extends sprites.BaseSprite {
     _fy: Fx8 // friction
     _sx: Fx8 // scale
     _sy: Fx8 // scale
-    _width: Fx8
-    _height: Fx8
+    _width: Fx8 // scaled width
+    _height: Fx8 // scaled height
 
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x" callInDebugger
@@ -632,7 +632,16 @@ class Sprite extends sprites.BaseSprite {
         if (!this.isScaled())
             screen.drawTransparentImage(this._image, l, t);
         else
-            screen.blit(l, t, this.width, this.height, this._image, 0, 0, this._image.width, this._image.height, true, false);
+            screen.blit(
+                // dst rect in screen
+                l, t,
+                this.width,
+                this.height,
+                // src rect in sprite image
+                this._image,
+                0, 0,
+                this._image.width, this._image.height,
+                true, false);
 
         if (this.flags & SpriteFlag.ShowPhysics) {
             const font = image.font5;
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -361,7 +361,7 @@ class Sprite extends sprites.BaseSprite {
     }
 
     private isScaled(): boolean {
-        return !(this._sx === Fx.oneFx8 || this._sy === Fx.oneFx8);
+        return this._sx !== Fx.oneFx8 || this._sy !== Fx.oneFx8;
     }
 
     //% group="Physics" blockSetVariable="mySprite"
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -109,7 +109,7 @@ class Sprite extends sprites.BaseSprite {
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x" callInDebugger
     get x(): number {
-        return Fx.toFloat(this._x) + (this.width / 2)
+        return Fx.toFloat(Fx.add(this._x, Fx.div(this._width, Fx.twoFx8)));
     }
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x"
@@ -120,7 +120,7 @@ class Sprite extends sprites.BaseSprite {
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="y" callInDebugger
     get y(): number {
-        return Fx.toFloat(this._y) + (this.height / 2)
+        return Fx.toFloat(Fx.add(this._y, Fx.div(this._height, Fx.twoFx8)));
     }
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="y"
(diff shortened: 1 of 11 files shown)


libs/game/sprite.ts
```diff
@@ -71,27 +101,31 @@ class Sprite extends sprites.BaseSprite {
...
+    _sx: Fx8 // scale
+    _sy: Fx8 // scale
+    _width: Fx8 // scaled width
+    _height: Fx8 // scaled height
 
     //% group="Physics" blockSetVariable="mySprite"
     //% blockCombine block="x" callInDebugger
     get x(): number {
-        return Fx.toFloat(this._x) + (this._image.width / 2)
+        return Fx.toFloat(this._x) + (this.width / 2)
```
jwunderl, line 112: we might wanna just do these calculations with fx8 since we have _width as fx8 now but doesn't matter too much

libs/game/sprite.ts
```diff
@@ -259,6 +333,15 @@ class Sprite extends sprites.BaseSprite {
     setImage(img: Image) {
         if (!img) return; // don't break the sprite
         this._image = img;
+        this.recalcSize();
```
jwunderl, line 336: might be worth wrapping in a img !== this._image check now since we're doing a decent amount of work in the case where the image is being trivially reset

jwunderl (approved):
Code changes look good to me (few nits / should change % to $ in block authoring), but we'll need to make sure to remember to compare perf on hardware before we release this to make we're not degrading perf too much / noticeably (I think it'll be fine looking at it, but lots of the changes are near the hotspots in the physics which can be VERY perf sensitive).

Also adding michal as a triple check on the c++ :)

libs/game/sprite.ts
```diff
@@ -939,6 +1083,90 @@ class Sprite extends sprites.BaseSprite {
         }
     }
 
+    setScaleCore(sx?: number, sy?: number, anchor?: ScaleAnchor, proportional?: boolean): void {
+        anchor = anchor || ScaleAnchor.Middle;
+
+        const hasSx = typeof sx === 'number';
```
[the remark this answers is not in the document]
mmoskal: `x != null` would be faster than `typeof x === "number"` but not by much

libs/game/sprite.ts
```diff
@@ -237,6 +309,8 @@ class Sprite extends sprites.BaseSprite {
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._vy)); k += 2;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._ax)); k += 2;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._ay)); k += 2;
+        buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._sx)); k += 2;
```
mmoskal, line 312: I think you need to increase the +12 to +16 above?

mmoskal (approved):

libs/game/sprite.ts
```diff
@@ -939,6 +1083,90 @@ class Sprite extends sprites.BaseSprite {
...
+            const newH = this.height;
+            const diff = newH - oldH;
+            const diffOver2 = (diff / 2) | 0;
+            if (anchor & ScaleAnchor.Top) { this.y += diffOver2; }
+            if (anchor & ScaleAnchor.Bottom) { this.y -= diffOver2; }
+        }
+    }
+
+    //% blockId=sprite_set_scale
+    //% block="set %sprite(mySprite) scale to $value, anchor on $anchor"
```
jwunderl, line 1129: oh, one last one, I think I might lean more towards `"set %sprite(mySprite) scale to $value anchored on $anchor"` as I don't recall any other blocks using punctuation as much, but that's just my opinion / if you feel it's better as is that's probably fine~

libs/game/sprite.ts
eanders-ms: good catch!

libs/game/sprite.ts
```diff
@@ -259,6 +333,15 @@ class Sprite extends sprites.BaseSprite {
     setImage(img: Image) {
         if (!img) return; // don't break the sprite
         this._image = img;
+        this.recalcSize();
+    }
+
+    calcDimensionalHash() {
+        return Fx.mul(Fx.mul(this._width, this._height), Fx8(this._image.revision()));
```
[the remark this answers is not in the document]
eanders-ms: True, but I'd like to keep it in for futureproofing. It seems like an implementation detail of Sprite that could change, and unknowingly break collision on scaled sprites.

I made `_image` private again. That we me being sloppy.

eanders-ms:
I'm going to test with a bunch of games before merging.

jwunderl:
Sounds good, there are a few good ones under https://arcade.makecode.com/physics-tests that don't require user input / are decent for testing fps ~

eanders-ms:
Merging, but there might be a regression when hitting walls and velocity is non-zero. I noticed some sprite jittering in Tiny Wizards, but unable to make a simple repro.

libs/game/sprite.ts
```diff
@@ -311,19 +394,29 @@ class Sprite extends sprites.BaseSprite {
...
+    private isScaled(): boolean {
+        return this._sx !== Fx.oneFx8 || this._sy !== Fx.oneFx8;
+    }
+
     //% group="Physics" blockSetVariable="mySprite"
-    //% blockCombine block="width"
+    //% blockCombine block="width" callInDebugger
     get width() {
-        return this._image.width
+        return Fx.toInt(this._width);
```
eanders-ms, line 410: @jwunderl I think a float return type would be more appropriate here, otherwise `s.right - s.left` would not necessarily equal `s.width`. I wonder what you think. Worth "fixing"?

jwunderl: Hm, I unless there's a subtlety that I'm missing with regards to the sprite scaling then this value should always be be an integer, no? Or do we count a 15x15 @ 1.5x scale as 22.5x22.5 pixels even though it the pixels will be rounded on render?

It shouldn't hurt as the normal case has it go directly so the image's integer dimensions, so I'd say up to you~

PR #1314, opened 2022-02-01
Scaling extension: move blocks to own category

eanders-ms:
Moved to own **Sprite Scaling** category, under **Advanced**.

Color is: 
image

### DIFF
--- a/libs/sprite-scaling/scaling.ts
+++ b/libs/sprite-scaling/scaling.ts
@@ -1,14 +1,14 @@
-namespace sprites {
+//% color="#7B2CBF" weight=100 icon="\uf021" block="Sprite Scaling"
+//% advanced=true
+namespace scaling {
     //% blockId=sprite_set_scale_ex
     //% block="set $sprite=variables_get(mySprite) scale to $value || $direction anchor $anchor"
-    //% advanced=true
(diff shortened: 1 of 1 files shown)


libs/sprite-scaling/scaling.ts
```diff
@@ -94,17 +88,15 @@ namespace sprites {
         sprite.setScaleCore(sx, sy, anchor, proportional);
     }
 
-    //% blockId=sprite_shrink_by_pixels
+    //% blockId=sprite_shrink_by_pixels_ex
```
riknoll, line 91: this is a breaking change. not a problem since it's beta just thought I'd point it out :)

riknoll (approved):

jwunderl (approved):

libs/sprite-scaling/scaling.ts
eanders-ms: > this is a breaking change. not a problem since it's beta, just thought I'd point it out :)

Yep, I would not have made this change if it were released.

abchatra:
Most of the extensions add the category to the top on instead of advanced as the intent of the coder is to use it immediately if they have added the extension.

PR #1315, opened 2022-02-01
Wifi login page

pelikhan:
Add feature to serve a login page that allows a user to join the esp32 wifi and enter a AP name, password.


libs/wifi---esp32/httpserver.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (buf_len > 1 && buf_len < 256) {
+        char* buf = (char*)malloc(buf_len);
+        if (httpd_req_get_url_query_str(req, buf, buf_len) == ESP_OK) {
+            char name[64];
+            char password[64];
+            /* Get value of expected key from query string */
+            if ((httpd_query_key_value(buf, "name", name, sizeof(name)) == ESP_OK) &&
+                (httpd_query_key_value(buf, "password", password, sizeof(password)) == ESP_OK)) {
+                // save ap info, let TS handle it
+                if (NULL == _lastApBuffer) {
```
pelikhan, line 40: to prevent races; only submit once

PR #1316, opened 2022-02-02
Ensure hitbox pos matches scaled sprite pos

eanders-ms:
Sprite's `left` and `top` accessors each return a float. When the sprite is rendered, these values are floored to get the sprite's pixel coordinates. This wasn't happening in the hitbox calculation, which doesn't matter much at scale=1, but at larger scales it becomes noticeable. I fixed the issue by correspondingly flooring the left and top values passed to the hitbox.

Other changes in this PR:
- Account for scale in a few places where I wasn't, but should have been
- Changed Sprite `width` and `height` to return float rather than int. This seems more correct (again, it didn't matter so much at scale=1)
- Changed Sprite serialize to write scale values as float rather than int

Tagging @mmoskal for the `fixed.ts` additions and Sprite serialization changes

Fixes https://github.com/microsoft/pxt-arcade/issues/4546

### DIFF
--- a/libs/base/fixed.ts
+++ b/libs/base/fixed.ts
@@ -60,6 +60,12 @@ namespace Fx {
         else
             return b
     }
+    export function floor(v: Fx8): Fx8 {
+        return ((v as any as number) & ~0xff) as any as Fx8;
+    }
+    export function ceil(v: Fx8): Fx8 {
+        return (v as any as number) & 0xff ? Fx.floor(Fx.add(v, Fx.oneFx8)) : v;
+    }
     export function leftShift(a: Fx8, n: number) {
         return (a as any as number << n) as any as Fx8
     }
--- a/libs/game/sprite.ts
+++ b/libs/game/sprite.ts
@@ -301,7 +301,7 @@ class Sprite extends sprites.BaseSprite {
     }
 
     __serialize(offset: number): Buffer {
-        const buf = control.createBuffer(offset + 16);
+        const buf = control.createBuffer(offset + 20);
         let k = offset;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._x)); k += 2;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._y)); k += 2;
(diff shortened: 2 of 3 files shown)


libs/game/sprite.ts
```diff
@@ -376,11 +376,11 @@ class Sprite extends sprites.BaseSprite {
         // If it's just a small change to the hitbox on one axis,
         // don't change the dimensions to avoid random clipping
         this._hitbox = newHitBox;
-        if (xDiff <= Fx.twoFx8) {
+        if (xDiff <= Fx.mul(Fx.twoFx8, this._sx)) {
```
jwunderl, line 379: Hmm not certain I follow the change here; if scale is 2 then we will avoid updating the hitbox for differences up to 4px? It's an existing hack for minor position differences so leaving it at just 2 px seems preferable to me -- any more than that and the hitbox not changing will start to be noticeable.

libs/game/sprite.ts
```diff
@@ -309,8 +309,8 @@ class Sprite extends sprites.BaseSprite {
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._vy)); k += 2;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._ax)); k += 2;
         buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._ay)); k += 2;
-        buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._sx)); k += 2;
-        buf.setNumber(NumberFormat.Int16LE, k, Fx.toInt(this._sy)); k += 2;
+        buf.setNumber(NumberFormat.Int16LE, k, Fx.toFloat(this._sx)); k += 2;
```
jwunderl, line 312: ^^ will this work as expected? I guess that's why Michal is tagged in the first place :)

eanders-ms: @mmoskal will this work? how can I test it?

libs/game/sprite.ts
eanders-ms: I made this change so that the behavior would be consistent at different scales, but I can see how it would be better to stick with the small visual diff. I'll back it out.

jwunderl (approved):

libs/game/sprite.ts
mmoskal: I guess the change is a no-op - the serialized value will be floored anyways.

I'm not sure where this serialization is used (and I don't think I wrote it).

eanders-ms: @pelikhan is this code path testable?

libs/base/fixed.ts
```diff
@@ -60,6 +60,12 @@ namespace Fx {
         else
             return b
     }
+    export function floor(v: Fx8): Fx8 {
+        return ((v as any as number) & ~0xff) as any as Fx8;
+    }
+    export function ceil(v: Fx8): Fx8 {
+        return (v as any as number) & 0xff ? Fx.floor(Fx.add(v, Fx.oneFx8)) : v;
+    }
```
eanders-ms, line 68: @mmoskal do these Fx8 additions look ok?

riknoll (approved):

libs/base/fixed.ts
mmoskal: Yes, not sure about negative numbers though

PR #1323, opened 2022-02-10
Allow extension authors to register additional button handlers

riknoll:
I'd like to sneak this in for the patch release.

Currently only one event handler can be registered for each button; this means that extensions (like the storytelling extension) can't use any of the button events without overwriting the user's code. Thus, you have to resort to [terrible hacks](https://github.com/microsoft/arcade-storytelling/blob/main/menuState.ts#L93) to replicate this behavior.

This simply adds a JavaScript-only API that lets you add and remove event handlers in addition to the usual onEvent behavior. To the user, there will be absolutely no change. Registering more than one event for a button will override it.

I tested this pretty thoroughly in the sim, including:
* Registering multiple button handlers with onEvent to make sure they still get overwritten
* Doing lots of scene push/pop shenaningans to make sure events are saved/restored for the scene

### DIFF
--- a/libs/game/controllerbutton.ts
+++ b/libs/game/controllerbutton.ts
@@ -41,6 +41,17 @@ namespace controller {
         return // missing in sim
      }
 
+    export class ButtonHandler {
+        constructor(public event: number, public callback: () => void) { }
+    }
+
+    export class ButtonEventHandlerState {
+        constructor(public id: number) {};
+
+        public user: ButtonHandler[];
+        public system: ButtonHandler[];
+    }
+
     //% fixedInstances
     export class Button {
         _owner: Controller;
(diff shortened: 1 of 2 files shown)


libs/game/controllerbutton.ts
```diff
@@ -103,7 +122,49 @@ namespace controller {
...
+         * not conflict with events added via onEvent. The same handler can
+         * not be added for the same event more than once.
+         *
+         * @param event     The event to subscribe to for this button
+         * @param handler   The code to run when the event triggers
+         */
+        addEventListener(event: ControllerButtonEvent, handler: () => void) {
+            this.getOrCreateHandlerForEvent(event);
+
+            const handlerState  = this.handlerState;
```
jwunderl, line 141:
```suggestion
            const handlerState = this.handlerState;
```

nit

libs/game/controllerbutton.ts
```diff
@@ -103,7 +122,49 @@ namespace controller {
...
+        }
+
+        /**
+         * Removes an event handler registered with addEventListener.
+         *
+         * @param event     The event that the handler was registered for
+         * @param handler   The handler to remove
+         */
+        removeEventListener(event: ControllerButtonEvent, handler: () => void) {
+            const handlerState  = this.handlerState;
```
jwunderl, line 159:
```suggestion
            const handlerState = this.handlerState;
```

jwunderl (approved):

PR #1331, opened 2022-03-16
Categorize radio and move the order of blocks

abchatra:
Fixes https://github.com/microsoft/pxt-microbit/issues/3247 
While there I categorized the blocks as well to look nice.

image

### DIFF
--- a/libs/radio-broadcast/radio-broadcast.ts
+++ b/libs/radio-broadcast/radio-broadcast.ts
@@ -21,6 +21,7 @@ namespace radio {
     //% weight=200
     //% blockGap=8
     //% help=radio/send-message
+    //% group="Broadcast"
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
Looks good to me, I think a possibility might be moving setGroup to be the end of the 'Broadcast' group - it feels like it could belong there to me ("who / what group does it broadcast to") & then we can cut out one of the groups. Up to you though / I'm fine with it either way.

abchatra:
You need to set the group before you broadcast. Also broadcast is an extension. Set group applies to all blocks like send\receive as well and not just broadcast.  I am inclined to keep as is.

abchatra:
I will merge this. If you strongly feel the other way let me know, I will produce another PR.

PR #1333, opened 2022-03-22
Tiles ref folder migrate

ganicke:
The current reference pages for `tiles` namespace API's are moved from `./scene` into `./tiles`. Help paths and page links are also updated.

RE: #1332


ganicke:
Closed in favor of #1337.

PR #1335, opened 2022-03-29
Spelling mistake

Kikketer:


ganicke:
@Kikketer - thanks for finding that one.

PR #1339, opened 2022-04-27
[common packages] Support for online multiplayer

eanders-ms:
Adds support for online multiplayer messaging. Changes in this PR:
* Added a new simulator component `MultiplayerState`, which is enabled when the `mp=client|server` URL parameter is passed to the simulator iframe. When enabled, the component will operate in one of two states according to the value of `mp`:
    * `client`: The component will post keyboard events to the parent frame.
    * `server`: The component will post the current screen to the parent frame every 50ms.

To launch the simulator in **server** mode:
  - `https://arcade.makecode.com/--run?id=${gameId}&mp=server`

To launch to simulator in **client** mode:
  - `https://arcade.makecode.com/--run?code=${encodeURIComponent("multiplayer.init()")}&mp=client`

Associated PRs:
https://github.com/microsoft/pxt/pull/8806
https://github.com/microsoft/pxt-arcade-sim/pull/55


libs/game/sim/multiplayer.ts
```diff
@@ -0,0 +1,134 @@
...
+        content: "Button";
+        button: number; // pxsim.Key.A, ...
+        state: "Pressed" | "Released" | "Held";
+    }
+
+    type MultiplayerOrigin = "client" | "server" | undefined;
+
+    export class MultiplayerState {
+        lastMessageId: number;
+        origin: "client" | "server" | undefined;
```
eanders-ms, line 58: I made a slight change here so the sim will start up with origin set to undefined (i.e. not in a multiplayer mode). This ensures no multiplayer messages are sent.

libs/game/multiplayer.ts
```diff
@@ -0,0 +1,12 @@
+namespace multiplayer {
+    // registration code should probably go in here, but it's giving an error so moving to startup.ts for now
```
jwunderl, line 2: Ah, this is where I had the code that is over in pxt-arcade originally but it was the issue I mentioned with dependencies -- there's a decently good chance that you can remove all the changes from pxt-arcade & uncomment here and it should work just fine now that it is in game. (though we should either register this in an init function or have the sim side send an event that enables this when in multiplayer mode)

jwunderl (approved):

libs/game/multiplayer.ts
```diff
@@ -0,0 +1,21 @@
...
+    export function init() {
+        game.addScenePushHandler(() => {
+            game.eventContext().registerFrameHandler(scene.MULTIPLAYER_MESSAGING_PRIORITY, () => {
+                if (getOrigin() === "client") {
+                    const im: Image = getCurrentImage();
+                    scene.setBackgroundImage(im);
+                }
+            });
+        });
+    }
```
jwunderl, line 20:
```suggestion
            game.eventContext().registerFrameHandler(scene.MULTIPLAYER_MESSAGING_PRIORITY, () => {
                if (getOrigin() === "client") {
                    const im: Image = getCurrentImage();
                    scene.setBackgroundImage(im);
                    // clear default menu button behavior
                    controller.menu.onEvent(ControllerButtonEvent.Pressed, () => {});
                }
            });
        });
        game.pushScene();
    }
```

Ah I think this is likely what's causing the issue you mentioned in standup, the scene push handler was running before it was registered when this code lived in pxt-arcade but now we should probably call a push scene to make sure the frame handler registers. (and also added a clear menu so we can't get into a weird state with that)

eanders-ms: Yay you fixed my bug! I was missing the scene push!

package.json
```diff
@@ -24,7 +24,7 @@
     "libs/**/*"
   ],
   "dependencies": {
-    "pxt-core": "^7.5.19"
+    "pxt-core": "7.5.23"
```
eanders-ms, line 27: @jwunderl @riknoll @abchatra I changed this to be an exact version match. This is how it is in pxt-arcade. Was there a reason for caret matching?

riknoll: Not that I'm aware of. This isn't a super important dependency anyhow! it's really just for typings

abchatra: Change looks good to me. As Richard said not very important dependency as it doesn't impact the target build.

Vegz78:
> Adds support for online multiplayer messaging. Changes in this PR:

This is really, really cool, and probably a much sought-after functionality!


I made an attempt at beta-testing this, and could get the simulator to start in a way in server mode, e.g.:
https://arcade.makecode.com/beta/--run?id=09468-27855-19100-63192&mp=server

But I could not figure out how to start the client side and connect with the _code_ string `code=${encodeURIComponent("multiplayer.init()")}&mp=client`:
> To launch to simulator in **client** mode:
> 
> * `https://arcade.makecode.com/--run?code=${encodeURIComponent("multiplayer.init()")}&mp=client`

Any friendly advice on how to get this cool feature to work for your enthusiastic and loyal fans? ;-)

riknoll:
@jwunderl @abchatra ?

abchatra:
@Vegz78 this feature is not ready yet. We are working on it. We will definitely update this issue: https://github.com/microsoft/pxt-arcade/issues/3016 as we make progress.

Thanks for the support.

PR #1342, opened 2022-05-19
[common] Support for remapping keyboard input

eanders-ms:
This PR adds support for remapping player inputs at runtime. This makes it possible for [Makey Makey](https://makeymakey.com/), for example, to create an extension that would define a block to remap the simulator for their board's input scheme. The extension might look like this:

```ts
namespace MakeyMakey {
    //% blockId=set_makeymakey_controls
    //% block="Remap simulator keys to Makey Makey defaults"
    export function setMakeyMakeyInputs() {
        // Configure input for player 1
        keymap.setPlayerKeys(
            1,
            keymap.KeyCode.UpArrow,
            keymap.KeyCode.DownArrow,
            keymap.KeyCode.LeftArrow,
            keymap.KeyCode.RightArrow,
            keymap.KeyCode.MouseLeftButton,
            keymap.KeyCode.Space);
        // Configure input for player 2
        keymap.setPlayerKeys(
            2,
            keymap.KeyCode.W,
            keymap.KeyCode.S,
            keymap.KeyCode.A,
            keymap.KeyCode.D,
            keymap.KeyCode.F,
            keymap.KeyCode.G);
    }
}
```

Projects built for Makey Makey would need to include this block in `on start`:
image

### Out of scope
* Runtime changes to keyboard inputs are not reflected in the keyboard popup:
 
image

### DIFF
--- a/libs/game/keymap.ts
+++ b/libs/game/keymap.ts
@@ -11,7 +11,7 @@ namespace keymap {
     ): void;
 
     //% shim=keymap::_setSystemKeys
-    declare function _setSystemKeys(screenshot: number, gif: number): void;
+    declare function _setSystemKeys(screenshot: number, gif: number, menu: number, reset: number): void;
 
     /**
      * Sets the keyboard input map for the given player.
@@ -39,15 +39,19 @@ namespace keymap {
      * Sets the keyboard input map for system keys.
      * @param screenshot The key code for 'screenshot'
      * @param gif The key code for 'gif'
+     * @param menu The key code for 'menu'
+     * @param reset The key code for 'reset'
      */
-    export function setSystemKeys(screenshot: KeyCode, gif: KeyCode) {
-        _setSystemKeys(screenshot, gif);
+    export function setSystemKeys(screenshot: KeyCode, gif: KeyCode, menu: KeyCode, reset: KeyCode) {
+        _setSystemKeys(screenshot, gif, menu, reset);
     }
 
     /**
      * Key codes
      */
     export enum KeyCode {
+        None = 0,
+
         Backspace = 8,
         Tab = 9,
         Enter = 13,
--- a/libs/game/keymap.ts
+++ b/libs/game/keymap.ts
@@ -1,0 +1,165 @@
+namespace keymap {
+    //% shim=keymap::_setPlayerKeys
+    declare function _setPlayerKeys(
+        player: number, // player number is 1-based
+        up: number,
+        down: number,
+        left: number,
+        right: number,
+        A: number,
+        B: number
+    ): void;
+
+    //% shim=keymap::_setSystemKeys
+    declare function _setSystemKeys(screenshot: number, gif: number): void;
+
+    /**
+     * Sets the keyboard input map for the given player.
+     * @param player The player number. 1 = Player1, etc.
+     * @param up The key code for 'up'.
+     * @param down The key code for 'down'
+     * @param left The key code for 'left'
+     * @param right The key code for 'right'
+     * @param A The key code for 'A'
+     * @param B The key code for 'B'
+     */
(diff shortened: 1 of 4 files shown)


eanders-ms:
Resolves https://github.com/microsoft/pxt-arcade/issues/4738

abchatra:
Yay 😹

libs/game/keymap.ts
```diff
@@ -0,0 +1,165 @@
...
+    ) {
+        _setPlayerKeys(player, up, down, left, right, A, B);
+    }
+
+    /**
+     * Sets the keyboard input map for system keys.
+     * @param screenshot The key code for 'screenshot'
+     * @param gif The key code for 'gif'
+     */
+    export function setSystemKeys(screenshot: KeyCode, gif: KeyCode) {
```
jwunderl, line 43: restart sim and menu button too maybe?

libs/game/sim/keymap.ts
```diff
@@ -0,0 +1,163 @@
...
+            return this.keymap[keyCode] || this.altmap[keyCode] || Key.None;
+        }
+
+        private saveMap(name: string, keyCodes: number[]) {
+            this.mappings[name] = keyCodes;
+        }
+
+        private clearMap(name: string) {
+            const keyCodes = this.mappings[name];
+            keyCodes?.forEach(keyCode => delete this.keymap[keyCode]);
```
jwunderl, line 159: I thought this runs alongside the static typescript compiler / gets mad at the `?.` feature of newer typescript? I guess if it works then I'm wrong though~

eanders-ms: This is in the sim.

eanders-ms: Does this run in the sim context, or STS? Now I'm not sure.

eanders-ms: The compiler expanded it to
```ts
keyCodes === null || keyCodes === void 0 ? void 0 : keyCodes.forEach(keyCode => delete this.keymap[keyCode]);
```

eanders-ms: So, it's called by the game (from inside the run/eval), but is not compiled by STS because it's a sim-only implementation of the api.

jwunderl: Ah right, looks good -- I think last time I ran into that was on a branch that hadn't fully completed the bluebird change and it stuck in my head as it used to happen a lot.

jwunderl (approved):

AqeeAqee:
Hi there!
This seams made keys works not very correctly in sim of my PC. Not sure, but seams begin after updated to the arcade 1.8.25. Maybe the issue come from following lines.
https://github.com/microsoft/pxt-common-packages/blob/796d06e4764862236b1f931ac46a2417b679bb01/libs/game/sim/keymap.ts#L69-L72

abchatra:
@AqeeAqee can you describe the issue? What doesn't work?

AqeeAqee:
Sorry, I should describe it more detail.

The phenomenon is:
Open a project of Arcade, in sim, using arrow buttons control with A,S,D,W keys. 
Press S key works as Right, and D key doesn't work.

Expect:
S key works as Down, D key works as Right.

abchatra:
@eanders-ms ?

eanders-ms:
@AqeeAqee thank you for reporting this! I’m taking a look now. Could please let me know:

* Which operating system?
* Which browser are you using?
* Is your system or keyboard configured for a non-English language?

Thanks!

eanders-ms:
Oh, I think the key mappings are just incorrect. Fixing...

eanders-ms:
@AqeeAqee Fix is live! Let us know if you're still seeing problems.

PR #1343, opened 2022-05-20
Add class to make it easier to extend sprite

riknoll:
Adds a new class "ExtendableSprite" with a number of features to make extending a sprite easier:

1. The constructor now automatically adds the sprite to the physics engine and runs the "on created" handlers
2. Adds a "draw" method that can be overridden without having to reimplement the sprite say or debug info drawing. It also passes in a left and top instead of the camera that already takes into account the RelativeToCamera flag
3. Adds an "update" method that takes in the delta time in milliseconds instead of seconds and can be overridden without messing up the other sprite update logic.

As an added bonus, the new draw and update methods have signatures that are much easier to remember than the ones from BaseSprite which I always have to look up whenever I'm doing this.

@jwunderl any other ideas on what would be nice to add to this?

### DIFF
--- a/libs/game/extendableSprite.ts
+++ b/libs/game/extendableSprite.ts
@@ -1,0 +1,51 @@
+namespace sprites {
+    /**
+     * A version of the Sprite class that is easier to extend.
+     * 
+     * Unlike the normal Sprite class, this class will automatically add
+     * itself to the physics engine and run all sprite created handlers
+     * in the constructor
+     */
+    export class ExtendableSprite extends Sprite {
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
maybe if it had like a 'sethitbox(width, height)' method or something that would handle setting width / height that would be nice? Can be done separately but I feel like we always end up wanting to do that anyways since we have to separately calculate it

eanders-ms (approved):
nice!

riknoll:
@jwunderl added!

PR #1356, opened 2022-06-24
temporarily change reset to 2 for kiosk

jwunderl:


abchatra:
This needs to be reverted after ISTE.

abchatra (approved):

PR #1357, opened 2022-06-29
Revert "temporarily change reset to 2 for kiosk"

abchatra:
Reverts microsoft/pxt-common-packages#1356


abchatra:
ISTE is done and we want the backspace as reset

jwunderl (approved):

PR #1365, opened 2022-07-28
Add sound effect blocks to pxt-common-packages

riknoll:
Adds the sound effect field editor to pxt-common-packages using the mixer sound engine instead of the codal one.

image

The "randomize" block is the same as pressing "generate similar sound" in the field editor


jwunderl (approved):
AdventureTimeGIF

PR #1367, opened 2022-08-09
Fix Off-by-One Error Traversing Image Rows

thsparks:
When we reset the height field at the end of each "row", we should actually set it to 1 instead of 0, since the code expects it to have been incremented once already at this point.

Fixes https://github.com/microsoft/pxt-arcade/issues/4690

Using the sample from the above bug, without fix:
image

With fix:
image


jwunderl (approved):
Ahh nice, that makes sense! I'm gonna add Michal to double check as well but should be good!

mmoskal (approved):

riknoll (approved):

PR #1368, opened 2022-08-17
Respect Transparency for Screen Edge Collisions

thsparks:
This change switches "Stay in Screen" and "Bounce on Wall" checks to use the sprite's Hitbox instead of the left, right, top, and bottom fields. Left, right, top, and bottom do not account for transparency around the sprite, so a sprite with a transparent border would look like it could never actually reach the edge (see https://github.com/microsoft/pxt-arcade/issues/4846 for an example). By using HitBox, we ensure the sprite will not stop/bounce until a non-transparent section of the sprite intersects with the edge.

I've also moved this check into the PhysicsEngine, which seems like a more natural place for it.

Fixes https://github.com/microsoft/pxt-arcade/issues/4846

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -195,6 +195,13 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                 if (tileMap && tileMap.enabled) {
                     this.tilemapCollisions(ms, tileMap);
                 }
+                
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
Awesome, nice job~

riknoll (approved):

PR #1371, opened 2022-09-15
Surfacing more image blocks

riknoll:
As has become tradition, I'd like to sneak a few missing blocks into this release.

This PR surfaces some already existent TS image apis into blocks:

image
image

In particular, the width/height block is extremely useful because the only way to get those values right now is to create a sprite (which you then have to immediately destroy).

### DIFF
--- a/libs/screen/image.d.ts
+++ b/libs/screen/image.d.ts
@@ -156,6 +156,7 @@ interface Image {
      */
     //% shim=ImageMethods::drawTransparentImage
     //% blockNamespace="images"
+    //% blockId=image_draw_transparent_image
     //% group="Drawing"
     //% block="$this draw $from at x $x y $y"
     //% this.shadow=variables_get
--- a/libs/screen/image.d.ts
+++ b/libs/screen/image.d.ts
@@ -31,6 +31,46 @@ interface Image {
     drawRect(x: number, y: number, w: number, h: number, c: color): void;
 
     /**
+     * Draw an empty circle.
+     *
+     * @param cx The center x coordinate of the circle
+     * @param cy The center y coordinate of the circle
+     * @param r The radius of the circle
+     * @param c The color to draw the circle
+     */
+    //% helper=imageDrawCircle
+    //% blockId=image_draw_circle
+    //% block="draw circle in $this at cx $cx cy $cy radius $r $c"
+    //% this.shadow=variables_get
+    //% this.defl=picture
+    //% c.shadow=colorindexpicker
+    //% inlineInputMode="inline"
+    //% weight=10
+    //% blockNamespace="images"
+    //% group="Drawing"
+    drawCircle(cx: number, cy: number, r: number, c: color): void;
+
+    /**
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

libs/screen/image.d.ts
```diff
@@ -110,6 +150,19 @@ interface Image {
...
+     */
+    //% shim=ImageMethods::drawTransparentImage
+    //% blockNamespace="images"
+    //% group="Drawing"
+    //% block="$this draw $from at x $x y $y"
+    //% this.shadow=variables_get
+    //% this.defl="picture"
+    //% from.shadow=screen_image_picker
+    //% weight=0
+    drawTransparentImage(from: Image, x: int32, y: int32): void;
```
eanders-ms, line 165: Does this need a blockId?

eanders-ms: Do you want to support scale? :)

eanders-ms (approved):

libs/screen/image.d.ts
riknoll: Added the blockid! Not opposed to supporting scale, but it might be better off in the scale extension?

Also, fun fact: none of the blocks in the image category other than the ones in this pr have block ids 😅

abchatra:
What does picture draw do? Add picture inside picture?

riknoll:
yeah, it draws one picture on top of another. it's very useful for things like HUDs, where you might want to draw some hearts for lives, etc.

Jaqster:
Why can't we put these inside an extension?  They seem like advanced scenarios to me...  Or if you feel like these are important, we could talk about removing some of the existing Image blocks?  Do we have telemetry on how popular these blocks are?

riknoll:
@Jaqster I'd be okay with removing the circle blocks (i just added those for completeness), but the other three blocks in this pr are all very useful and I think belong in core. I don't think these are any more advanced than the rest of the blocks in the image category.

I think all of these blocks fall into the category of things you would expect any reasonable game making tool to have.

My other motivation is that there isn't a clear extension that these blocks belong in. I want to avoid having a million micro extensions that each contain less than five blocks.

abchatra:
I wonder if entire Image category should be an extension? (breaking change for sure). Are there any tutorials\skillmaps use this category?

Jaqster:
How many blocks total would be in the Images category?  The most we have in 1 category is 30 (Sprite and Scene) - and I think that's too much.  We should talk about maybe setting some limits on # of blocks per category?  Makes it really hard to find stuff...

Speaking of, I just did a quick count, and this is what it looks like across our code editors (not counting common blocks):
- Minecraft - 128 blocks, 8 categories
- Arcade - 137 blocks, 9 categories
- Micro:bit - 147 blocks, 10 categories

I think we can do some clean-up in micro:bit (remove some categories from Advanced), and let's really try and focus on only having the most commonly used blocks in the mainline editors.

thsparks (approved):

PR #1375, opened 2022-10-05
Added: getCountdown, changeCountDownBy blocks

srietkerk:
Added two blocks so that users can get the current countdown of their game and also change the countdown by negative or positive seconds. Negative makes it so there is less time left on the countdown, positive makes it so there is more time on the countdown. 

**Quick Demo:** 

https://user-images.githubusercontent.com/49178322/194131975-ac438c6c-afd5-4234-a06f-e951f82faf48.mp4

**The Blocks** (the two new are at the bottom of the screenshot)

countdown-group

Closes https://github.com/microsoft/pxt-arcade/issues/2248

**The target:** https://arcade.makecode.com/app/bf585e360f8e57bd8312ad0b0ea40b6f706ddba7-421bc06f42#editor

### DIFF
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -365,8 +365,9 @@ namespace info {
     //% blockId=gamegetcountdown
     //% weight=79
     //% group="Countdown"
-    export function getCountdown() {
-        return (infoState.gameEnd - game.currentScene().millis()) / 1000;
+    export function countdown(): number {
+        initHUD();
+        return infoState.gameEnd ? ((infoState.gameEnd - game.currentScene().millis()) / 1000) : 0;
     }
 
     /**
@@ -383,18 +384,15 @@ namespace info {
     }
 
     /**
-     * Get the value of the current count down
+     * Change the running countdown by the given number of seconds
+     * @param seconds the number of seconds the countdown should be changed by
      */
     //% block="change countdown by $seconds (s)"
     //% blockId=gamechangecountdown
     //% weight=77
     //% group="Countdown"
     export function changeCountdownBy(seconds: number) {
-        if (!infoState) {
-            startCountdown(seconds);
-            return;
-        }
-        infoState.gameEnd += (seconds * 1000);
+        startCountdown((countdown() + seconds));
     }
 
     /**
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -384,6 +384,26 @@ namespace info {
     }
 
     /**
+     * Get the value of the current count down
+     */
+    //% block
+    //% blockId=gamegetcountdown
+    //% group="Countdown"
+    export function getCountdown() {
+        return (infoState.gameEnd - game.currentScene().millis()) / 1000;
+    }
+
+    /**
+     * Get the value of the current count down
+     */
+    //% block="change countdown by $seconds (s)"
+    //% blockId=gamechangecountdown
+    //% group="Countdown"
+    export function changeCountdownBy(seconds: number) {
+        infoState.gameEnd += (seconds * 1000);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


libs/game/info.ts
```diff
@@ -383,6 +383,26 @@ namespace info {
...
+    }
+
+    /**
+     * Get the value of the current count down
+     */
+    //% block="change countdown by $seconds (s)"
+    //% blockId=gamechangecountdown
+    //% group="Countdown"
+    export function changeCountdownBy(seconds: number) {
+        infoState.gameEnd += (seconds * 1000);
```
jwunderl, line 403: This should probably call into startcountdown so the flags get set - changeCountdownBy(5) when no countdown is set should be the same as a startCountdown(5)

libs/game/info.ts
```diff
@@ -383,6 +383,26 @@ namespace info {
...
     }
 
+    /**
+     * Get the value of the current count down
+     */
+    //% block="countdown"
+    //% blockId=gamegetcountdown
+    //% group="Countdown"
+    export function getCountdown() {
+        return (infoState.gameEnd - game.currentScene().millis()) / 1000;
```
jwunderl, line 393: Can you add weights to these two so the order matches score / life - should be getter first, then setter, then change by,  and then stop / on countdown end

jwunderl (changes requested):
Couple minor changes

libs/game/info.ts
srietkerk: Yes, definitely. I noticed that the weights for life / score are all different (ie score is in the 90s while life is in the 80s). Why do the different groups have to be in different ranges?

jwunderl: No real reason, just historical - I don't think we had groups in this category at first and when the groups were added no one saw a reason to change the existing weights.

For what it's worth I'll sometimes still keep that pattern just so it's easier to visualize without accounting for the group - even if it doesn't matter cross group it can be helpful when laying out an entire group

thsparks (approved):

libs/game/info.ts
```diff
@@ -358,23 +358,49 @@ namespace info {
...
     }
 
+    /**
+     * Get the value of the current count down
+     */
+    //% block="change countdown by $seconds (s)"
+    //% blockId=gamechangecountdown
+    //% weight=77
+    //% group="Countdown"
+    export function changeCountdownBy(seconds: number) {
```
jwunderl, line 392: Infostate should also exist if score or life is used without startcountdown - id just always do `startCountdown((countdown() + seconds) * 1000)` i think

libs/game/info.ts
```diff
@@ -358,23 +358,49 @@ namespace info {
...
     }
 
+    /**
+     * Get the value of the current count down
+     */
+    //% block="countdown"
+    //% blockId=gamegetcountdown
+    //% weight=79
+    //% group="Countdown"
+    export function getCountdown() {
```
jwunderl, line 368: Tiny nit i should have mentioned: for consistency with life / score / other getters should just be countdown(), not getCountdown()

srietkerk: Thank you for the tips! They return 0, so I will do the same for countdown.

srietkerk: Should I upload a new target to be played with?

jwunderl (approved):
LGTM

libs/game/info.ts
jwunderl: It seems pretty safe / I wouldn't bother personally. Changes lgtm, should be fine / can always fix if we notice an issue lster on while it's in beta

abchatra:
Nice job @srietkerk

PR #1378, opened 2022-10-17
Multiplayer fix host can pretend to be all players

jwunderl:
setButton basically just calls handleKeyEvent, and is the location at which I'm rejecting buttons from players 2+ in multiplayer sessions, so just bypassing that.

Also minor typings fix to get rid of red squiggle, and fix usage of safe navigation operator not supported in ts2.6

part of fix for https://github.com/microsoft/pxt-arcade/issues/5138

This should go in at same time as or just before arcade sim side change, it's safe to be on it's own.


aznhassan (approved):
:shipit:

PR #1380, opened 2022-10-20
only post screen when at end of render, incl palette in sim<->host messages

jwunderl:
two main things:

* Send the screen from within the game engine, after the screen update step. Currently, we just send it at a random time in the loop, causing things to occasionally be misrendered / missing if they are done part way through the render (that is, it's not uncommon to see screens that are supposed to have a dialog flash the scene behind it without the dialog)
* Patch on palette to the messages that go between sim <-> host frame -- tested on the sending side, not on the consuming side yet. Have to check with Eric on how we want to format this in the --multiplayer <-> backend communication where it's binpacked -- it's currently a 48b uint8array, we can safely chop off the first 3 bytes if we want as those are always 0x00000 (transparency) -- even if someone tries to set it we just ignore that value.

~dependent on adding throttle to pxtsim/runtime.ts: https://github.com/microsoft/pxt/pull/9130, will need to bump pxt version in package.json after that's in~ pxt version bumped so ready to go~

### DIFF
--- a/libs/game/multiplayer.cpp
+++ b/libs/game/multiplayer.cpp
@@ -2,18 +2,15 @@
 
 namespace multiplayer {
     //%
-    void postImage(Image_ im, String goal) {
-        // no support >:(
+    void postImage(Image_ im) {
     }
 
     //%
     void setOrigin(String origin) {
-        // no
     }
 
     //%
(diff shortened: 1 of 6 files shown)


eanders-ms (approved):
LGTM

PR #1382, opened 2022-11-01
multiplayer send around audio channels

jwunderl:
One question, do we want to attach a millis field (should just be `runtime.runningTime()`) in case we eventually want to do some type of timing?

(also unrelated I snuck in some ?. by accident last time and got rid of them here, it's not a big deal as browsers support them nowadays but figure it's better to minimize errors in the build when it's easy)

Build will fail until pxt pr which contains `soundEventCallback` is in (https://github.com/microsoft/pxt/pull/9165)~

### DIFF
--- a/libs/game/sim/multiplayer.ts
+++ b/libs/game/sim/multiplayer.ts
@@ -8,10 +8,11 @@ namespace pxsim.multiplayer {
             return;
         const asBuf = pxsim.image.toBuffer(im);
         const sb = board() as ScreenBoard;
+        const screenState = sb && sb.screenState;
         throttledImgPost(<MultiplayerImageMessage>{
             content: "Image",
             image: asBuf,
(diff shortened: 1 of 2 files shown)


eanders-ms (approved):

libs/game/sim/multiplayer.ts
```diff
@@ -84,6 +91,17 @@ namespace pxsim {
         init(origin: string) {
             this.origin = origin;
             runtime.board.addMessageListener(msg => this.messageHandler(msg));
+            if (this.origin === "server") {
+                pxsim.AudioContextManager.soundEventCallback = (ev: "playinstructions" | "muteallchannels", data?: Uint8Array) => {
+                    this.send(<pxsim.MultiplayerAudioEvent>{
```
riknoll, line 96: extremely personal nit: use `as` instead of angle brackets

riknoll (approved):

PR #1385, opened 2022-12-05
Fixing camera read loc

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5440

here's a build for it as it's easiest to just test it https://arcade.makecode.com/app/708640d9e59404f6d619b6d9e551c9854dfa2596-d43654d8b7

This makes it so we basically pass through to the sprites position when grabbing camera state - so the pause(0) hack we did is no longer necessary, and it's fine bumping the code out of the `run code after game engine updates the camera` back into an on game update if we want. Here's a version where I did that: https://arcade.makecode.com/app/708640d9e59404f6d619b6d9e551c9854dfa2596-d43654d8b7#pub:_FFLTaRKXePxE

### DIFF
--- a/libs/game/camera.ts
+++ b/libs/game/camera.ts
@@ -8,6 +8,8 @@ namespace scene {
         drawOffsetX: number;
         drawOffsetY: number;
         sprite: Sprite;
+        protected _lastUpdatedSpriteX: number;
+        protected _lastUpdatedSpriteY: number;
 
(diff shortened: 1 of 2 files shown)


riknoll (approved):

riknoll:
We should make sure to test this with a bunch of games before releasing. we've regressed camera stuff before (though i expect this fixes many more problems than it causes)

PR #1387, opened 2022-12-10
Add multiplayer extension

eanders-ms:
This initial PR just adds Richard's multiplayer blocks, with some re-namespacing so that they extend existing toolbox categories rather than collect into a Multiplayer category. It does also introduce a Multiplayer category for blocks that didn't fit nicely anywhere else.

In a future PR, I will:
* Add game over blocks
* Add countdown block
* Add blocks for setting presence bar icon


eanders-ms:
ping... this needs review.

jwunderl (approved):
LGTM to merge and we can iterate as needed

abchatra:
I see blocks being added to all different categories. I am not sure how user will know these blocks will be added to the categories.  May be add them all to the "Multiplayer" category or at both locations? Usually, students will look into the new category which is added.

eanders-ms:
> I see blocks being added to all different categories. I am not sure how user will know these blocks will be added to the categories. May be add them all to the "Multiplayer" category or at both locations? Usually, students will look into the new category which is added.

Multiplayer is a feature, not a category itself. It spans all the categories. I see the discoverability problem you're highlighting, but putting all these blocks in a single category has arguably greater discoverability issues. When a user wants to find the `set player-1 sprite to mySprite` block, where do you imagine them looking first: in Sprites? or Multiplayer? Adding all these blocks to a single category would result in a monstrously long scrolling toolbox drawer.

eanders-ms:
Is it possible to add a block in two locations?

abchatra:
I don't think it is possible without a separate shim Api.

I am ok adding in the current way and test with the user. First place I looked for the new blocks is multiplayer category and saw only few blocks and wondered where the rest of the stuff. 99% extensions add only blocks to single category and user is likely trained to expect that.  

>> When a user wants to find the set player-1 sprite to mySprite block, where do you imagine them looking first: in Sprites? or Multiplayer?

I don't know. It is good to test this as well.

PR #1388, opened 2022-12-29
Reduce path animation memory and synchronize animations on frame time

riknoll:
It's time for another classic "richard sneaks in a change he's been wanting to make forever while nobody is paying attention  over winter break" PR!

This PR refactors the path animation code to improve a few things:

1. Paths are now parsed as they execute instead of upfront, which reduces memory usage
2. Paths now correctly parse floating point arguments (previously, only integers worked)
3. Added some helpful exceptions when there is an error parsing a path
4. Both types of animation (path and image) now use the same delta time that the physics engine uses for timing. Previously, each animation called control.millis() as it updated which caused all of the animations to drift out of sync over time
5. Reduces total compiled code size by 0.1% (tested with https://github.com/microsoft/pxt-arcade/issues/5466)

Here's a sample program that will cause an out of memory panic on the pygamer with the current implementation but runs just fine with these changes (albeit at 2 frames per second):

https://makecode.com/_0soAvH7Xx0Uh

That program is also a great illustration of the drifting issue.

As part of this effort, I also took a stab at redoing all the math to use Fx8 to improve hardware perf. Unfortunately, doing so made the animations jittery. I expect this was a rounding error of some kind but I couldn't figure it out.

### DIFF
--- a/libs/game/animation.ts
+++ b/libs/game/animation.ts
@@ -152,8 +152,7 @@ namespace animation {
             }
         }
 
-        public run(interval: number, target: Sprite, startedAt: number): boolean {
-            const runningTime = control.millis() - startedAt; // The time since the start of the path
+        public run(interval: number, target: Sprite, runningTime: number): boolean {
             const nodeIndex = Math.floor(runningTime / interval); // The current node
             const nodeTime = runningTime % interval; // The time the current node has been animating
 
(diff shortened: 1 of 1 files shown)


abchatra (approved):
looks great

eanders-ms (approved):

PR #1389, opened 2023-01-11
pass up player icons & reaction icons to multiplayer page

jwunderl:
final portion of the support, depends on https://github.com/microsoft/pxt/pull/9287 and https://github.com/microsoft/pxt-backend/pull/748 . @eanders-ms if you're not in the middle of a pr I'll just go ahead and swap the `*.multiplayer` namespaces to `*.mp` so that I can remove the hack in `/mp.ts`. Here's a test game that sets the two user sprites + a pizza for slot one: 
https://makecode.com/_XrmfzuFdyLHt

(no blocks for setting icons directly, just the javascript api -- I'll add it to sprite util extension or something unless we decide to include one)

for the javascript apis, there's a possibly slightly tricky 'implicitly set vs explicitly set' state for whether the icon gets sent or not -- this is so we can add the default behavior of 'set / update player icon whenever they setPlayerSprite' without overriding images that the player has explicitly set (e.g. with the extension blocks I mentioned above)

### DIFF
--- a/libs/multiplayer/mp.ts
+++ b/libs/multiplayer/mp.ts
@@ -1,5 +1,4 @@
-const postPresence = multiplayer.postPresenceIcon;
-namespace sprites.multiplayer {
+namespace sprites.mp {
     //% blockId=mp_setPlayerSprite
     //% block="set $player sprite to $sprite=variables_get(mySprite)"
     //% player.shadow=mp_playernumber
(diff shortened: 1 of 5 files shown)


libs/multiplayer/mp.ts
```diff
@@ -1,3 +1,4 @@
+const postPresence = multiplayer.postPresenceIcon;
```
jwunderl, line 1: This is a hack to get around the `multiplayer` namespace getting shadowed by `sprites.multiplayer`, won't merge in as is~

eanders-ms (approved):

kimprice (approved):

PR #1392, opened 2023-01-19
Multiplayer extension updates

eanders-ms:
* Moves all extension blocks to the "Multiplayer" category
* Makes "Player" an object (no longer a number)
* Adds handler block for controller events: connected/disconnected
* Adds "get player from sprite" block


eanders-ms:
Shoot, I used `git mv` to rename mp.ts to player.ts, but the PR still sees them as different files. 😤

eanders-ms:
Test target: https://arcade.makecode.com/app/623272af2d34999e0057c30e0f7ec087557935cc-ea04783fc3

libs/multiplayer/player.ts
```diff
@@ -0,0 +1,672 @@
...
+        //% block="player 3"
+        Three = 3,
+        //% block="player 4"
+        Four = 4
+    }
+
+    export enum PlayerProperty {
+        //% block="index"
+        Index = 1,
+        //% block="slot"
```
jwunderl, line 21: I'm not sure I'd get what slot means here necessarily on the block -- could this maybe be something `PlayerNumber` / on the block as just `player number`? Maybe we should ask on forum to see if people 'get' it?

jwunderl (approved):
Overall looks good to me, though I'm a little iffy on the 'slot' terminology going on a block / whether that's a term users would get

libs/multiplayer/player.ts
abchatra: just _number_ would do.

abchatra:
Can mp_multiplayerstate be "score" and "life" instead of "Score" & "Lives"

image

abchatra:
Also can we have player from slot api to be consistent along with following? Or combine them to one? 
image

eanders-ms:
> Can mp_multiplayerstate be "score" and "life" instead of "Score" & "Lives"
> 
> image

These are "kinds" in the multiplayer extension. Uppercase is consistent with how we name "kinds" elsewhere. Do you want to introduce this inconsistency?

eanders-ms:
Using _number_ now:
arcade-screenshot (12)

eanders-ms:
The latest getters:
image

PR #1398, opened 2023-01-24
Game over updates for multiplayer

eanders-ms:
Partially addresses https://github.com/microsoft/pxt-arcade/issues/513. One more change coming after this in a separate PR to modify the multiplayer extension.

This change adds new blocks for configuring game over, and adds support for multiplayer scoring.

### New blocks
image

**use effect**: Set the effect that occurs when the game is over.
**use sound**: Set the music that occurs when the game is over.
**use message**: Set the message that displays when the game is over.
**use scoring**: Set the method of judging the best score for the game.

### Modified blocks & block behaviors
**game over**: Deprecated the effect parameter on existing block.
**on countdown end**, **on life zero**: removed default sound and visual effect from game over dialog. They can be re-added using the new blocks above.

### Single player examples
image
image
image
image
image

### Multiplayer examples
image
image
image

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -9,10 +9,96 @@ namespace game {
      */
     export let debug = false;
     export let stats = false;
-    export let winEffect: effects.BackgroundEffect = undefined;
-    export let loseEffect: effects.BackgroundEffect = undefined;
-    let loseSound: music.Melody = undefined;
-    let winSound: music.Melody = undefined;
+
+    export enum ScoringType {
+        //% block="high score"
+        HighScore,
+        //% block="low score"
+        LowScore,
+        //% block="none"
+        None
+    }
+
+    export class GameOverConfig {
+        scoringType: ScoringType;
+        winEffect: effects.BackgroundEffect;
+        loseEffect: effects.BackgroundEffect;
+        loseSound: music.Melody;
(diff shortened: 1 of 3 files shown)


libs/game/game.ts
```diff
@@ -139,31 +215,78 @@ namespace game {
...
 
     /**
-     * Set the music that occurs when the player wins
-     * @param win
+     * Set the music that occurs when the game is over
+     * @param win whether the sound should play on a win (true) or lose (false)
      * @param effect
      */
+    //% blockId=game_setgameoversound
+    //% block="use sound $sound for $win=toggleWinLose"
```
jwunderl, line 241: it's preferred to avoid setting shadows in block definitions, as that makes them harder to translate / introduces chances for translators to make mistakes (we haven't gone through and updating existing ones to avoid breaking translations for the blocks)

you can set the shadow like this https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/tilemap.ts#L644 to

```typescript
//% win.shadow=toggleWinLose
```

jwunderl (commented):
Looks great, just minor adjustment to block definitions

jwunderl (approved):

clkantner:
It looks good to me too!

PR #1399, opened 2023-01-24
Unify all music playback blocks

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/5552

This PR unifies all of the various melody/song/tone/effect blocks into a single playback block with different shadows for each type. The playback block has three options for playback: "until done", "in background", or "looping in background". Here's what the toolbox looks with this change:

image
image
image
image

All of the shadow blocks here have the "duplicate on drag" behavior so you can drag them out of the block to assign to variables or whatever. Not only does this change unify all of the blocks, it also reduces the overall number of blocks in the toolbox.

No upgrade rules are needed. Here's a build link: 

https://arcade.makecode.com/app/9896b4c680b9f43609dfc35cf51de2d68d18f0ba-e997f1cd4e

### DIFF
--- a/libs/mixer/instrument.ts
+++ b/libs/mixer/instrument.ts
@@ -1,6 +1,8 @@
 namespace music.sequencer {
     const BUFFER_SIZE = 12;
 
+    let currentSequencer: sequencer.Sequencer;
+
     /**
      * Byte encoding format for songs
      * FIXME: should this all be word aligned?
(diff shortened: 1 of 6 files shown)


Jaqster (approved):
Oh wow!!  This is so much better!!  Thanks Richard!

eanders-ms (approved):
noice!

jwunderl (approved):

PR #1400, opened 2023-01-24
add 'none' option for particle effects dropdown

jwunderl:
I believe this is what you were looking for? This will add it to all effects dropdowns but 🤷, maybe we can make it so this clears all sources to treat it as a proper 'clear all effects' effect (if you wanna try that just making prune call destroy on every other element in the array)


eanders-ms (approved):
ty!

PR #1401, opened 2023-01-24
Game namespace cleanup

eanders-ms:
* Pull namespace config into one location
* Move Game Over sub-category up to second position (tentative! need to look at telemetry for any surprise usage patterns)

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -1,8 +1,6 @@
 /**
  * Game transitions and dialog
  **/
-//% color=#8854d0 weight=97 icon="\uf111"
-//% groups='["Gameplay", "Prompt"]'
 namespace game {
     /**
      * Determines if diagnostics are shown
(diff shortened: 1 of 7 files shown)


libs/game/game.ts
```diff
@@ -1,8 +1,6 @@
 /**
  * Game transitions and dialog
  **/
-//% color=#8854d0 weight=97 icon="\uf111"
-//% groups='["Gameplay", "Prompt"]'
```
jwunderl, line 5: I would probably look for the contents of ns.ts here rather than in a separate file, but not a big deal either way

jwunderl (approved):

libs/game/game.ts
eanders-ms: This seems to be a common pattern across many common packages. I think it's a good one, helps to remove the guesswork to find the config.

PR #1402, opened 2023-01-25
Add `game over [player]` block to multiplayer extension

eanders-ms:
Adds this block to the multiplayer extension:

image

It overrides the active scoring method to explicitly set who won the game.

Resolves https://github.com/microsoft/pxt-arcade/issues/5559

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -82,10 +82,10 @@ namespace game {
             else this.loseMessage = message;
             if (explicit) this.messageSetByUser = true;
         }
-        getMessage(win: boolean) {
+        getMessage(win: boolean, preferMultiplayer?: boolean) {
             if (this.messageSetByUser)
                 return win ? this.winMessage : this.loseMessage;
-            else if (info.multiplayerScoring() && this.scoringType !== ScoringType.None)
+            else if (preferMultiplayer)
                 return win ? this.winMessageMultiplayer : this.loseMessage;
             else
                 return win ? this.winMessage : this.loseMessage;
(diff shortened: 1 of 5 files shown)


libs/game/game.ts
```diff
@@ -332,19 +336,20 @@ namespace game {
...
+            const judged = !winnerOverride && goc.scoringType !== ScoringType.None;
             const playersWithScores = info.playersWithScores();
             const prevBestScore = judged && info.highScore();
             const winner = judged && win && info.winningPlayer();
             const scores = playersWithScores.map(player => new GameOverPlayerScore(player.number, player.impl.score(), player === winner));
 
-            // Save all scores as relevant to the game.
-            if (winner) {
+            // Save scores if this was a judged game and there was a winner (don't save in the LOSE case).
+                if (judged && winner) {
```
jwunderl, line 346:
```suggestion
            if (judged && winner) {
```

libs/game/textDialogs.ts
```diff
@@ -494,41 +516,48 @@ namespace game {
...
+                        // Multiplayer special case: Only one player scored
+                        const score = scores[0];
                         score.str = "P" + score.player + ":" + score.value;
                         this.image.printCenter(
                             score.str,
                             currY,
                             screenColor(1),
                             image.font5
                         );
+                        if (score.winner) {
```
jwunderl, line 532: I probably should have thought of this before but for multiplayer, what should we do if the two 'best' players have the same score? Doesn't really matter for this new block as we're explicitly passing in the winner but I guess we probably need a 'these 2 / 3 / all players tied' message in case of ties? Fine to pass this off / wouldn't be part of this pr anyways

jwunderl (approved):
Nice!

libs/game/textDialogs.ts
eanders-ms: Ties are not supported at this point. Will file an issue to track it.

PR #1407, opened 2023-01-27
hook up player leave / join to multiplayer state

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5193

event ids are a little tricky, I guess we just do "roll a dice and hope it doesn't collide" after checking `dal.d.ts` (e.g. https://github.com/microsoft/pxt-common-packages/pull/1119/files#diff-3446560eec31e791faeaee0c3b08bfefebbb06df29fd92a46bd5054576a1c094R3 is the other 'recent' sim only one that comes to mind)

quick test build: https://arcade.makecode.com/app/d83a329c52d33eb16de47b9b06193c1f1db0a87f-e53085d68d--multiplayer?host=_Rkud4LTboCqc, only host really matters for testing / it's fine if guests are just on beta

image

pxt side over in https://github.com/microsoft/pxt/pull/9332

### DIFF
--- a/libs/game/multiplayer.ts
+++ b/libs/game/multiplayer.ts
@@ -25,6 +25,8 @@ namespace multiplayer {
         game.pushScene();
     }
 
+    const MULTIPLAYER_PLAYER_JOINED_ID = 3241;
+    const MULTIPLAYER_PLAYER_LEFT_ID = 3242;
     export function initServer() {
         if (getOrigin() === "server") {
             game.eventContext().registerFrameHandler(scene.MULTIPLAYER_POST_SCREEN_PRIORITY, () => {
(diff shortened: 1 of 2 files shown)


libs/game/multiplayer.ts
```diff
@@ -25,14 +25,53 @@ namespace multiplayer {
...
+    const MULTIPLAYER_PLAYER_LEFT_ID = 3242;
     export function initServer() {
         if (getOrigin() === "server") {
             game.eventContext().registerFrameHandler(scene.MULTIPLAYER_POST_SCREEN_PRIORITY, () => {
                 if (getOrigin() === "server") {
                     postImage(screen);
                 }
             })
+
+            for (let p = 1; p <= 4; p++) {
```
eanders-ms, line 38: 4 is hard-coded in so many places...

eanders-ms (approved):
LG! Something to think about: How would we implement a "is player connected" block?

libs/game/multiplayer.ts
jwunderl: Yeahh, we should probably do a pass and add a multiplayer.MAX_PLAYERS or something.

And for the other point, should just be adding a helper block for reading out https://github.com/microsoft/pxt-common-packages/blob/af1990f29230881ea56822fa9b5323d02334fde9/libs/game/controller.ts#L227

PR #1408, opened 2023-01-27
fix to support legacy set game over sound call

eanders-ms:


libs/game/game.ts
```diff
@@ -260,12 +260,19 @@ namespace game {
     //% weight=80
     //% blockGap=8
     //% help=game/set-game-over-sound
-    export function setGameOverSound(win: boolean, sound: music.Playable, looping: boolean) {
+    export function setGameOverSound2(win: boolean, sound: music.Playable, looping: boolean) {
```
abchatra, line 263: setGameOverSoundLoop may be a better API name? 

Also can't the third parameter be optional?

eanders-ms: Encoding the meaning of some parameters in the function name is a convenient way to differentiate from the call you're replacing, but the result is an odd api design.

Making the parameter optional might be a better way to fix this. Let me try that.

eanders-ms: @abchatra, Optional param won't work because the type of the 2nd parameter changed as well. @jwunderl , @riknoll do you have a preference on how to name functions that supersede older ones? I'm preferential to adding a number on the end. I feel that encoding additional meaning into the new function name that wouldn't have been there otherwise creates a less understandable API over time. Whereas a "revision number" appended is more generally applicable to more methods, and more scalable over time as functions are rev'ed multiple times.

jwunderl (approved):

PR #1415, opened 2023-01-28
null protection for sprite destroy

eanders-ms:
Let's eliminate one of the leading causes of the dread "sim error: failed cast on null".

If this error message were at all helpful, we might choose to leverage it as a teaching moment. But as it stands, it only serves as a blocking condition a beginner will hit and have no recourse but to walk away from Arcade.

### DIFF
--- a/libs/game/sprites.ts
+++ b/libs/game/sprites.ts
@@ -62,7 +62,9 @@ namespace sprites {
 
     //% group="Effects"
     //% weight=80
-    //% blockId=spritedestroy2 block="destroy $sprite=variables_get(mySprite) || with $effect effect for $duration ms"
+    //% blockId=spritedestroy2 block="destroy $sprite || with $effect effect for $duration ms"
+    //% sprite.shadow=variables_get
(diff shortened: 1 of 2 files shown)


libs/game/sprites.ts
```diff
@@ -60,6 +60,17 @@ namespace sprites {
         return sprites.create(img, kind);
     }
 
+    //% group="Effects"
+    //% weight=80
+    //% blockId=spritedestroy2 block="destroy $sprite=variables_get(mySprite) || with $effect effect for $duration ms"
```
jwunderl, line 65:
```suggestion
    //% blockId=spritedestroy2 block="destroy $sprite || with $effect effect for $duration ms"
    //% sprite.shadow=variables_get
    //% sprite.defl=mySprite
```

jwunderl (approved):
Was just thinking about doing this the other day surprisingly, nice!

eanders-ms:
@abchatra if you would allow this in the hotfix, I will cherry pick it today.

PR #1416, opened 2023-01-29
add "array of all players" block

eanders-ms:
Also reorg'd the top of the Multiplayer drawer to surface the more useful utilities.
image

Fixes https://github.com/microsoft/pxt-arcade/issues/5600

### DIFF
--- a/libs/multiplayer/player.ts
+++ b/libs/multiplayer/player.ts
@@ -409,8 +409,8 @@ namespace mp {
     //% blockId=mp_getPlayerSprite
     //% block="$player sprite"
     //% player.shadow=mp_playerSelector
-    //% group=Sprites
-    //% weight=100
+    //% group=Player
+    //% weight=80
     //% blockGap=8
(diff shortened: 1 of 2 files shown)


libs/multiplayer/player.ts
```diff
@@ -438,6 +438,58 @@ namespace mp {
...
+     */
+    //% blockId=mp_allPlayers
+    //% block="array of all players"
+    //% group=Player
+    //% weight=90
+    //% blockGap=8
+    //% help=multiplayer/get-all-players
+    //% parts="multiplayer"
+    export function allPlayers(): Player[] {
+        return _mpstate().players;
```
jwunderl, line 469: maybe worth a slice to lessen the chances of anyone accidentally mutating it? I guess `_mpstate` is exported anyways so it'd just be to protect users in blocks from accidents~


```suggestion
        return _mpstate().players.slice();
```

jwunderl (approved):

PR #1424, opened 2023-02-03
Fix for Kiosk: Always save scores on game over

eanders-ms:


jwunderl (approved):

eanders-ms:
@kiki-lee if the carnival extension's game over called `info.saveAllScores();`, it should fix the issues running in Kiosk.

kiki-lee:
Rock on!  I do believe it does.

PR #1428, opened 2023-02-10
Better support for playing songs in the browser sim

riknoll:
This PR requires https://github.com/microsoft/pxt/pull/9381

Fixes https://github.com/microsoft/pxt-arcade/issues/5675

This separates playback of songs between hardware and the browser sim. On hardware, nothing changes (except the bug I fixed). The simulator now uses the same sequencer used by the music editor, which has much more accurate timing.

Here is a program that illustrates the issue:

https://arcade.makecode.com/S22715-41600-35351-76629

Try playing that in the sim and then opening the code and playing it in the music editor. The sim is significantly slower and drifts over time.

While I was at it, I also added some events for the browser simulator that I thought would be useful.

Here's an upload target with these changes: https://arcade.makecode.com/app/abc67de31823239df2cb53cc6844fdecb7f10558-06911e4696'

### DIFF
--- a/libs/mixer/instrument.ts
+++ b/libs/mixer/instrument.ts
@@ -133,18 +133,36 @@ namespace music.sequencer {
         }
 
         play(playbackMode: PlaybackMode) {
-            if (currentSequencer) currentSequencer.stop();
-            currentSequencer = new sequencer.Sequencer(this);
+            if (control.deviceDalVersion() === "sim") {
+                const seq = new _SimulatorSequencer();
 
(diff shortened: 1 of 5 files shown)


riknoll:
Build is going to fail until that pxt pr is merged/released

jwunderl (approved):
Just to confirm, these still pipe through playInstructionsAsync, correct? It looks like it but wanted to double check as multiplayer audio hooks in there https://github.com/microsoft/pxt/pull/9165

riknoll:
yes!

PR #1435, opened 2023-02-22
[Stable10.3] Cherry pick sim music fix

riknoll:
* Better support for playing songs in the browser sim

* upate core

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "libs/**/*"
   ],
   "dependencies": {
-    "pxt-core": "8.4.16"
+    "pxt-core": "8.6.10"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "libs/**/*"
   ],
   "dependencies": {
-    "pxt-core": "8.4.16"
+    "pxt-core": "8.5.31"
```
abchatra, line 27: Major version change?

abchatra (approved):
Why do we use pxt-core in pxt-common-packages?

riknoll:
the simulator implementations in pxt-common-packages reference code in pxt/pxtsim/

it's just for the builds in this repo, though, when we build targets we use the version that the target uses

PR #1436, opened 2023-02-27
Multiple score handlers per PlayerState

BenVillalobos:
Fixes https://github.com/microsoft/pxt-arcade/issues/5725.

## Context
Today, if you have multiple "on score" blocks, **the last one wins**. Here, on score 6 never runs because onScore 5 is the last one defined after converting to javascript.
image

[This is because `PlayerState` only contains a single `ScoreReachedHandler`.](https://github.com/microsoft/pxt-common-packages/blob/054b7c4576147e37c6f2ce823a495349e9ac62cc/libs/game/info.ts#L36)

## Changes Made
- `PlayerState` now has an array of `ScoreReachedHandler`s.
- If multiple handlers are added for the same score, last one wins (mimics multiplayer score handler behavior).

## ScoreManager?
There are two places now that handle this "last one wins" behavior. If for some reason we wanted to enable multiple handlers for the same score, we'd need to remember to change two places of code.

Ideally there would be one place in some `ScoreManager` class where you call `AddHandlerForScore` that determines this behavior. The `MPState ` and `PlayerState` would both have this `ScoreManager `and call its functions.

All in all, this is low pri and it's probably too soon to create a `ScoreManager`. I'll create an issue on it for now.

@riknoll suggested putting this into infostate instead of its own manager class because of how state is tracked with the scene stack.

### DIFF
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -33,7 +33,7 @@ namespace info {
         // null: reached 0 and callback was invoked
         public life: number;
         public lifeZeroHandler: () => void;
-        public scoreReachedHandler: ScoreReachedHandler[];
+        public scoreReachedHandlers: ScoreReachedHandler[];
 
         public showScore?: boolean;
         public showLife?: boolean;
@@ -45,7 +45,7 @@ namespace info {
             this.showScore = undefined;
             this.showLife = undefined;
             this.showPlayer = undefined;
-            this.scoreReachedHandler = [];
+            this.scoreReachedHandlers = [];
         }
     }
 
@@ -712,7 +712,7 @@ namespace info {
             const oldScore = state.score || 0;
             state.score = (value | 0);
 
-            state?.scoreReachedHandler.forEach(srh => {
+            state.scoreReachedHandlers.forEach(srh => {
                 if ((oldScore < srh.score && state.score >= srh.score) ||
                     (oldScore > srh.score && state.score <= srh.score)) {
                     srh.handler();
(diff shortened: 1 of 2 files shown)


libs/game/info.ts
```diff
@@ -711,12 +712,12 @@ namespace info {
             const oldScore = state.score || 0;
             state.score = (value | 0);
 
-            if (state.scoreReachedHandler && (
-                (oldScore < state.scoreReachedHandler.score && state.score >= state.scoreReachedHandler.score) ||
-                (oldScore > state.scoreReachedHandler.score && state.score <= state.scoreReachedHandler.score)
-            )) {
-                state.scoreReachedHandler.handler();
-            }
+            state?.scoreReachedHandler.forEach(srh => {
```
jwunderl, line 715: Static typescript code (~~user level code) is compiled using [typescript 2.6](https://github.com/microsoft/pxt/tree/master/pxtcompiler/ext-typescript), so can't use optional chaining unfortunately as that wasn't added tille like 3.7 or so

libs/game/info.ts
```diff
@@ -766,7 +767,14 @@ namespace info {
 
         onScore(score: number, handler: () => void) {
             const state = this.getState();
-            state.scoreReachedHandler = new ScoreReachedHandler(score, handler);
+            state.scoreReachedHandler?.forEach(element => {
+                if (element.score === score) {
+                    // Score handlers are implemented as "last one wins."
+                    element.handler = handler;
+                    return;
```
jwunderl, line 774: I don't think this will work as you intend, this will return in the anonymous fn for that value but still iteratw through the rest (and, more importantly, still push onto the array after the foreach is done)

Try a `for (const element of state.scoreReachedHandler){` instead of `[].forEach`and same code should work correctly

jwunderl (commented):
Two quick things to fix (no ?. allowed here and sus return) but other than that looks good!

libs/game/info.ts
BenVillalobos: thanks for the catch!

jwunderl (approved):
Nice!

PR #1438, opened 2023-03-17
Add `Color` Enum

BenVillalobos:
### Problem
See this snippet from one of the MakeCode tutorials

```ts
function startGame() {
    // Set the background color to cyan
    scene.setBackgroundColor(9);
```

You could plug in numbers to see what happens, but that isn't a good experience. So I added a color enum to make our javascript "onboarding experience" smoother:

image

cc @riknoll the templates are going to be super helpful, so it'd be nice to update your template project to include the enum when this eventually gets merged.

### Testing
Tested by running arcade locally

### DIFF
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -17,6 +17,10 @@ namespace game {
         None
     }
 
+    /**
+     * The available colors for Arcade.
+     * NOTE: If the color pallette is changed, these values will change along with it.
+     */
     export enum Color {
         transparent = 0,
         white = 1,
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -17,11 +17,11 @@ namespace game {
         None
     }
 
+    // To stay synchronized with https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/ImageEditor/sprite/Palette.tsx#L98.
     /**
      * The available colors for Arcade.
-     * To stay synchronized with https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/ImageEditor/sprite/Palette.tsx#L98.
-     * NOTE: If the color pallette is changed, these values will change along with it.
-     */
+     * NOTE: If the color palette is changed, these values will change along with it.
+     **/
     export enum Color {
         Transparent = 0,
         White = 1,
--- a/libs/game/game.ts
+++ b/libs/game/game.ts
@@ -19,6 +19,7 @@ namespace game {
 
     /**
      * The available colors for Arcade.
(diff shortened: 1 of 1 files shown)


BenVillalobos:
The comment could probably be improved, but I added a note mentioning the color pallette. Not sure there's a better way to get that info across in the API.

libs/game/game.ts
```diff
@@ -17,6 +17,29 @@ namespace game {
...
+        teal = 6,
+        green = 7,
+        blue = 8,
+        lightblue = 9,
+        purple = 0xa,
+        lightpurple = 0xb,
+        darkpurple = 0xc,
+        tan = 0xd,
+        brown = 0xe,
+        black = 0xf
```
jwunderl, line 40: Enum values are typically pascal cased
```suggestion
        Transparent = 0,
        White = 1,
        Red = 2,
        Pink = 3,
        Orange = 4,
        Yellow = 5,
        Teal = 6,
        Green = 7,
        Blue = 8,
        LightBlue = 9,
        Purple = 0xa,
        LightPurple = 0xb,
        DarkPurple = 0xc,
        Tan = 0xd,
        Brown = 0xe,
        Black = 0xf
```

libs/game/game.ts
```diff
@@ -17,6 +17,29 @@ namespace game {
         None
     }
 
+    /**
+     * The available colors for Arcade.
+     * NOTE: If the color pallette is changed, these values will change along with it.
```
jwunderl, line 22: maybe a pointer here as this is the other place I listed these out https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/ImageEditor/sprite/Palette.tsx#L98

libs/game/game.ts
```diff
@@ -17,6 +17,30 @@ namespace game {
         None
     }
 
+    /**
+     * The available colors for Arcade.
+     * To stay synchronized with https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/ImageEditor/sprite/Palette.tsx#L98.
+     * NOTE: If the color pallette is changed, these values will change along with it.
+     */
```
jwunderl, line 24:
```suggestion
    // To stay synchronized with https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/ImageEditor/sprite/Palette.tsx#L98.
    /**
     * The available colors for Arcade.
     * NOTE: If the color palette is changed, these values will change along with it.
     **/
```

nit: lets swap this so the dev pointer is a separate comment, as is it will show up in hovertext

image
vs.
image

also nit: pallette -> palette~

jwunderl (approved):
LGTM

PR #1439, opened 2023-03-28
Fill Triangle/Polygon4 in yield mode.

AqeeAqee:

PR #1440, opened 2023-03-30
Triangle fill / polygon fill from aqeeaqee

jwunderl:
I mirrored over the code from the cpp to the sim side in https://github.com/microsoft/pxt-common-packages/commit/9b878f8bf9ded9d9f372269be2e881270a6127b5 and https://github.com/microsoft/pxt-common-packages/commit/45d18049986d57511cb780147e77cf22049284bd + one more coercing out decimal places, left the cpp as is besides a few more nitpicks on spacing that I had missed before. CPP side already reviewed in https://github.com/microsoft/pxt-common-packages/pull/1439

here's richard's pretty test: 

trianglz

https://arcade.makecode.com/app/7695d0da8f8e68416984aa13f6e208cc78d11849-a60c32b437#pub:S49613-21549-53722-36190

(I haven't actually run the polygon4 one yet as I'm going to make dinner / didn't have a test case handy)

It's not letting me explicitly add you as a reviewer but I'd love it if you checked over / gave any feedback if you have any / gave it a thumbs up if you have time @aqeeaqee! Thanks again for the great pr, and I saw that you were working on a few more tweaks; feel free to the changes against this branch still (or we can approve / merge this in and you can make them against the master branch if preferred!)

### DIFF
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -1159,8 +1159,193 @@ void _fillCircle(Image_ img, int cxy, int r, int c) {
     fillCircle(img, XX(cxy), YY(cxy), r, c);
 }
 
+typedef struct 
+{
+    int x, y;
+    int x0, y0;
+    int x1, y1;
+    int W,H;
+    int dx, dy;
+    int yi, xi;
+    int D;
+    int nextFuncIndex;
+}LineGenState; // For keeping track of the state when generating Y values for a line, even when moving to the next X.
+
+typedef struct
+{
+    int min;
+    int max;
+}ValueRange;
+
+void nextYRange_Low(int x, LineGenState *line, ValueRange *yRange) {
+    while (line->x == x && line->x <= line->x1 && line->x < line->W) {
+        if (0 <= line->x) {
+            if (line->y < yRange->min) yRange->min = line->y;
+            if (line->y > yRange->max) yRange->max = line->y;
+        }
+        if (line->D > 0) {
+            line->y += line->yi;
+            line->D -= line->dx;
+        }
+        line->D += line->dy;
+        ++line->x;
+    }
+}
+
+void nextYRange_HighUp(int x, LineGenState *line, ValueRange *yRange) {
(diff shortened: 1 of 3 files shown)


riknoll (approved):

AqeeAqee:
Tested the perf with [this project](https://arcade.makecode.com/S80801-48202-76007-39839).
😲 So amazing is so fast! About 0.03ms each triangle, 0.087ms each polygon4.
As I remembered it takes serveral ms when I write in TypeScript, so I migrate codes in sim/image.ts back to this project, still 2~4ms. So why it can be faster 100 times in sim with almost the same codes❓

riknoll:
@AqeeAqee that's the difference between "user code" and "native code". I put that in quotes because it's not *really* native in this case... A more precise description would be code that passes through our compiler (user code) and code that does not (native code).

When we compile user code, we don't just run it directly. Instead we compile it into a special format that lets us mimic the behaviors of the hardware runtime, such as multiple threads. You can actually see this if you expand the "built" section of the file explorer in the JavaScript editor and open binary.js; that's the code that is actually being run by the simulator. The downside is that there's a performance tradeoff.

jwunderl:
This is running the operations as native, browser javascript operations; in the project code, we compile down to a binary.js file and run that to make our behavior match as closely as possible on hardware and in the browser (if you go to the javascript editor, in the file explorer after you run your code you'll see it under `built/binary.js`) -- this matches pretty closely with output we use for hardware (e.g. 
image
), which is good because it means the project will run the same in both the browser and on hardware, and most of the time doesn't make a huge difference (/ computers are fast), but when you're doing 100s or 1000s of math operations it ends up with tons of extra code produced to match -- instead of just calling a function, it ends up emulating pushing onto registers, calling, cleaning up, etc.

jwunderl:
@AqeeAqee your code's now live in https://arcade.makecode.com/beta v1.13.12:

image

Thanks again!

AqeeAqee:
Great! Try it later.
new PR is on the way of final test and format.
BTW, found fillCircle() has the float issue you mentioned, if you set R a float number.

PR #1441, opened 2023-03-30
fillTriangle/Polygon() use new added drawVLineCore()

AqeeAqee:
Add drawVLine() as well.

### DIFF
--- a/libs/screen/image.cpp
+++ b/libs/screen/image.cpp
@@ -1267,6 +1267,46 @@ LineGenState initYRangeGenerator(int16_t X0, int16_t Y0, int16_t X1, int16_t Y1)
     }
 }
 
+// core of draw vertical line for repeatly calling, eg.: fillTriangle() or fillPolygon4()
+// value range/safety check not included
+// prepare "img->makeWritable();" and "uint8_t f = img->fillMask(c);" outside required.
+// bpp=4 support only right now
+void drawVLineCore(Image_ img, int x, int y, int h, uint8_t f) {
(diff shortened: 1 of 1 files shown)


AqeeAqee:
This version faster about 30%+ than calling fillRect().
BPP=4 only right now, add BPP=1 later if need support ?
fillCircle can faster 25%+ by calling DrawVLine() also, update later if needed.

libs/screen/image.cpp
```diff
@@ -1267,6 +1267,46 @@ LineGenState initYRangeGenerator(int16_t X0, int16_t Y0, int16_t X1, int16_t Y1)
...
+    uint8_t f = img->fillMask(c);
+    if (x < 0 || x >= width(img) || y >= H || y+h-1 < 0)
+        return;
+    if (y < 0){
+        h += y;
+        y = 0;
+    }
+    if (y + h > H)
+        h = H - y;
+    drawVLineCore(img,x,y,h,f);
```
jwunderl, line 1307:
```suggestion
    int H = height(img);
    uint8_t f = img->fillMask(c);
    if (x < 0 || x >= width(img) || y >= H || y + h - 1 < 0)
        return;
    if (y < 0){
        h += y;
        y = 0;
    }
    if (y + h > H)
        h = H - y;
    drawVLineCore(img, x, y, h, f);
```

jwunderl (approved):
Looks good to me, shouldn't need any change /sim side as it's just optimizing the hot path a bit and sim is plenty fast enough as is :)

riknoll (approved):

PR #1446, opened 2023-04-25
Bump animation above advanced

abchatra:


abchatra:
image

libs/game/animation.ts
```diff
@@ -3,7 +3,7 @@
 */
 //% color="#03AA74" weight=100 icon="\uf021" block="Animation"
 //% groups='["Animate", "Advanced"]'
-//% advanced=true
+//% weight = 5
```
riknoll, line 6:
```suggestion
//% weight=5
```

riknoll (approved):
LGTM! Just remove spaces around the equals

jwunderl (approved):
Change looks good, though we'll need to get galen to check how many tutorials need to get updated & update them to remove any 'go to advanced' type wording

abchatra:
Yes, this is for next release, so we don't forget. We will have to update docs closer to the release than now. 

I intend to merge and test in /beta and see first before changing the docs

PR #1448, opened 2023-05-25
Fix radio send string to allow 19 chars

kimprice:
Fixes https://github.com/microsoft/pxt-microbit/issues/4849

A while back, when rewriting radio from C++ code to Typescript (https://github.com/microsoft/pxt-microbit/pull/2000/files), we accidentally subtracted 2 instead of 1 from the max payload for sending strings. We only need to adjust by 1 so we have room to store the length. 

I tested this change on hardware as well and everything is working fine.

### DIFF
--- a/libs/radio/radio.ts
+++ b/libs/radio/radio.ts
@@ -395,7 +395,7 @@ namespace radio {
     function getMaxStringLength(packetType: number) {
         switch (packetType) {
             case PACKET_TYPE_STRING:
-                return MAX_PAYLOAD_LENGTH - 2;
+                return MAX_PAYLOAD_LENGTH - 1;
             case PACKET_TYPE_VALUE:
             case PACKET_TYPE_DOUBLE_VALUE:
                 return MAX_FIELD_DOUBLE_NAME_LENGTH;


riknoll:
already spoke to kim about this, but i'm the one who checked this in and i have no memory of why i made this change. could be a mistake!

kimprice:
I checked the micropython reference and there is nothing that specifies how long the message can be to send a string.
https://python.microbit.org/v/3/reference/radio

I ran the tests that were checked in with https://github.com/microsoft/pxt-microbit/pull/2000/files and I couldn't get it to work in the simulator (it would hang on the first test), but it worked just fine on hardware. Here's a share link of that: https://makecode.microbit.org/_8HF4auYs2P49

I also ran some tests manually and it worked fine in the simulator and on hardware - sending the 19 chars as expected. Here's a share link for that: https://makecode.microbit.org/_HRs5MhWwHLuT

It's also worth noting that in our [documentation](https://makecode.microbit.org/reference/radio/send-string) we say the maximum length is 19 chars, but in reality, it's 19 bytes. I'm not sure if that's worth updating or if that's more confusing. I guess the max is 19, but it could be less if your chars aren't single bytes.

jwunderl (approved):

PR #1450, opened 2023-06-16
Add shader package

riknoll:
This adds a new package called shader that has some useful functions for shading:

1. mapImage - similar to mapRect except that it remaps colors according to an image. It takes in a buffer of color mappings and each color in the map image will point to a different section of the buffer, allowing you to shade the destination image by different amounts.
2. mergeImage - takes two images and finds the minimum pixel value at each coordinate. Useful for merging multiple light sources into one "lightSpace" that you can shade all at once

### DIFF
--- a/libs/shader/sim/shader.ts
+++ b/libs/shader/sim/shader.ts
@@ -1,0 +1,37 @@
+/// <reference path="../../screen/sim/image.ts" />
+
+namespace pxsim.ShaderMethods {
+    export function _mergeImage(dst: RefImage, src: RefImage, xy: number) {
+        mergeImage(dst, src, pxsim.ImageMethods.XX(xy), pxsim.ImageMethods.YY(xy));
+    }
+
+    function mergeImage(dst: RefImage, src: RefImage, x0: number, y0: number) {
+        for (let x = 0; x < src._width; x++) {
+            for (let y = 0; y < src._height; y++) {
+                pxsim.ImageMethods.setPixel(
+                    dst,
+                    x0 + x,
+                    y0 + y,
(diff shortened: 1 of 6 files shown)


jwunderl (approved):

libs/shader/sim/shader.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference path="../../screen/sim/image.ts" />
+
+namespace pxsim.ShaderMethods {
+    export function _mergeImage(dst: RefImage, src: RefImage, xy: number) {
+        mergeImage(dst, src, pxsim.ImageMethods.XX(xy), pxsim.ImageMethods.YY(xy));
+    }
+
+    function mergeImage(dst: RefImage, src: RefImage, x0: number, y0: number) {
```
eanders-ms, line 8: Just wondering, why reimplement this? Could you call `pxsim.ImageMethods.mergeImage/mapImage` here instead?

eanders-ms (approved):

libs/shader/sim/shader.ts
riknoll: this is the simulator implementation!

riknoll: these apis are the ones i'm adding, so they don't already exist in imagemethods

eanders-ms: Would this be equivalent?
```ts
    function mergeImage(dst: RefImage, src: RefImage, x0: number, y0: number) {
        pxsim.ImageMethods.mergeImage(dst, src, x0, y0);
    }
```

PR #1451, opened 2023-06-21
update checkout, setup-node actions

jwunderl:
see https://github.com/microsoft/pxt/pull/9550; we're getting a spurious diagnostic on every build 
image
that appeared to be causing some build inconsistencies on occasion (in the build linked from that pr, it's my best guess as the random build error can come from wrong node version being used).

will do the same pr over in microbit, minecraft, arcade for now~

### DIFF
--- a/.github/workflows/pxt-buildmain.yml
+++ b/.github/workflows/pxt-buildmain.yml
@@ -17,9 +17,9 @@ jobs:
         node-version: [8.x]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@main
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@main
         with:
           node-version: ${{ matrix.node-version }}
       - name: npm install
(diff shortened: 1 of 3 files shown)


.github/workflows/pxt-buildmain.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         node-version: [8.x]
```
jwunderl, line 17:
```suggestion
        node-version: [16.x]
```

.github/workflows/pxt-buildpr.yml
```diff
@@ -12,9 +12,9 @@ jobs:
         node-version: [8.x]
```
jwunderl, line 12:
```suggestion
        node-version: [16.x]
```

.github/workflows/pxt-buildpush.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         node-version: [8.x]
```
jwunderl, line 17:
```suggestion
        node-version: [16.x]
```

eanders-ms (approved):

PR #1455, opened 2023-08-01
remove dup merge image in sim/image.ts

jwunderl:
in https://github.com/microsoft/pxt-common-packages/pull/1450 looks like these got duped by accident when they were moved out to a separate package


eanders-ms (approved):
LGTM

PR #1457, opened 2023-09-26
Update music api reference for playable and sound effect apis

ganicke:
Set the help paths, add jsDoc playable, add the missing reference pages.


Jaqster (approved):
Thanks Galen!

PR #1458, opened 2023-10-11
Add icon blit helper function

riknoll:
Adds a helper method for using the blit function with icons.

Icons in arcade are 1 bit per pixel, so they don't contain any color information. Our other icon methods take a color argument to use for the non-transparent pixels; this simply adds a copy of blit that passes such an argument.

I need this for my fancy text extension (or, rather, for my mini-menu extension which I am converting to use the fancy-text extension).


eanders-ms (approved):
LGTM! Conditionals in the inner loop aren't ideal, but something we can optimize later if the need arises.

jwunderl (approved):
LGTM, as eric said bit of a hotpath so maybe worth quick hardware tests to double check but can defer that till later too

abchatra:
@riknoll ?

PR #1467, opened 2024-01-05
Add keyboard and mouse events extension

riknoll:
This PR adds a browser-only extension that enables support for keyboard and mouse events.

There are two more PRs for this, one in pxt-arcade and one in pxt-arcade-sim.

The API for keyboard keys treats each key as an individual controller button. There is no "get last character typed" block because this API is not meant for typing text, just letting the player have extra button options (typing in our sim would be super cumbersome anyhow given that we have keys that map to special functions like taking a screenshot).

Also, as an added safety feature all keyboard and mouse events are disabled in mutliplayer (the code doing the disabling is in pxt-arcade-sim).

image

image

### DIFF
--- a/libs/browser-events/browserEvents.cpp
+++ b/libs/browser-events/browserEvents.cpp
@@ -1,0 +1,29 @@
+#include "pxt.h"
+
+namespace browserEvents {
+
+//%
+int mouseX() {
+    return -1;
+}
+
+//%
+int mouseY() {
+    return -1;
+}
+
+//%
+int wheelDx() {
+    return -1;
+}
+
+//%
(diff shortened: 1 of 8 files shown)


eanders-ms:
I see cpp additions. How does this affect the v1 hex file size? Can we regression test hex files that were at the limit before?

eanders-ms:
> I see cpp additions. How does this affect the v1 hex file size? Can we regression test hex files that were at the limit before?

Oh I guess this isn't for microbit. nevermind!

Brohann3214:
Can you add a block to show/hide the mouse pointer?

PR #1469, opened 2024-01-08
Add support for system keyboard in string/number prompts

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/4845

Adds an optional "use system keyboard" parameter for the "ask for string" and "ask for number" APIs that allows you to type using your actual keyboard and not the in-game one. There will be a corresponding PR in pxt-arcade-sim that has the simulator side of this.

This will require an upgrade rule. Or, if we want, we can deprecate this block and make a new one that always pops the system keyboard.

### DIFF
--- a/libs/game/numberprompt.ts
+++ b/libs/game/numberprompt.ts
@@ -4,19 +4,21 @@ namespace game {
      * Ask the player for a number value.
      * @param message The message to display on the text-entry screen
      * @param answerLength The maximum number of digits the user can enter (1 - 10)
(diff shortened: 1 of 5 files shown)


Jaqster:
OMG!  Amazing!!!  Thank you Richard.  This has been a big pain point for teachers getting through the "User Input" curriculum unit.  Much appreciated.

Jaqster (approved):

kimprice (approved):
LGTM!

PR #1471, opened 2024-02-07
Add support for tilemaps with 4x4 tiles

riknoll:
The pxt-common-packages part of https://github.com/microsoft/pxt/pull/9853

This adds no new blocks, but opens up the ability for me to add a block in an extension (like tile util)


jwunderl (approved):

riknoll:
hehehe looks like you can occasionally break through walls in 4x4 tilemaps! let's still merge this since it isn't surfacing anything and we can fix it later.

PR #1480, opened 2024-07-15
Game over block: made saveAllScores take the scoring method

srietkerk:
Needed for: https://github.com/microsoft/pxt/pull/10041

This is a small change to the game over block to include the scoring method in the event that game over sends to the simulator. The event that was modified only gets used by the kiosk.

### DIFF
--- a/libs/game/info.ts
+++ b/libs/game/info.ts
@@ -241,7 +241,7 @@ namespace info {
         return players ? players.filter(item => item.impl.hasScore()) : [];
     }
 
-    export function saveAllScores() {
+    export function saveAllScores(scoringType: string) {
         const allScoresKey = "all-scores";
         let allScores: number[];
         const pws = playersWithScores();
@@ -252,7 +252,12 @@ namespace info {
             allScores = [];
         }
 
-        settings.writeJSON(allScoresKey, allScores);
+        const scoresObj = {
+            "allScores": allScores,
+            "scoringType": allScores.length ? scoringType : "None"
+        }
+
+        settings.writeJSON(allScoresKey, scoresObj);
     }
 
     export function winningPlayer(): PlayerInfo {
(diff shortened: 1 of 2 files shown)


libs/game/info.ts
```diff
@@ -252,7 +252,12 @@ namespace info {
             allScores = [];
         }
 
-        settings.writeJSON(allScoresKey, allScores);
+        const scoresObj = {
+            "allScores": allScores,
+            "scoringType": allScores.length ? scoringType : "None"
+        }
+
+        settings.writeJSON(allScoresKey, scoresObj);
```
thsparks, line 260: I'm not totally familiar with how setting/reading these settings works, but will we hit compatibility issues with games that are shared before the update which someone then tries to play or import into kiosk after the update?

If it's just an issue with importing into kiosk, that may be fine. But if it somehow breaks the game loading at all, that'd probably be an issue.

thsparks: If it does cause issues, a potential fix could be to save this with a different key and change the loading logic to account for both (giving this one priority)

srietkerk: I might be misunderstanding your question, but I made sure to have backwards compatibility on the kiosk side. If the JSON that is received by the kiosk doesn't have an entry for "allScores" we just parse the JSON and use the entry like we did before.

srietkerk: Oh, and it's important to note that this key is not being used anywhere else, only for kiosk.

abchatra: When we release arcade, both kiosk page and simulator are released together. I don't see a compat issue even if the formats of communication change. I don't know of any other host using these at this moment other than kiosk

thsparks (approved):

libs/game/info.ts
thsparks: @abchatra I was more concerned with games shared before the update being imported after the update (like games on the home page, for example).

But from Sarah's comments, seems like it'll be fine regardless. Thanks for the info!

PR #1491, opened 2025-02-07
add date APIs to browser events extension

riknoll:
adds date APIs to the browser-events extension. these let you get the current time in milliseconds since the epoch and also break that time into its constituent parts using the local browser/system settings. this is currently javascript only, but i might add blocks later. also this does nothing on hardware (obviously)

this is a frequent request on the forum.

rtc


eanders-ms (approved):
👍

abchatra:
Nice

PR #1494, opened 2025-02-14
fix image variable names and add width/height + draw

riknoll:
a few tweaks to the image category:

* adds a width/height block for images
* adds draw image to image block
* fixes image variable names
* deprecates the image_picker block which is basically just a duplicate of the regular create image block

### DIFF
--- a/libs/screen/image.ts
+++ b/libs/screen/image.ts
@@ -1,6 +1,13 @@
 type color = number
 
 namespace image {
+    export enum Dimension {
+        //% block="width"
+        Width,
+        //% block="height"
+        Height
+    }
+
     export function repeatY(count: number, image: Image) {
         let arr = [image]
         while (--count > 0)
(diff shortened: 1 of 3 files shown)


libs/screen/image.ts
```diff
@@ -24,6 +31,24 @@ namespace image {
         }
         return r
     }
+
+    /**
+     * Returns the width or height of a picture.
+     *
+     * @param picture The picture to get the width or height of
+     * @param dimension The dimension to get
+     * @returns
```
srietkerk, line 40: Was this meant to be filled out?

libs/screen/image.ts
```diff
@@ -24,6 +31,24 @@ namespace image {
...
+     */
+    //% blockId=image_get_dimension
+    //% group="Create"
+    //% blockNamespace="images"
+    //% block="$picture $dimension"
+    //% picture.shadow=variables_get
+    //% picture.defl=picture
```
eanders-ms, line 47: Should the value `picture` be quoted here? That's how it appears where it's been added elsewhere in the PR, e.g.: `//% picture.defl="picture"`

eanders-ms (approved):
Looks good. One minor question.

srietkerk (approved):
Just a question about the doc comment for the new block, but looks good!

libs/screen/image.ts
riknoll: it isn't required! only required if there is a space in the name

PR #1495, opened 2025-02-20
Add system keyboard support to number and string prompts

riknoll:
requires https://github.com/microsoft/pxt-arcade-sim/pull/110

this adds support for requesting strings using the system keyboard so that you can just type responses to the ask for string and ask for number blocks. i also greatly refactored the prompt code and deleted all of the duplicated code in the number prompt class.

see the sim pr for details on how the simulator side of this works. also, this is not supported on hardware.


riknoll:
see also https://github.com/microsoft/pxt-arcade/pull/6700

eanders-ms (approved):
Nice cleanup!

PR #1504, opened 2025-03-11
optimize sprite overlap step

riknoll:
@jwunderl and i were playing with the physics engine on livestream yesterday and discovered that the biggest performance bottleneck in games with lots of overlapping sprites was our SpriteMap's overlaps function. for those who haven't looked at this code before, we use a form of spatial hashing in our overlap detection where we divide the screen/tilemap into buckets and then place each sprite into every bucket it overlaps with. then, during the overlap step in our physics engine we loop over all moving sprites and build up a list of potential overlappers by merging all of the buckets that the sprite is in. this ends up being faster than the N^2 alternative for games with lots of sprites but just barely because merging those buckets is expensive and we end up doing it several times per frame.

this PR restructures that code so that now we just keep track of all the filled buckets as we insert sprites into the map and loop over those buckets instead of looping over the sprites. this eliminates all the bucket merging and results in a _huge_ performance increase. i also now store a list of already checked sprites so that we don't duplicate work

here is the test program i was using:

https://makecode.com/_g2wWrcPgrF8q

i want to do some more detailed testing on different scenarios, but i think this is safe to check in for now.

### DIFF
--- a/libs/game/physics.ts
+++ b/libs/game/physics.ts
@@ -189,7 +189,7 @@ class ArcadePhysicsEngine extends PhysicsEngine {
                 s._x = Fx.add(s._x, stepX);
                 s._y = Fx.add(s._y, stepY);
 
-                if (!(s.flags & SPRITE_NO_SPRITE_OVERLAPS)) {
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

eanders-ms (approved):
💥

PR #1505, opened 2025-03-14
Fix More Category in Radio Extension

thsparks:
Fixes https://github.com/microsoft/pxt-arcade/issues/4860

There was an error in the block definition which caused parsing to fail, so the toolbox wasn't loading correctly. Specifically, the `control_event_source_id` and `control_event_value_id` block ids referenced for the `src` and `value` params do not exist in this repo. (They're specific to micro:bit.)

I've removed the references to those parameters and hidden the block. I didn't want to fully delete the code in case some project somewhere was actually using it, but this way no-one else will try to use it.

TO DO: I have not checked if micro:bit is depending upon this block for its (presumably working) radio implementation.


abchatra:
This would mess up microbit right?

PR #1508, opened 2025-03-24
flip some mouse events

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6875

Also fixes a similar bug mentioned on the forum that I confirmed was a problem: https://forum.makecode.com/t/arcade-2025-update-is-live/34864/4

### DIFF
--- a/libs/browser-events/browserEvents.ts
+++ b/libs/browser-events/browserEvents.ts
@@ -4,8 +4,8 @@
 //% block="Browser Events"
 namespace browserEvents {
     export enum Event {
-        PointerDown = 6857,
-        PointerUp = 6858,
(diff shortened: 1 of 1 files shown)


thsparks (approved):
As a curiosity, is there a chance this is some sort of browser-specific issue? Which browsers did you test on?

srietkerk:
> As a curiosity, is there a chance this is some sort of browser-specific issue? Which browsers did you test on?

I'm not sure about mac, but I tested on Chrome, Edge, and Firefox and this fixes the bug.

abchatra:
@srietkerk port to stable12.2. Seems safe right?

PR #1510, opened 2025-03-26
add rotation property to sprites

riknoll:
this PR adds a new rotation property to sprites. this property works similarly to the two scale properties (sx and sy) but instead of stretching the sprite it rotates it using the "three shears" algorithm. this adds two new properties to the sprite dropdown:

image

they both map to the same thing but obviously use different units. radians is what you want if you're going to be using this with trigonometric functions like sin and cos, but degrees is easier to understand/think about. 

the drawing for the rotation is implemented in c++ with fixed point math and it runs pretty well on hardware, though it's still much slower than non-rotated sprites. however, because we're talking about rotating pixel art here, the results don't always look amazing especially for smaller sprites. for example, here is our friend the duck rotated about 45 degrees:

image

however it looks much better for larger sprites:

image

and even better when in motion, which is what i mainly think this will be used for:

rotating-house

this implementation is also compatible with scaling. scaling is applied *before* rotation, like so:                   

stretchy-house-final

now for some caveats:

1. ~in order to make this performant, **sprite overlaps for rotated sprites are *not* pixel-perfect**. instead, they are calculated based on the rotated rectangle of the sprite. this is really only an issue for sprites that have a lot of transparency in their images~
2. **the tilemap hitbox of the rotated sprite is the axis aligned bounding box for the whole rotated image**. for long sprites, this can look a little weird but it's exactly the same as it would be if you were to draw the rotated image for the sprite. see below for an image of what this looks like

and here is the promised picture of the hitbox for a rotated and scaled sprite. the red box indicates the hitbox:

image

### DIFF
--- a/libs/screen/sim/image.ts
+++ b/libs/screen/sim/image.ts
@@ -956,19 +956,44 @@ namespace pxsim.ImageMethods {
         return v | 0;
     }
 
-    export function _drawScaledRotatedImage(dst: RefImage, src: RefImage, args: RefCollection) {
-        drawScaledRotatedImage(dst, src, args);
+    interface ParsedShearArgs {
+        sx: number;
+        sy: number;
+        scaledWidth: number;
+        scaledHeight: number;
+        minX: number;
+        minY: number;
+        maxX: number;
+        maxY: number;
+        xShear: number;
+        yShear: number;
+        flip: boolean;
     }
 
-    export function drawScaledRotatedImage(dst: RefImage, src: RefImage, args: RefCollection) {
-        const xDst = args.getAt(0) as number;
-        const yDst = args.getAt(1) as number;
-        const sx = ((args.getAt(2) as number) * FX_ONE);
-        const sy = ((args.getAt(3) as number) * FX_ONE);
-        let angle = args.getAt(4) as number;
+    function parseShearArgs(src: RefImage, args: RefCollection, argIndex: number): ParsedShearArgs {
+        const parsed: ParsedShearArgs = {
+            sx: 0,
+            sy: 0,
+            scaledWidth: 0,
+            scaledHeight: 0,
+            minX: 0,
+            minY: 0,
+            maxX: 0,
+            maxY: 0,
+            xShear: 0,
+            yShear: 0,
+            flip: false
+        };
 
-        if (sx <= 0 || sy <= 0 || xDst >= dst._width || yDst >= dst._height) {
-            return;
+        const sx = ((args.getAt(argIndex) as number) * FX_ONE);
+        const sy = ((args.getAt(argIndex + 1) as number) * FX_ONE);
+        let angle = args.getAt(argIndex + 2) as number;
+
+        parsed.sx = sx;
+        parsed.sy = sy;
+
+        if (sx <= 0 || sy <= 0) {
+            return parsed;
         }
 
         angle %= TWO_PI;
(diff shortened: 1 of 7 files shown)


eanders-ms:
I'm concerned that if we don't support per-pixel collision on rotated sprites, it will feel broken. Maybe we could expose an engine setting, allowing users to set bounding-box or per-pixel mode, but the behavior should be consistent across all sprite orientations.

riknoll:
@eanders-ms fair enough, i added pixel perfect overlaps back in.

there are two overlaps implementations: one checks a rotated image against an unrotated image, the other checks two rotated images against each other. i figure the first scenario is going to be much more common than the second so it makes sense to split the two.

libs/screen/sim/image.ts
```diff
@@ -933,6 +933,136 @@ namespace pxsim.ImageMethods {
...
+        const sy = ((args.getAt(3) as number) * FX_ONE);
+        let angle = args.getAt(4) as number;
+
+        if (sx <= 0 || sy <= 0 || xDst >= dst._width || yDst >= dst._height) {
+            return;
+        }
+
+        angle %= TWO_PI;
+        if (angle < 0) {
+            angle += Math.PI;
```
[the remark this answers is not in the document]
eanders-ms: I wondered about this too. Was copilot wrong here?

libs/screen/sim/image.ts
```diff
@@ -933,6 +933,412 @@ namespace pxsim.ImageMethods {
...
+        let flip = false;
+        if (angle > HALF_PI && angle <= THREE_HALF_PI) {
+            flip = true
+            angle = (angle + Math.PI) % TWO_PI
+        }
+
+        const xShear = (-Math.tan(angle / 2) * FX_ONE);
+        const yShear = (Math.sin(angle) * FX_ONE);
+
+        const scaledWidth = src._width * sx;
```
eanders-ms, line 1013: It seems the TS implementation isn't actually using fixed point, but if it were, this line would be mixing fixed and floating point math.

libs/screen/sim/image.ts
```diff
@@ -933,6 +933,412 @@ namespace pxsim.ImageMethods {
...
+            shearedX = fxFloor(shearedX + fxMul(shearedY, shearArgs.xShear));
+        }
+
+
+        if (shearArgs.flip) {
+            for (let y = 0; y < shearArgs.scaledHeight; y += FX_ONE) {
+                for (let x = 0; x < shearArgs.scaledWidth; x += FX_ONE) {
+                    let color = getPixel(
+                        src,
+                        fxToInt(fxDiv((shearArgs.scaledWidth - x - 1), shearArgs.sx)),
```
eanders-ms, line 1162: Mixing fixed and float here? Should this use FX_ONE instead of the literal `1`?

eanders-ms (commented):
Left a few comments. I think I noticed some mixing of fixed and float math. I'd recommend switching to actual fixed point in the TS implementation, to ensure it's consistent. It might make maintaining parity with the cpp implementation easier too (more 1 to 1).

libs/screen/sim/image.ts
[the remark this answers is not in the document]
riknoll: yeah, you're right; not sure why i skipped this. i'll fix it

libs/screen/sim/image.ts
riknoll: this is fine, though, since we're dealing with integers and this would just require unnecessary bit shifting to convert into FX and immediately convert back

libs/screen/sim/image.ts
riknoll: oh yeah, though i fixed all of these when i fixed that in the C++...

riknoll:
@eanders-ms not sure if there is much to gain by switching to fixed point in the browser since you can't accurately replicate the division/multiplication of the hardware because of overflow issues... i guess i could try using BigInt but i'll need to update the tsconfig lib our simulator targets

eanders-ms:
Ya, it may not be worth it. Calling fxMul/fxDiv on the TS side in order to keep the two implementations as close as possible is the more valuable aspect. If you did want to use fixed in both, casting to `int64_t` on the c++ side helps a lot:

```cpp
inline int fxMul(int a, int b) {
    return (int)(((int64_t)a * b) >> FX_SHIFT);
}

inline int fxDiv(int a, int b) {
    return (int)(((int64_t)a << FX_SHIFT) / b);
}
```

Then on the TS side, for precision, `| 0` the results to mimic c++ int32 wrapping.

riknoll:
yeah i copied your C++ fx functions so that cast is already there! (thanks for those by the way :smile:)

eanders-ms (approved):
Looks good!

riknoll:
here's the upload target! https://arcade.makecode.com/app/ceb234fed84bd2ecec33607dd7ec392d039dbfe8-5a91efdc18

PR #1511, opened 2025-03-27
fix pause until bug, change pointer events back

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6884

There are three bugs in this issue. One and three I technically already addressed here: https://github.com/microsoft/pxt-common-packages/pull/1508, but when testing and poking for _this_ bug, I realized that the pointer events had the correct associated numbers and we were actually handling the pressed state wrong. I just assumed that flipping them would be the solution since that was the fix for the mouse and wheel bug also fixed in the above PR. By flipping them, I actually introduced the same bug in the more generic `browswerEvents` block. I've switched the numbers back and fixed where the actual problem was with pressed so all of the events blocks behave as expected now.

Bug 2 in that issue helped reveal the problems above and is also fixed by using `MouseButtonEvent` rather than `KeyEvent`. `KeyEvent` still has pressed and released, but `MouseButtonEvent` has the correct event number that we want, so I made sure that all places with `KeyEvent` has been replaced with `MouseButtonEvent`.

### DIFF
--- a/libs/browser-events/browserEvents.ts
+++ b/libs/browser-events/browserEvents.ts
@@ -4,8 +4,8 @@
 //% block="Browser Events"
 namespace browserEvents {
     export enum Event {
-        PointerDown = 6858,
-        PointerUp = 6857,
+        PointerDown = 6857,
+        PointerUp = 6858,
         PointerMove = 6859,
         PointerLeave = 6860,
         PointerEnter = 6861,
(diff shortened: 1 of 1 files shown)


cameronary999:
@srietkerk when should this fix come to makecode extension

riknoll (approved):
this will require upgrade rules in pxt-arcade before it can be released. ping me for details if you haven't done that before

thsparks (approved):

PR #1515, opened 2025-04-15
fix queueing of browser events and handle scenes with the mouse

riknoll:
this fixes a bug reported on the forum here: https://forum.makecode.com/t/browser-events-extension-bug/35709

basically the bug worked like this:
1. someone presses a button and the simulator fires the pressed event
2. the game engine updates the pressed state of the button to true and invokes the user's button handler
3. the button handler pauses for some amount of time
4. the button is released and the simulator fires the released event
5. the game engine updates the pressed state of the button to false
6. the user presses the button a second time
7. because the button handler from step 3 is still paused, the pressed event is queued by the event bus
8. the button is released
9. the handler from step 3 finishes pausing and the queued event from step 7 is invoked
10. the game engine updates the button to "pressed" even though the corresponding released event already fired in step 8. the button is now stuck in the "pressed" state

to fix it, i did the same thing that we do for controller buttons which is to create separate internal events for pressed/released that we can subscribe to in the game engine (whereas the public pressed/released events are subscribed to by the user code). this prevents the queueing behavior from delaying the update of the button state.

i also refactored the mouse button code so that it too handles scene push/pop, which i forgot to do  in https://github.com/microsoft/pxt-common-packages/pull/1514

### DIFF
--- a/libs/browser-events/browserEvents.ts
+++ b/libs/browser-events/browserEvents.ts
@@ -48,40 +48,50 @@ namespace browserEvents {
         Released = Event.PointerUp,
     }
 
+    //% whenUsed
+    const INTERNAL_POINTER_DOWN = 6868;
+    //% whenUsed
+    const INTERNAL_POINTER_UP = 6869;
+
+    type MouseHandler = (x: number, y: number) => void;
+
     //% fixedInstances
     export class MouseButton {
         protected _pressed: boolean;
-        protected pressHandler: (x: number, y: number) => void;
-        protected pressListeners: ((x: number, y: number) => void)[];
-        protected releaseHandler: (x: number, y: number) => void;
-        protected releaseListeners: ((x: number, y: number) => void)[];
+
+        protected sceneStack: _SceneButtonHandlers<MouseHandler>[];
+
+        protected get state(): _SceneButtonHandlers<MouseHandler> {
+            return this.sceneStack[this.sceneStack.length - 1];
+        }
 
         constructor(public id: number) {
-            control.internalOnEvent(Event.PointerDown, this.id, () => this.setPressed(true), 16);
-            control.internalOnEvent(Event.PointerUp, this.id, () => this.setPressed(false), 16);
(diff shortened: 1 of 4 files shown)


libs/browser-events/browserEvents.ts
```diff
@@ -48,40 +48,50 @@ namespace browserEvents {
...
         protected _pressed: boolean;
-        protected pressHandler: (x: number, y: number) => void;
-        protected pressListeners: ((x: number, y: number) => void)[];
-        protected releaseHandler: (x: number, y: number) => void;
-        protected releaseListeners: ((x: number, y: number) => void)[];
+
+        protected sceneStack: _SceneButtonHandlers<MouseHandler>[];
+
+        protected get state(): _SceneButtonHandlers<MouseHandler> {
+            return this.sceneStack[this.sceneStack.length - 1];
```
srietkerk, line 65: Do we need protections for a possibly empty `sceneStack`?

libs/browser-events/sim/mouseState.ts
```diff
@@ -43,10 +48,19 @@ namespace pxsim.browserEvents {
                 "pointerout",
             ];
 
+            let eventId = 6857 + events.indexOf(event.type);
+
+            if (event.type === "pointerdown") {
+                eventId = INTERNAL_POINTER_DOWN;
+            }
+            else if (event.type === "pointerup") {
+                eventId = INTERNAL_POINTER_UP;
```
srietkerk, line 57: Do we also want to take advantage of key up and down here?

srietkerk (commented):
Couple questions

jwunderl (approved):

libs/browser-events/browserEvents.ts
riknoll: nope, that's handled by the push/pop handlers

libs/browser-events/sim/mouseState.ts
riknoll: those are fired in keyboard.ts

PR #1518, opened 2025-04-22
remove game dependency in mixer lib

riknoll:
should have fixed this a while ago... percent chance is defined in libs/game so it's not usable in mixer. works fine in arcade since game is already added to the project.


riknoll:
ah wait, there is one more...

PR #1538, opened 2025-04-25
update github actions

eanders-ms:


srietkerk (approved):
I don't fully understand why and where the `fromJSON` is needed, and I'm not super familiar with our workflows, but this looks clean to me!

PR #1546, opened 2025-05-30
add a math constants block

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/4276

adds a math constants block with all the usual constants, but most importantly pi

we already have all of the trigonometric functions in the toolbox, and you really need PI for them to be useful. 

image
image

### DIFF
--- a/libs/base/math.ts
+++ b/libs/base/math.ts
@@ -1,12 +1,27 @@
 namespace Math {
-
-    export const E = 2.718281828459045;
-    export const LN2 = 0.6931471805599453;
-    export const LN10 = 2.302585092994046;
-    export const LOG2E = 1.4426950408889634;
-    export const LOG10E = 0.4342944819032518;
+    //% blockIdentity="Math._constant"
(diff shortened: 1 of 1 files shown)


libs/base/math.ts
```diff
@@ -69,13 +84,21 @@ namespace Math {
...
     //% help=math/icos weight=10 advanced=true blockGap=8
     export function icos(theta: number) {
         return isin(theta + 16384);
     }
+
+    //% shim=TD_ID
+    //% block="$MEMBER"
+    //% constantShim
+    //% weight=0
+    export function _constant(MEMBER: number): number {
```
srietkerk, line 99: Does something special happen by using MEMBER? Or is it the `constantShim` that correctly populates all of the constants?

riknoll: yeah, this is me tricking the decompiler into using the right name for the field this generates. i would have added a comment but then it would have shown up in the tooltip for the block

PR #1609, opened 2026-03-22
Add an error message in the problems panel for when the condition is false.

DeveloperTryingToCodeLikeOtherOfThem:
Added an error message in the problems panel so users can immediately see the assertion failure message there, instead of needing to always inspect it in the console log panel every time. The difference is that this function adds a panic code merged in the problems panel.

Behaves similar to the image below.
image


DeveloperTryingToCodeLikeOtherOfThem (commented):
Replaced the explanation in the description.

PR #1610, opened 2026-03-22
Add clearer message and error code reporting for failed assertions

DeveloperTryingToCodeLikeOtherOfThem:
Edited file `assert.md` for explaining what the "Assertion failed" message represents in the problems panel.

### DIFF
--- a/libs/base/docs/reference/control/assert.md
+++ b/libs/base/docs/reference/control/assert.md
@@ -29,7 +29,7 @@ forever(function () {
 })
 ```
 
-## See also # see also
+## See also #see also
 
 [panic](/reference/control/panic)
 
--- a/libs/base/docs/reference/control/assert.md
+++ b/libs/base/docs/reference/control/assert.md
@@ -10,9 +10,6 @@ You can insist that your program will stop at an **assert** block if a certain c
 An 'Assertion failed' message containing the error code number is generated. This message is shown in the problems panel and also output to the the serial port. The format of the message looks like this:
 
 ``Fatal failure: Assertion failed, code=49``
-
-And also in the problems panel, there will be a message displaying "Assertion failed", similar to how the serial output logs the message.
-
 ## Parameters
 
 * **cond**: a [boolean](/types/boolean) where `true` means everything is ok or `false` which means, stop the program!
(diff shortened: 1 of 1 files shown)


ganicke (approved):
Some edits to align with what the source shows:

```typescirpt
  /**
   * Display an error code and stop the program when the assertion is `false`.
   */
    //% help=control/assert weight=30
    //% blockId="control_assert" block="assert %cond|with value %code"
    export function assert(cond: boolean, code: number) {
        if (!cond) {
            fail("Assertion failed, code=" + code)
        }
    }

    export function fail(message: string) {
        console.log("Fatal failure: ")
        console.log(message)
        dmesg(message)
        panic(108)
    }
```

DeveloperTryingToCodeLikeOtherOfThem:
The new edits are changed here. [New Changes](https://github.com/microsoft/pxt-common-packages/pull/1609) Also unchopped the 'see also' heading back into to the docs.

libs/base/docs/reference/control/assert.md
```diff
@@ -1,16 +1,21 @@
...
 control.assert(false, 0)
 ```
 
-You can insist that your program will stop at an assert block if a certain condition you check is false. The error number in the assert is written to the serial port with a failure message.
+You can insist that your program will stop at an **assert** block if a certain condition you check is false. 
+An 'Assertion failed' message containing the error code number is generated. This message is shown in the problems panel and also output to the the serial port. The format of the message looks like this:
+
+``Fatal failure: Assertion failed, code=49``
+
+And also in the problems panel, there will be a message displaying "Assertion failed", similar to how the serial output logs the message.
```
ganicke: This is redundant with the statement above the example.

libs/base/docs/reference/control/assert.md
```diff
@@ -24,10 +29,10 @@ forever(function () {
 })
 ```
 
-## See also #seealso
+## See also # see also
```
ganicke: Heading tags are used for document inheritance in the case that the target platform wants to insert extra information under that heading. They should have no spaces in them. They are not rendered. You previously removed the others which were put in for the purpose of document inheritance. We'll either remove them or leave the as they were.

DeveloperTryingToCodeLikeOtherOfThem:
Removed the repeated statement and removed the space making no space for the heading.

PR #1614, opened 2026-04-14
Fix simulator restart not stopping all concurrent sound threads

KohanMathers:
Fixes issue #7552 in microsoft/pxt-arcade.

The `onStopAll` callback in the mixer simulator code was only set up when sequencers were created, and it only stopped sequencer-based audio. Direct calls to `music.playInstructions()` bypassed this mechanism and weren't stopped on simulator restart.

### Fix
- Added a `setupOnStopAll()` function that ensures the `onStopAll` callback is registered to call `AudioContextManager.muteAllChannels()` when the simulator stops audio.
- Modified `playInstructions()` and `queuePlayInstructions()` to call `setupOnStopAll()`, ensuring the callback is set up whenever these functions are used.
- Enhanced the `onStopAll` callback to call `muteAllChannels()` in addition to stopping sequencers.

This ensures that all concurrent sound threads are properly stopped when the simulator restarts, resolving the sound stacking issue.

### DIFF
--- a/libs/mixer/sim/music.ts
+++ b/libs/mixer/sim/music.ts
@@ -1,9 +1,11 @@
 namespace pxsim.music {
     export function playInstructions(b: RefBuffer) {
+        setupOnStopAll();
         return AudioContextManager.playInstructionsAsync(b.data)
     }
 
     export function queuePlayInstructions(when: number, b: RefBuffer) {
+        setupOnStopAll();
         AudioContextManager.queuePlayInstructions(when, b)
     }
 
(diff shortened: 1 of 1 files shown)


KohanMathers:
@riknoll This is the fix to the issue in arcade :)

libs/mixer/sim/music.ts
```diff
@@ -30,19 +32,27 @@ namespace pxsim.music {
         sequencer: Sequencer;
     }
 
-    let sequencers: SequencerWithId[];
-    let nextSequencerId = 0;
+    let onStopAllSetup = false;
```
riknoll, line 37: i think you accidentally removed the `sequencers` and `nextSequencerId` declarations here

riknoll (changes requested):

KohanMathers:
@riknoll you're right, oops - fixed in 41fa123

riknoll (approved):

PR #1615, opened 2026-04-17
Add in new and shuffled missing reference docs - pass 1

ganicke:
Adding some docs that were either not written or misplaced in previous library shuffles. These were caught with my c0pilot generated scan utility.


abchatra:
Very cool @ganicke. @riknoll can you review the tilemap help pages? 

I am bit surprised map was missing. That block is widely used for a while.

abchatra (approved):

ganicke:
The 'map' doc has been around but it wasn't included when this lib was put together.

PR #1624, opened 2026-06-14
Check buffer size in __packUnpackCore

samtygier:
Reimplement `packedSize()` so that it does not depend on running `__packUnpackCore()`

The `if (buf)` is no longer needed, because it is only used in the old `packedSize()` path to find the format size.

This helps catch the issue in https://github.com/microsoft/pxt-microbit/issues/6894 by refusing to unpack if the buffer is smaller than described in the format


samtygier:
@microsoft-github-policy-service agree

PR #1638, opened 2026-08-20
fix chunking of text

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/7627

test proj / build https://arcade.makecode.com/app/3a9c080d03e50e68b94f4b7be89820870d6d1b0e-c521772cee#pub:_iRk0Cg1pD2oY

fix chunking when the last break is the dstart of the current line, so that it doesn't add an extra line then overflow

image


aznhassan (approved):
:shipit:

PR #1642, opened 2026-09-23
detach accelerometer between runs

jwunderl:
add way to detach accelerometer events, to avoid memory leak of runtime between runs


aznhassan (approved):
Good catch!

